PR #119, opened 2021-09-10
Add explicit casts with "the"

### LINKED ISSUE #87: Add COALTON:THE for local type declarations
It is sometimes necessary to make a local type declaration to disambiguate a type. I propose adding a new syntax node called `(coalton:the <type> <expr>)` so that one can stipulate that `<expr>` has type `<type>`. It seems kind of gnarly to do if there are repeated type variables or something.

This came up because I have a `(dividable Integer Fraction)` and

```
(coalton-toplevel
  (define xxx (/ 1 2)))
```

errors saying it can't deal with `(DIVIDABLE INTEGER :A)`, which makes sense. If I do

```
(coalton-toplevel
  (declare xxx Fraction)
  (define xxx (/ 1 2)))
```

it works fine. But imagine this were more nested. We'd either have to hope types propagate correctly from the toplevel declaration, or we'd be out of luck.

eliaslfox:
Resolves #87

### DIFF
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -395,3 +395,12 @@ Coalton has a do-notation macro that works similary to do notation in Haskell.
     ;; [6+3, 5+3, 5+3, 6+2, 5+2, 4+2, 6+1, 5+1, 4+1]
    (define xs (f (make-list 1 2 3) (make-list 4 5 6))))
 ```
+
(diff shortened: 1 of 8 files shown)


colescott (changes requested):
Please add doumentation for `the` to the intro to coalton doc

colescott (approved):
Looks great!

PR #120, opened 2021-09-10
Add a warning when intermediary form in seq is a function

### LINKED ISSUE #60: Add a warning when a form in a seq returns a function type
It is too easy to not fully apply a function.

eliaslfox:
This is to stop unintentionally partially applying a function.

Resolves #60

### DIFF
--- a/src/typechecker/derive-type.lisp
+++ b/src/typechecker/derive-type.lisp
@@ -195,10 +195,14 @@ Returns (VALUES type predicate-list typed-node subs)")
       (loop :for elem :in initial-elements :do
         (multiple-value-bind (tyvar preds_ node subs_)
             (derive-expression-type elem env subs)
-          (declare (ignore tyvar))
+
           (setf subs subs_)
           (setf preds (append preds preds_))
-          (push node nodes)))
+          (push node nodes)
+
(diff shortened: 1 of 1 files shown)


colescott:
This is probably the way to go in the short term but I would like to hear your thoughts on forcing all expressions (except the last) in a `progn` to unify to `Unit` then providing a shorthand for `(const Unit <expr>)`

(cc @stylewarning)

stylewarning:
I agree with this PR and @colescott's wondering.

src/typechecker/derive-type.lisp
```diff
@@ -195,10 +195,14 @@ Returns (VALUES type predicate-list typed-node subs)")
...
             (derive-expression-type elem env subs)
-          (declare (ignore tyvar))
+
           (setf subs subs_)
           (setf preds (append preds preds_))
-          (push node nodes)))
+          (push node nodes)
+
+          (when (function-type-p (apply-substitution subs tyvar))
+            (alexandria:simple-style-warning "Expression ~A in progn evaluates to a function."
```
stylewarning, line 204: Maybe you should add something like

```
This may be OK, but is also a common mistake in the event a function is accidentally curried."
```

?

colescott (approved):
lgtm

Issue #37, opened 2021-08-31
Pointfree Transform


eliaslfox:
Transform the following code
```
(define f
  (map (+ 1))
```
Into this code

```
(define (f xs)
  (map (+ 1) xs))
```

This will reduce closures and allow for more direct calls.

eliaslfox:
Should also support

(define (length) (fold ...)) -> (define (length l) (fold ... l))

PR #124, opened 2021-09-10
Add initial pointfree transform.

eliaslfox:
Add an initial pointfree transform which can handle the following cases

(define m map) -> (define (m a b) (map a b))

(define (m x) (map x)) -> (define (m x a) (m map x a))

See #37


colescott (approved):
lgtm

PR #138, opened 2021-09-12
Add a few improvements to the Intro to Coalton doc

kilimanjaro:
A few of the examples gave errors for me when I went through the intro doc, so I've fixed those up. Two main things

- `coalton-toplevel` only allows definitions or declarations, so for examples where we just want to evaluate something at a repl i've adopted the `coalton` form. I wasn't sure if this was kosher, but I do think it's important to be clear that people shouldn't rely on the Lisp repl to evaluate Coalton stuff (well, they can do it, but whether it works is another story).
- `Int` type is now `Integer`


eliaslfox (approved):
Thanks for the improvements.

PR #149, opened 2021-09-13
add JSON parser example

stylewarning:
This adds an example piggybacking off of an existing JSON library.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,3 +55,6 @@ jobs:
       - uses: 40ants/run-tests@v2
         with:
           asdf-system: coalton
+          run-tests: |
+            (ql:quickload :coalton/tests)   ; get deps
+            (asdf:test-system :coalton)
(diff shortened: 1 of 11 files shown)


stylewarning:
Tests pass locally with `make test`.

stylewarning:
The build error happens because this commit adds a new dependency `json-streams` which isn't being downloaded by the test runner:

```
  An unhandled error condition has been signalled:
     Component #:JSON-STREAMS not found, required by #<SYSTEM "coalton-json">
```

eliaslfox:
Is this a caching issue?

stylewarning:
> Is this a caching issue?

I believe so. I'm not sure how or when or where the system detects when to download dependencies.

Presumably existing dependencies are cached and new ones aren't auto-downloaded.

stylewarning:
@eliaslfox @colescott I modified the workflow to force a `quickload` on the test package... not sure if this is the best way to do things from a caching perspective, but it works.

eliaslfox (approved):
shipit

eliaslfox:
My guess is that because the dependency is located in a separate asd file, it isn't invalidating the cache. This dependency is going to be pulled on every test run so it should be fixed soon.

If that is the issue, then either we use a single asd file or change the cache heuristics for the ci pipeline.

Issue #145, opened 2021-09-12
Replace `COALTON-LIBRARY:BOOLEAN` with `CL:T` and `CL:NIL`


eliaslfox:
This will allow using standard if constructs instead of using trivia to match on a class.

PR #167, opened 2021-09-13
fix Unit matching, add Lisp-native Boolean type

### LINKED ISSUE #166: error matching against `Unit`
```
                (define (f x)
                  (match x
                    ((Unit) 0))))
```

==>

```
Unknown symbol in match branch. UNIT is not a constructor.
in branch (UNIT)
in match on X
in definition of F
in COALTON-TOPLEVEL
```

stylewarning:
This fixes #166 and #145


stylewarning:
I took the "easiest" route to getting pattern compilation to work, but I don't think it's the best way at the end.

I'd prefer that there be some way to generalize the `classname` arg to ctor-entries to allow for something more freeform; perhaps it can be an arbitrary Trivia pattern?

I think it deserves thought, but I don't think we need to nail it perfectly for this PR.

eliaslfox (approved):
This is kinda cursed.

stylewarning:
I'm going to wait for @colescott to review as well, so he's in-the-loop about this.

eliaslfox:
Don't merge this yet. I want to attempt to write a slightly less cursed version.

eliaslfox:
#174 fixes this mostly. Coalton booleans still aren't lisp booleans but comparing symbols compiles to almost the same code as using cl:if on lisp booleans.

PR #176, opened 2021-09-13
Use lisp booleans for coalton booleans

### LINKED ISSUE #145: Replace `COALTON-LIBRARY:BOOLEAN` with `CL:T` and `CL:NIL`

eliaslfox:
Fixes #145


colescott (approved):
:)

PR #178, opened 2021-09-14
Compile `COALTON:IF` into `CL:IF`

eliaslfox:


eliaslfox:
```
COALTON-USER> (coalton-toplevel (define (f x)
                (if x
                    4
                    5)))
; No value
COALTON-USER> (cl:disassemble #'f)
; disassembly for F
; Size: 34 bytes. Origin: #x53178587                          ; F
; 87:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; 8B:       488945F8         MOV [RBP-8], RAX
; 8F:       80FB17           CMP BL, 23
; 92:       BA08000000       MOV EDX, 8
; 97:       41BB0A000000     MOV R11D, 10
; 9D:       490F44D3         CMOVEQ RDX, R11
; A1:       488BE5           MOV RSP, RBP
; A4:       F8               CLC
; A5:       5D               POP RBP
; A6:       C3               RET
; A7:       CC10             INT3 16                          ; Invalid argument count trap
```

```
COALTON-USER> (cl:defun g (x)
                (cl:if x
                       4
                       5))
COALTON-USER> (cl:disassemble #'g)
; disassembly for G
; Size: 38 bytes. Origin: #x531788B0                          ; G
; B0:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; B4:       488945F8         MOV [RBP-8], RAX
; B8:       4881FB17011050   CMP RBX, #x50100117              ; COMMON-LISP:NIL
; BF:       BA08000000       MOV EDX, 8
; C4:       41BB0A000000     MOV R11D, 10
; CA:       490F44D3         CMOVEQ RDX, R11
; CE:       488BE5           MOV RSP, RBP
; D1:       F8               CLC
; D2:       5D               POP RBP
; D3:       C3               RET
; D4:       CC10             INT3 16                          ; Invalid argument count trap
```

stylewarning:
Isn't making a new NODE like taking a sledgehammer to the problem?

stylewarning:
What about just detecting two-branch ENUM `match` and just codegenning an `if` of the first pattern?

eliaslfox:
Yeah kinda. The alternative is having an `if` macro that expands into a match and then compiling that match back into an if.

stylewarning:
> Yeah. The alternative is having an `if` macro that expands into a match and then compiling that match back into an if.

smol language core

PR #179, opened 2021-09-14
Remove all boxing

### LINKED ISSUE #28: Add "newtype" functionality
We should add Haskell-inspired `newtype` functionality, which allows us to create a "cheap" type abstraction that gets type checked but has no runtime penalty.

I'm not sure what the syntax should be but assuming it's

```
(define-newtype (F :t1 :t2 ...) (F ty))
```

then `(F s1 s2 ...)` would be a type whose representation is runtume-equivalent to `ty[:ti = si]`. This means that the constructor function `#'F` would just be identity (and ought to be inlined, since it can still be used as a high-order function).

I *think* codegen would be trivial (i.e., do nothing), _except_ in the case of codegenning `match`, which must just be a let-rebinding, since runtime-type inspection can't happen:

```
(match outer
  ((Foo inner) ...))

; ==>

(let ((inner outer))
  ...)
```

eliaslfox:
Instead of boxing an unknown value, Lisp-Object is now only a flag for the compiler to generate `t` as the type annotation for sbcl.

Also added auto newtyping in the second commit.

Resolves #28


stylewarning:
Update docs too?

eliaslfox:
@stylewarning which docs?

eliaslfox:
Vectors are fully unboxed now
```
CL-USER> (coalton-library:make-vector-capacity 10)
#()
```

src/typechecker/parse-type-definition.lisp
```diff
@@ -9,7 +9,8 @@
   (name :type symbol)
   (type :type ty)
   (runtime-type :type t)
-  (compressed-type :type t)
+  (compressed-type :type boolean)
```
stylewarning, line 12: Can you add or reference documentation for these two slots?

src/typechecker/environment.lisp
```diff
@@ -47,93 +62,106 @@
...
             :type tString
-            :compressed-type nil))
+            :compressed-type nil
+            :newtype nil))
 
           ('coalton:Lisp-Object
            (type-entry
             :name 'lisp-object
-            :runtime-type 'coalton-impl::lisp-object
+            :runtime-type t
```
stylewarning, line 153: `'t`

src/typechecker/environment.lisp
```diff
@@ -47,93 +62,106 @@
...
             :name 'U64
             :runtime-type '(unsigned-byte 64)
             :type tU64
-            :compressed-type nil))
+            :compressed-type nil
+            :newtype nil))
 
           ('coalton:Integer
            (type-entry
             :name 'Integer
```
stylewarning, line 98: These type names should be corrected.

```
:name 'coalton:Integer￼
:type 'cl:integer
```

Likewise for the rest (character, *-float, string, etc)

stylewarning (changes requested):
Very exciting change! Just some small things.

stylewarning:
I think auto-newtyping isn't good. I think there's tangible value in being able to box, because you might be creating type-safe data structures for use in *Lisp*. I would highly suggest adding a new top level form and letting `define-type` always create a new class for user-defined code.

*Or* we can think about some syntax to adorn `define-type` with to make it a newtype.

src/typechecker/environment.lisp
```diff
@@ -190,7 +218,7 @@
             :scheme (to-scheme (qualify nil tBoolean))
             :arguments nil
             :classname 'coalton::Boolean/False
```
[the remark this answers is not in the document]
eliaslfox: Kinda. Because Boolean is a "compressed type" (better name pending) that field isn't used.

eliaslfox:
I think auto-newtyping is because then Coalton will automatically do the right thing. I think instead there should be a syntax to disable auto-newtyping  for FFI purposes. See [repr_transparent](https://doc.rust-lang.org/1.26.2/unstable-book/language-features/repr-transparent.html) in Rust.

eliaslfox:
Sorry `repr_transparent` was a bad example. I meant [repr_c](https://doc.rust-lang.org/nomicon/other-reprs.html) as an example of opting into worse codegen for ABI reasons.

stylewarning:
Don't you think that auto-NT is a compromise on the Lisp interface, and violates the principle of least surprise? In Rust it makes sense; Rust is a self-contained independent language that only has to interact with itself. In the rare event you'll interact with C, you have functions for nice ABI handling.

With the current Coalton setup, in Lisp, a `define-type` can be inhabited by
- special native Lisp values (e.g. `Boolean`)
- symbols (for enum-like types)
- native Lisp objects
- "pass throughs" in the case of single-constructor single-argument
- entirely new class definitions otherwise

These decisions seem most beneficial for the speed of Coalton, but seem to sacrifice three things:

1. A clear and easy to understand expectation of what a particular type definition will be in Lisp,

2. Abject incompatibility with CLOS, and

3. Continuity in an incremental development model (e.g., if I add a constructor, it's very likely my whole representation will change).

My instinct suggests that the default should be the least surprising (e.g., just `DEFCLASS` everything), and something like an optimize directive should give make things fast.

stylewarning:
Two clarifications:

First, when I said "CLOS incompatibility", I meant that some `define-type`s could be used with `defmethod` and some can't—and that can suddenly change if the constructors change. (Think `print-object`; or think of QUILC AST generic functions.)

Second, when I said "`DEFCLASS` everything", I meant for new user-defined types, and not for some choice standard library types, like `Boolean`.

eliaslfox:
I think that ossifying aspects of Coalton's default codegen for Lisp interop will inhibit optimizations down the road. 

Although the main benefit of Coalton is its lisp interop, I imagine that in real programs it is unlikely that Lisp and Coalton will frequently call back and forth in a single call stack. Instead, I think it is much more likely that large Lisp and Coalton systems will communicate by passing a limited number of objects back and forth, similar to calling an HTTP API. If the number of types passed between Lisp and Coalton are limited compared to the total number of types defined in Coalton, then it makes sense to annotate the types being used for interop.

In addition, requiring explicit annotations on types used for interop will let us evolve Coalton's codegen without breaking backward compatibility. Coalton users will then get faster programs without having to add annotations to use new compiler features.

I added another commit that shows what this could look like. By adding `(repr :lisp)` above a `declare-type` Coalton will use the lisp friendly struct based codegen always.

```
(coalton-toplevel
  (repr :lisp) ;; Opt out of the newtype optimization
  (define Username (Username String))

  (repr :lisp) ;; Opt out of the enum optimization
  (define Color Red Green Blue))
```

stylewarning:
Your suggestion isn't wrong or flawed, but does present a large philosophical mismatch with Common Lisp. If I had to describe Lisp's tenets:

1. Everything is run-time re-definable (unless you opt-out).
2. Everything is accessible and inspectable (unless you opt-out).
3. Optimization is opt-in.

To give concrete Lisp examples of the above:

1. Functions and classes can be redefined incrementally and at run-time. (Structures cannot but structures are opt-in and their functionality is covered by classes.)
2. Class definitions are accessible by the programmer. Function definitions can be looked up. Etc.
3. Optimization (usually including TCO) requires an explicit DECLARE; inlining must be explicit; inter-module optimization must be explicit; stack allocation must usually be explicit; function calls are indirect (allowing redefinition).

Commits like this one and previous ones depart from this philosophy.

So I think we must ask ourselves: What should Coalton's philosophy be? (Assuming you agree with the following dichotomy:) Do we want it to be a traditional, batch-compiled, aggressively optimized language by default? Or do we want an incremental and interactive language where a programmer can reasonably expect how everything works by default?

You've already hinted at your thoughts by suggesting ordinary Coalton programs will only expose a very small amount to ordinary Lisp programs.

Part of the answer to this question is, I think, the audience. If we make Coalton as close to Rust as possible with language design choices, whom are we appeasing?

In my original vision, Coalton was certainly of the latter philosophy, because I wasn't ambitious enough to think that anybody would be writing 100% Coalton programs, and instead making use of the plethora of existing Common Lisp code and exposing Coalton to the world as if it were "ordinary" Lisp. QUILC was the main motivation, which has lots of hairy data structures and so whatever Coalton produces, it has to be CL compatible, at least for the foreseeable future.

Obviously with your suggestion and mine, we can have it both ways, but it's a matter of philosophy and how far we decide to take this philosophy. I think honoring a consistent philosophy (and codifying it) is essential to stable and coherent programming language development.

P.S. Wouldn't adopting this philosophy also suggest that we also ought to abstract away actual functions from Lisp users, requiring functions too be marked as Lisp, so compiler authors enjoy the same freedoms with function optimizations as they do with type optimizations?

stylewarning:
See #180 for related discussion.

stylewarning (approved):
we agreed on a follow-on PR to support the interaction mode, add some `deftypes` and a few other things

PR #186, opened 2021-09-17
Development mode

eliaslfox:


src/codegen/program.lisp
```diff
@@ -40,6 +40,14 @@
           :else :if (and (typep thing '(cons (member cl:defstruct)))
                          (not (sub-struct-p thing)))
                   :collect thing :into defstructs
+          ;; (DEFCLASS (FOO ... (:INCLUDE ...
+          :if (and (typep thing '(cons (member cl:defclass)))
+                   (not (null (third thing))))
+            :collect thing :into defsubstructs
+          ;; (DEFSTRUCT FOO ...
+          :else :if (and (typep thing '(cons (member cl:defclass)))
```
stylewarning, line 48: this is matching against DEFCLASS here. Do you mean DEFSTRUCT or are the docs wrong?

src/codegen/compile-type-definition.lisp
```diff
@@ -19,6 +24,58 @@
...
+                            (declare (type stream stream)
+                                     (type ,classname self)
+                                     (values ,classname))
+                            (format stream "#.(~s~{ ~s~})"
+                                    ',(constructor-entry-name constructor)
+                                    (list ,@(mapcar (lambda (slot) `(slot-value self ',slot)) slot-names)))
+                            self)))
+
+             :append (cond
+                       ((= 0 (constructor-entry-arity constructor))
```
stylewarning, line 66: zerop

src/codegen/compile-type-definition.lisp
```diff
@@ -19,6 +24,58 @@
...
+      ((eql :development *interaction-mode*)
+       `((defclass ,(type-definition-name def) ()
+           ())
+
+         ,@(loop
+             :for constructor :in (type-definition-constructors def)
+             :for classname := (constructor-entry-classname constructor)
+             :for slot-types := (mapcar #'coalton-impl/typechecker::fresh-inst (constructor-entry-arguments constructor))
+             :for slot-names := (ctor-make-slot-names (length slot-types) package)
+             :append
```
stylewarning, line 36: minor, but I prefer

```
:append
(list `(...))
```

To me, the backtick is a signifier of "syntax" (aka: "im writing a backtick because i want to quote syntax"), and which it's also used to construct a list, it can be confusing, even if equivalent

src/codegen/compile-type-definition.lisp
```diff
@@ -1,5 +1,10 @@
 (in-package #:coalton-impl/codegen)
 
+(deftype interaction-mode ()
+  `(member :development :release))
+(declaim (type interaction-mode *interaction-mode*))
+(defvar *interaction-mode* :development)
```
stylewarning, line 6: document

stylewarning (changes requested):

src/codegen/program.lisp
eliaslfox: The comment was wrong

src/codegen/compile-type-definition.lisp
eliaslfox: The pattern of ``    `(( `` is used throughout Coalton already, and `(list` adds more indentation.

Also isn't a list of quoted forms "quote syntax"?

stylewarning: Ideally

    `(foo ...)

implies that

    (foo ...)

is a valid form (i.e., evaluatable expression, or sub-expression of an evaluatable expression).

    `((foo ...))

rarely falls under this class.

This is a purely stylistic point (like `':quoted-keywords`, two-branch-`if`, `cond`-for-effects, etc).

eliaslfox: That makes sense. I can update this file.

src/codegen/program.lisp
```diff
@@ -40,6 +40,14 @@
           :else :if (and (typep thing '(cons (member cl:defstruct)))
                          (not (sub-struct-p thing)))
                   :collect thing :into defstructs
+          ;; (DEFCLASS (FOO ... (:INCLUDE ...
+          :if (and (typep thing '(cons (member cl:defclass)))
```
stylewarning, line 44: Two issues:

1. The comment should read `(DEFCLASS FOO (...) ...)`
2. You need an `:else` in front of this `:if`!

The second point is causing things to be defined multiple times.

stylewarning (changes requested):

stylewarning:
Tests pass, but there's still a DEFSTRUCT to convert in `compile-class-definitions`. I bet it would be easiest to just make a utility macro `defstruct-as-class` or something so you don't have to write all the accessors out and everything.

eliaslfox:
@stylewarning this is ready for review. 

Also I seem to have confused github actions. Feel free to merge this whenever you think it's ready.

src/coalton.lisp
```diff
@@ -18,6 +18,10 @@
                                            coalton:define-instance
                                            coalton:repr))
 
+(when (string= (uiop:getenv "COALTON_ENV") "release")
+  (format t "COALTON starging in release mode~%")
```
colescott, line 22: starging -> starting

colescott: Also do we for sure want it to print something in release mode or would it make more sense to have it only print when coalton starts up in development mode (with some text telling you how to start in release)

src/codegen/compile-type-definition.lisp
```diff
@@ -54,33 +75,25 @@
...
+                          (list `(coalton-impl::define-global-lexical
+                                     ,(constructor-entry-name constructor)
+                                   ,entry))))))
 
-(defun ctor-make-slots (slot-names slot-types env)
-  (loop :for slot-name :in slot-names
-        :for slot-type :in slot-types
-        :collect (ctor-make-slot slot-name slot-type env)))
+         ,@(when (eql coalton-impl::*interaction-mode* :release)
+             #+sbcl
```
colescott, line 96: I think in this case you should splat out a list which could contain multiple possible flags (for non-sbcl impl specific stuff) and just have `#+sbcl` to add to the list

src/codegen/optimizer.lisp
```diff
@@ -21,7 +21,10 @@
   ;; Run direct application twice so that options for direct
   ;; application are not removed by the pointfree transform
   (let* ((node (direct-application-transform node optimizer))
-         (node (pointfree-transform node optimizer))
+         ;; Classes cannot be stack allocated
+         (node (if (eql coalton-impl:*interaction-mode* ':release)
```
colescott, line 25: I'm seeing this `(if (eql coalton-impl:*interaction-mode* ':release) ...` a lot. Would it make sense to have a `with-release-mode` and `with-development-mode` or something similar?

src/codegen/program.lisp
```diff
@@ -189,9 +196,7 @@
                      :for return-type := (typed-node-type (typed-node-abstraction-subexpr node))
                      :append (compile-function name vars type return-type subexpr env))))
 
-          ((and (= 1 (length scc-typed-bindings))
-                t ;; TODO: We need to check that the variable is non-self-recursive using free-variables
```
colescott, line 193: Does this get checked earlier on now?

src/codegen/program.lisp
```diff
@@ -203,42 +208,42 @@
...
+              (loop :for (superclass-pred . field-name) :in (ty-class-superclass-dict class)
+                    :for superclass := (lookup-class env (ty-predicate-class superclass-pred))
+                    :collect (make-struct-or-class-field
+                              :name field-name
+                              :type (ty-class-codegen-sym superclass)))
+
+              (loop :for method :in (ty-class-unqualified-methods class)
+                    :collect (make-struct-or-class-field
+                              :name (car method)
+                              ;; TODO: Add type annotation here
```
colescott, line 231: TODO

colescott (approved):
Looks great! I just put a few comments down for small things but otherwise looks good to merge.

I am assuming documentation for this is waiting on #19 to be done

src/coalton.lisp
eliaslfox: We could log in development mode, that is what React does. I thought that logging to the repl every time Coalton is loaded sounded kinda annoying.

src/codegen/program.lisp
eliaslfox: Yeah this is now checked in derive-type

eliaslfox: See [https://github.com/coalton-lang/coalton/blob/main/src/typechecker/derive-type.lisp#L315-L322](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/derive-type.lisp#L315-L322)

src/codegen/optimizer.lisp
eliaslfox: I don't see that as being easier to read, but maybe I'm not really understanding what you're suggesting. Would that look like this?

```lisp
(with-release-mode
  (do stuff for release mode)
  (do stuff for development mode))
```

stylewarning: I myself would prefer:

```
(ecase *development-mode*
  (:development ...)
  (:release ...))
```

so that we are always explicit about what is what, and we have no other cases

src/codegen/program.lisp
```diff
@@ -36,10 +36,18 @@
           :if (and (typep thing '(cons (member cl:defstruct)))
                    (sub-struct-p thing))
             :collect thing :into defsubstructs
-          ;; (DEFSTRUCT FOO ...
+          ;; (DEFSTRUCT FOO () ...
```
stylewarning, line 39: nit: `defstruct foo ()` is invalid syntax; it was correct before because it doesn't have an `:include` clause

src/codegen/compile-type-definition.lisp
```diff
@@ -54,33 +75,24 @@
...
-  `(,name (error "") :type ,(lisp-type type env) :read-only t))
+                          (list `(coalton-impl::define-global-lexical
+                                     ,(constructor-entry-name constructor)
+                                   ,entry))))))
 
-(defun ctor-make-slots (slot-names slot-types env)
-  (loop :for slot-name :in slot-names
-        :for slot-type :in slot-types
-        :collect (ctor-make-slot slot-name slot-type env)))
+         ,@(when (eql coalton-impl::*interaction-mode* :release)
```
stylewarning, line 95: nit: `':release`

src/codegen/utilities.lisp
```diff
@@ -0,0 +1,60 @@
...
+                          (mode (error "Mode required")))
+  (declare (type symbol classname)
+           (type symbol constructor)
+           (type symbol superclass)
+           (type list fields)
+           (type (member :class :struct) mode))
+
+  (let ((field-names (mapcar #'struct-or-class-field-name fields)))
+
+    (cond
```
stylewarning, line 24: maybe a `case` or `ecase`?

src/settings.lisp
```diff
@@ -0,0 +1,26 @@
...
+;; *interaction-mode* determines how redefinable code generated by Coalton is.
+;;
+;; In development mode Coalton turns all declare-type forms into CLOS
+;; objects to support a repl based workflow.
+;;
+;; In release mode Coalton compiles declare-type forms into frozen
+;; structs or even more optimal layouts which may not support
+;; redefinition.
+;;
+;; This variable *must* be set before the standard library is loaded.
```
stylewarning, line 22: Are Coalton development mode and release mode mixable? Can one library be compiled as release and another as development? Can Coalton be compiled with release, and the rest of the world with development?

src/coalton.lisp
```diff
@@ -18,6 +18,10 @@
                                            coalton:define-instance
                                            coalton:repr))
 
+(when (string= (uiop:getenv "COALTON_ENV") "release")
```
stylewarning, line 21: I would use case-insensitive `string-equal` just for kicks.

src/coalton.lisp
```diff
@@ -18,6 +18,10 @@
...
+  (format t "COALTON starting in release mode~%")
```
stylewarning, line 22: I would prepend the line with a `;; `

src/coalton.lisp
stylewarning: Why not have this set in `config.lisp`?

stylewarning (changes requested):
a few questions and comments

src/settings.lisp
eliaslfox: In the current implementation I think that even within a single module different `coalton-toplevel` forms could be compiled in different modes. However I would prefer not to codify that as allowed behavior because we may want to do some form of whole program analysis in the future.

tests/graph-tests.lisp
```diff
@@ -60,7 +60,7 @@
      (coalton-library::NodeIndex 0)
      g)
 
-    (is (equalp (coalton-library::tarjan-scc g)
+    (is (coalton-library::tarjan-output-equal (coalton-library::tarjan-scc g)
```
stylewarning, line 63: nit: indentation is off here and below

src/codegen/compile-type-definition.lisp
```diff
@@ -6,41 +6,62 @@
...
+               `(coalton-impl::define-global-lexical
+                    ,(constructor-entry-name constructor)
+                    (coalton-impl/codegen::F1 #',(constructor-entry-name constructor))))))
 
       (t
-       `((defstruct (,(type-definition-name def)
-                     (:constructor nil)
-                     (:predicate nil)
-                     (:copier nil)))
+       `(,(if (eql coalton-impl::*interaction-mode* :release)
```
stylewarning, line 25: nit: quote `':release`

eliaslfox:
CI is straight up not having a good time. I've had to disable caching twice on this pr to get tests to pass :/

stylewarning (approved):

PR #188, opened 2021-10-02
Rename shadow-realm -> immutable-map

eliaslfox:
And rename shadow-list -> immutable-listmap


eliaslfox:
@stylewarning I think you might need to update the settings for blockings PRs on CI.

stylewarning (approved):

PR #194, opened 2021-10-12
Add replace-node-type for seq nodes

eliaslfox:
Not sure how this hasn't been an issue before.


stylewarning:
@eliaslfox on a roll

PR #198, opened 2021-11-06
Describe Coalton-Lisp interop

stylewarning:
Here is an attempt to describe exactly what Coalton promises in interop. I would love feedback, things that are wrong, and things that I've missed.

### DIFF
--- a/docs/coalton-lisp-interop.md
+++ b/docs/coalton-lisp-interop.md
@@ -1,0 +1,199 @@
+# Coalton-Lisp Interoperation
+
+## Introduction
+
+Coalton is a language embedded in Lisp, and indeed, Coalton compiles to Lisp code. This document describes what Coalton promises for interoperation.
+
+## Interaction Mode
+
+First and foremost, there are two ways to globally compile Coalton. This is determined by an environment variable `COALTON_ENV` which in turns sets a global variable `*interaction-mode*`. There are two options for this variable:
+
+- `COALTON_ENV=development`, which sets `*interaction-mode*` to `:development`
+- `COALTON_ENV=release`, which sets `*interaction-mode*` to `:release`
+
+As the names imply, these two modes have different behaviors that optimize for either development or release.
+
+**WARNING**: *All* Coalton code, *including* the standard library, *must* be compiled in the same mode. We advise application writers to set the environment variable during their application's build process before Coalton is compiled, and to ensure caches are not stale.
+
+By default, the interaction mode is `:development`.
+
+### Development Mode
+
+Development mode allows most structures to be re-defined. This means:
+
+* Most types turn into CLOS classes.
+* Most types are not flattened or unwrapped in any way.
+* Certain optimizations that obscure debugging are disabled.
+
+### Release Mode
+
+Release mode freezes most structures and optimizes them.
+
+* Most types are defined as frozen `defstruct`s or flattened Lisp data types.
+* Many optimizations are performed.
+
+### Pitfalls of Having Two Modes
+
+Unfortunately, having two modes may mean one inadvertently depend on the behavior in one mode that is not supported in the other. We advise testing your code against both modes.
+
+## Promises of Data and Basic Data Types
+
+**PROMISE**: Every Coalton value exists as a Lisp value. The structure of said value may not always be defined, however, and must be considered an element of type `T` in general circumstances.
+
+**PROMISE**: Coalton `Integer`, `Character`, `String`,  `Single-Float`, and `Double-Float` values correspond to their Lisp counterparts.
+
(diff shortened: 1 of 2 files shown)


docs/intro-to-coalton.md
```diff
@@ -410,14 +428,9 @@ The following type classes are similar to the typeclasses with the same name in
 * `Applicative`
 * `Monad` - monad does not have `return`, use `pure` from applicative instead
 * `Alternative` - `<|>` is called `alt` in Coalton
-
-
-
```
eliaslfox, line 415: Those were load bearing newlines. `Into` and `TryInto` aren't in Haskell. There should probably be an explanation that those two type classes are analogous  to the traits with the same names in Rust.

docs/intro-to-coalton.md
```diff
@@ -446,7 +459,7 @@ Inline type annotations can be added to resolve ambiguities when using type clas
 
 ## Shorthand Function Call Syntax
 
-Coalton does not have nullary functions. However `(make-vector)` is a shorthand for `(make-vector Unit)`.
+Coalton does not have nullary functions. However, function with the type signature `Unit -> *` can be written without explicitly passing `Unit`. For instance, `(make-vector)` is a shorthand for `(make-vector Unit)`.
```
eliaslfox, line 462: Should this say "can be called without explicitly passing Unit" instead?

docs/coalton-lisp-interop.md
```diff
@@ -0,0 +1,121 @@
...
+
+**Todo**: `COALTON` operator
+
+## Coalton-Calls-Lisp Bridge
+
+**Todo**: `LISP` operator
+
+## `LISP-VALUE` Type
+
+**Todo**
```
eliaslfox, line 117: Lisp-Object isn't a type at runtime. It get's compiled to `(declare (type t) ...)`. See https://github.com/coalton-lang/coalton/blob/913870e09e2776aeac4f55fd9696cd616f34faca/src/typechecker/environment.lisp#L150-L156.

docs/coalton-lisp-interop.md
```diff
@@ -0,0 +1,132 @@
...
+For the most part, `define-type` will not provide guarantees across interaction modes. One reason to understand why is to consider the type
+
+```
+(define-type Wrapper
+  (Wrap Integer))
+```
+
+Depending on the interaction mode, `Wrap` may actually be similar to the `IDENTITY` function, and `Wrapper` may be representation-equivalent to `Integer`.
+
+Nonetheless, there is one practical guarantee: the existence of Lisp functions for unconstrained constructors. *Unconstrained* means that there are no constraints on any type variables present in the type that describes the value. Consider the following type:
```
colescott, line 60: I think here it might be useful to clarify that you are talking about typeclass constraints, possibly explain how they arise (and how you can check for them using `type-of`) and provide an example of a constrained value which is not allowed.

docs/coalton-lisp-interop.md
```diff
@@ -0,0 +1,132 @@
...
+## Interaction Mode
+
+First and foremost, there are two ways to globally compile Coalton. This is determined by an environment variable `COALTON_ENV` which in turns sets a global variable `*interaction-mode*`. There are two options for this variable:
+
+- `COALTON_ENV=development`, which sets `*interaction-mode*` to `:development`
+- `COALTON_ENV=release`, which sets `*interaction-mode*` to `:release`
+
+As the names imply, these two modes have different behaviors that optimize for either development or release.
+
+**WARNING**: *All* Coalton code, *including* the standard library, *must* be compiled in the same mode. We advise application writers to set the environment variable during their application's build process before Coalton is compiled, and to ensure caches are not stale.
```
colescott, line 16: Maybe we should provide some way to either allow interop between modes or to tag definitions with their mode and error when new code is compiled using a different mode

colescott (commented):
Looks great so far!

docs/coalton-lisp-interop.md
eliaslfox: > tag definitions with their mode and error when new code is compiled using a different mode

I like this. I think we could have a header in the generated code that checks the mode instead of tagging each definition

stylewarning: Added issue #206

stylewarning:
Does anyone know what this last
```
build Expected — Waiting for status to be reported 
```
is supposed to be? It looks hung there, but all `CI / build` matrix tests passed.

eliaslfox:
I think that comes from the initial pr settings. We had Github block prs until "build" completed. Adding test matrixes changed the name of the job(s).

eliaslfox:
This looks really great btw.

stylewarning:
> Also this looks really great!

@eliaslfox Anything I should add or change? I've re-requested reviews from you and @colescott

eliaslfox (approved):
Looks good

PR #200, opened 2021-11-07
Test for and fix issue 199

gefjon:
I reordered the definitions emitted by `codegen-program` so that instances come last, as they're the only things that eagerly read from variables at load time. I think. Reviewer: is that actually the case, or will other `coalton-codgen` `setf` forms read from Coalton variables at load time?

In any case, I built a test case which fails prior to the final commit in this PR, and passes after. I also wrote a slick macro in tests/utilities.lisp which automates writing forms into temporary files so they can be compiled as top-level forms.

cheers,
phoebe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
-.PHONY: test
+.PHONY: test test-safe
 test:
 	sbcl --non-interactive \
+	     --eval "(asdf:test-system :coalton)"
+
+test-safe:
+	sbcl --non-interactive \
+	     --eval "(sb-ext:restrict-compiler-policy 'safety 3)" \
 	     --eval "(asdf:test-system :coalton)"
 
 .PHONY: docs
--- a/coalton.asd
+++ b/coalton.asd
@@ -120,4 +120,5 @@
                (:file "tarjan-scc-tests")
                (:file "type-inference-tests")
                (:file "environment-persist-tests")
-               (:file "graph-tests")))
+               (:file "graph-tests")
+               (:file "letrec-order-tests")))
--- a/src/codegen/program.lisp
+++ b/src/codegen/program.lisp
@@ -100,10 +100,13 @@
 
        ;; Define ...
        ,@(reshuffle-definitions (append
-                                 ;; ... instance structs
-                                 (compile-instance-definitions instance-definitions optimizer)
                                  ;; ... functions and variables
-                                 (compile-toplevel-sccs bindings sccs env)))
+                                 (compile-toplevel-sccs bindings sccs env)
+                                 ;; ... instance structs
+                                 ;; unlike other definitions, instances may read eagerly
+                                 ;; from variables. so initialize them last, after all
+                                 ;; the variables have been defined.
+                                 (compile-instance-definitions instance-definitions optimizer)))
 
        ;; Emit documentation
        ,@(compile-docstring-forms docstrings)
--- a/tests/letrec-order-tests.lisp
+++ b/tests/letrec-order-tests.lisp
@@ -1,6 +1,6 @@
 (in-package #:coalton-tests)
 
-(deftest test-letrec-order ()
+(deftest test-def-before-instance ()
   "Test that bindings are initialized in a correct order so that constructing instance structs never reads an uninitialized variable.
 
 From coalton-lang/coalton issue #199"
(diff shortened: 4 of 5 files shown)


eliaslfox:
This looks great.  Unfortunately it is possible for a variable to eagerly read from an instance at load time. 

```
 (coalton-toplevel
  (define-type T T)
                
  (define-instance (Semigroup T)
    (define (<> a b) T))
                
  (define-instance (Monoid T)
    (define mempty T))
                
  (declare x T)
  (define x mempty))
```

This works on `main` but is broken in this pr. We get instances of invalid codegen with either ordering, and I think that having the standard library compiling on safety 3 sounds good. So this should be merged anyway.

Long term codegen needs to compute SCCs between definitions and instances and then output them in the correct order. It will also be necessary to detect cycles between immediately evaluated forms. A simplified version of this cycle checking is already done for variables in https://github.com/coalton-lang/coalton/blob/main/src/typechecker/derive-type.lisp#L315-L322. 

Detecting dependencies between definitions and instances has two challenges

* The current SCC code works on definitions only. The SCC of an instance should be the highest SCC containing one of its members.
* Instances are currently computed during codegen. Ordering will need to take into account not only instances but also the dependencies of those instances.

```
 (coalton-toplevel
  (define-type T T)
                
  (define-type (T2 :a) (T2 :a))
                
  (define-instance (Semigroup T)
    (define (<> a b) T))
                
  (define-instance (Monoid T)
    (define mempty T))
                
  (define-instance (Semigroup :a => (Semigroup (T2 :a)))
    (define (<> a b)
      (match (Tuple a b)
        ((Tuple (T2 a) (T2 b)) (T2 (<> a b))))))
                
  (define-instance (Monoid :a => (Monoid (T2 :a)))
    (define mempty (T2 mempty)))
                
  (declare x (T2 T))
  (define x mempty))
```

Note that `x` is dependent not only on the instance `Monoid T2` but also on all three previous instances. This leads me to believe that Coalton should have a second typed AST with resolved instances. This would allow computing dependencies as well as make some instance dict deduplication optimizations possible.

Also does it make sense to have both test and test-safe? Or should test always run on safety 3?

gefjon:
> Also does it make sense to have both test and test-safe? Or should test always run on safety 3?

I think it's probably valuable to have both: `safety 3` will detect some bugs, like this one, but it can also potentially make some other bugs disappear. For example, there's been some discussion on sb-devel about whether arrays will be implicitly zero-initialized at low safety, or left holding garbage values.

I tried to put together a PR that would run CI tests both at `safety 3` and at the default, but I couldn't figure out how GitHub Actions work. One of y'all could probably make that happen, by adding a CI job that runs the `sb-ext:restrict-compiler-policy` form before compiling Coalton.

> This works on main but is broken in this pr. We get instances of invalid codegen with either ordering

Ah... If the "fix" part of this PR doesn't actually solve the problem, you probably shouldn't merge it. Darn.

I have a potential fix in mind that shouldn't need a new IR, I think:

If a `define-instance Into` form currently expands into code equivalent to:

```
(define-global MY-INTO-INSTANCE :uninit)

(setf MY-INTO-INSTANCE (make-instance 'into :into MY-IMPLEMENTATION))
```

which then gets interleaved with other definitions, it should be possible to replace that with:

```
(define-global MY-INTO-INSTANCE (make-instance 'into))
;; MY-INTO-INSTANCE is an instance of INTO, but its slots are unbound. reads from
;; the variable MY-INTO-INSTANCE will see the correct object, but uses of that object
;; will probably fail.

(setf (slot-value MY-INTO-INSTANCE 'into) MY-IMPLEMENTATION)
;; now uses of the object from MY-INTO-INSTANCE will succeed
```

This will not work if Coalton inlines / eagerly evalutates reads from instances, though. That is, if Coalton tries to replace `(funcall (slot-value MY-INTO-INSTANCE 'into) ...)` with `(funcall (load-time-value (slot-value MY-INTO-INSTANCE 'into)) ...)`, my solution won't work.

eliaslfox:
>  One of y'all could probably make that happen, by adding a CI job that runs the sb-ext:restrict-compiler-policy form before compiling Coalton.

I can work on this.

gefjon:
I'd like to hear confirmation from @stylewarning (or whoever) that my proposed solution will work, or at least improve the situation, before I commit time to implementing it. But if it will work, I'd be happy to write the code.

gefjon:
@eliaslfox could you change your workflow in .github/workflows/main.yml so that the "low-safety" run is with `0`, not `1`? Even though the default `safety` is 1, `sb-ext:restrict-compiler-policy` sets a lower bound (and an upper bound, if supplied), so the parameter for "leave lower bound unchanged" is 0.

eliaslfox:
@gefjon Yeah I can do that

stylewarning:
@gefjon I like your idea for binding the implementation late a lot more than what currently happens. At least the types appear to be correct, and the objects get into the right places, even if the implementations aren't filled in.

With that said, I still think we'll have issues because we aren't "coordinated" about what needs to exist when. Honestly, it's not clear to me, and I need to spend time sitting down and writing out what needs to happen. I think we need to systematically work through what ought to get evaluated when in all circumstances, including a policy for top-level values. Putting everything inside of `EVAL-WHEN`s has felt wrong, dirty, and unlispy. (I never have this issue when I'm writing "normal" LIsp code, so I don't think there should be an issue when I'm writing normal Coalton code either.)

In all of these, evaluation of top-level values at compile-time seems to be the killer. Doubly so if there's mutual dependence/recursion.

gefjon:
Closing because this doesn't fix the issue. I tested it locally, and my proposal also didn't fix it. I've been told that the actual solution is to treat the top-level definitions as a directed graph, compute its strongly connected components, treat each one as a letrec, and do a topological sort of said components to determine evaluation order. I do not have time to implement such a scheme right now.

Issue #199, opened 2021-11-06
Build failure: :|@@unbound@@| is not of type (OR COALTON-IMPL/CODEGEN::FUNCTION-ENTRY (FUNCTION (T) *))


gefjon:
When compiling Coalton on SBCL 2.1.10.35-669281695 on MacOS 12.0.1 Arm64, I get the following error first from string.lisp, and then again from cell.lisp if I invoke the ACCEPT restart on the string one:

```
The value
  :|@@unbound@@|
is not of type
  (OR COALTON-IMPL/CODEGEN::FUNCTION-ENTRY (FUNCTION (T) *))
   [Condition of type COMMON-LISP:TYPE-ERROR]
```

I did a bit of playing around with string.lisp in the REPL, isolating forms into individual `coalton-toplevel` blocks, and the error seems to occur if and only if `String`'s `Monoid` and `Into` instances are in the same `coalton-toplevel` form. That is, the following change seems to prevent the error:

```
modified   src/library/string.lisp
@@ -64,17 +64,18 @@
   (define-instance (Ord String)
     (define (<=> a b)
       (lisp Ord (a b)
-         (cl:cond
-           ((cl:string> a b) GT)
-           ((cl:string< a b) LT)
-           (cl:t EQ)))))
+        (cl:cond
+          ((cl:string> a b) GT)
+          ((cl:string< a b) LT)
+          (cl:t EQ)))))
 
   (define-instance (Semigroup String)
     (define (<> a b) (concat-string a b)))
 
   (define-instance (Monoid String)
-    (define mempty ""))
+    (define mempty "")))
 
+(coalton-toplevel
   (define-instance (Into String (List Char))
     (define into unpack-string))
```

cheers,
phoebe

eliaslfox:
Hi @gefjon,
Thanks for the bug report. Are you using the latest quicklisp distribution?

gefjon:
I have
```
(ql:dist-version "quicklisp")
"2021-10-21"
```
Also, of possible note, my `sbclrc` does:
```
(proclaim '(optimize (debug 3) (speed 3) (safety 3) (space 1) (compilation-speed 1)))
```

eliaslfox:
Adding `(proclaim '(optimize (debug 3) (speed 3) (safety 3) (space 1) (compilation-speed 1)))` to my `sbclrc` breaks loading Coalton.

gefjon:
[...]

PR #201, opened 2021-11-08
Matrix tests across safety 1 and 3

eliaslfox:
Add tests for safety 3 to detect issues like #199.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,8 +13,10 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


eliaslfox:
These should be failing on safety 3. I'm not sure why they're failing on safety 1.

stylewarning:
Will this have strange load-order or caching behavior?

I'm not sure if the compiler policy is being restricted early enough. Is Coalton getting loaded before the policy is being restricted?

eliaslfox:
It might. I think I need to look more deeply into how the CI pipeline is configured.

eliaslfox:
@stylewarning this is ready for review

stylewarning (approved):

PR #210, opened 2021-11-30
Add vector sorting

eliaslfox:


src/library/vector.lisp
```diff
@@ -196,6 +205,25 @@
...
+      ((Vector v)
+       (progn
+         (lisp :a (v f)
+           (cl:sort
+            v
+            (cl:lambda (a b)
+              (coalton-impl/codegen::A2 f a b))))
+         Unit))))
+
+  (declare vector-sort (Ord :a => ((Vector :a) -> Unit)))
```
stylewarning, line 221: Do we want to use `!` or something to indicate this? That's what Scheme does.

stylewarning (approved):
Other than my suggestion about in-place operations, LGTM. (Decide about naming, then rebase, then I'll merge.)

eliaslfox:
I don't want to use `!` as a mutation suffix. Vector has lots of other functions that mutate already. And the return type of unit make it clear that this function mutates.

PR #217, opened 2021-12-01
Fix stack allocation warnings in development

eliaslfox:
I accidently disabled the wrong optimization in development mode.


colescott (approved):
lol rip. good catch

PR #222, opened 2021-12-01
Add more zip functions

eliaslfox:


stylewarning (approved):
At some point I can imagine we will MACROLET this when we want 64 tuples.

eliaslfox:
Yeah definitely. I also would love to have reader macro syntax for tuples. Something like `#[:a :b :c]` for types and `#[1 "hello" Nil]` for values.

PR #230, opened 2021-12-03
Conveniences found by AoC

stylewarning:
Small, miscellaneous conveniences will be collected here.

### DIFF
--- a/src/ast/parse-form.lisp
+++ b/src/ast/parse-form.lisp
@@ -61,7 +61,9 @@ This does not attempt to do any sort of analysis whatsoever. It is suitable for 
           ((null rands)
            (parse-application expr (first expr) `(coalton-library:Unit) m package))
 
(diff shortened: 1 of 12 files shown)


src/library/more-instances.lisp
```diff
@@ -0,0 +1,13 @@
+(in-package #:coalton-library)
+
+(coalton-toplevel
```
eliaslfox, line 3: Is there a reason this can't be in classes.lisp?

stylewarning: `not` wasn't defined earlier enough. Now it's a non-problem.

stylewarning:
The commit history has been cleaned and made sort of nice.

This PR adds functionality, but also fixes lots of little tech debt things (like ~5 `STYLE-WARNING`s)

The build failure looks like a failure to download SBCL... tests pass locally for me...

eliaslfox (approved):
Looks good

PR #231, opened 2021-12-03
Remove Show

eliaslfox:
The Show type class is almost entirely unused because values can already
be printed in the REPL. The definition of Show is slow because it
requires repeatedly concatenating strings in memory.

In the future other type classes for printing structures should be
explored. They should use string streams or another way of only building
a complete string once.

Also multiple printing classes similar to Debug and Display in Rust
should be explored.


stylewarning (approved):

stylewarning:
@eliaslfox ruhbase

PR #235, opened 2021-12-04
Fix definition of tail function

eliaslfox:


eliaslfox:
This might be a mistake.

PR #236, opened 2021-12-04
Add last and init functions.  Add list of Coalton quirks to the intro doc.

HM0880:


docs/intro-to-coalton.md
```diff
@@ -486,3 +486,9 @@ The following functions all take an optional package parameter
 * `print-value-db` - Print the type of every toplevel value
 * `print-class-db` - Print every class and their methods
 * `print-instance-db` - Print the instances of every class
+
+
+## Coalton quirks
```
stylewarning, line 491: "Quirks and Differences from Common Lisp"

docs/intro-to-coalton.md
```diff
@@ -486,3 +486,9 @@ The following functions all take an optional package parameter
...
+
+- Coalton uses `True` (not `t`).
```
stylewarning, line 493: "As such, `t` can be used as an ordinary variable name."

docs/intro-to-coalton.md
```diff
@@ -486,3 +486,9 @@ The following functions all take an optional package parameter
 * `print-value-db` - Print the type of every toplevel value
 * `print-class-db` - Print every class and their methods
 * `print-instance-db` - Print the instances of every class
+
+
+## Coalton quirks
+
```
stylewarning, line 492: "Lists in Coalton must be homogeneous."

"Coalton uses `fn`, *not* `lambda`, to denote anonymous functions."

"Numerical operators like `+` only take 2 arguments."

"Negation is done with `negate`. The form `(- x)` is a curried function equivalent to `(fn (z) (- x z))`."

src/library/list.lisp
```diff
@@ -15,13 +15,26 @@
...
+    "Returns the last element of a list."
+    (match l
+      ((Cons x (Nil)) (Some x))
+      ((Cons _ xs) (last xs))
       ((Nil) None)))
 
+  (declare init ((List :a) -> (List :a)))
+  (define (init l)
+    "Returns every element except the last in a list."
+    (reverse (tail (reverse l))))
```
stylewarning, line 36: CL:BUTLAST could also be used

stylewarning (changes requested):
Small requests and additions

src/library/list.lisp
HM0880: Should I use CL:BUTLAST here?  I have no preference.

stylewarning: It will be about 2x as efficient, but it doesn't matter. (It'll just make one copy, not two.

stylewarning (approved):

PR #242, opened 2021-12-05
Replace Coalton:List with CL:List

### LINKED ISSUE #144: Replace `COALTON-LIBRARY:LIST` with `CL:LIST`
Replace the list definition in the standard library with a primitive type that is a standard lisp list under the hood. This should improve performance and allow using standard lisp functions for lisp manipulation.

eliaslfox:
Also remove functions for converting between Coalton lists and lisp
lists. Fixes #144

### DIFF
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -191,19 +191,7 @@ Numbers implement the `Num` typeclass, which has methods `+`, `-`, `*`, and `fro
 
 ## Lists
 
-Coalton has a list data type defined as
-
-```lisp
-(coalton-toplevel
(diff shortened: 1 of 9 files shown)


colescott:
Amazing work! 

The only thing I think might need to be fixed is `intro-to-coalton.md` needs to be updated to say that coalton lists _are_ CL lists and the reference generator should be updated to include early types as part of its output.

eliaslfox:
I updated the intro document. I would prefer to not block this pr on the documentation generator.

colescott:
I agree. Looks good now!

colescott (approved):

PR #247, opened 2021-12-05
Rename WithDefault -> Unwrappable

eliaslfox:
- add unwrap method


stylewarning (approved):
tbh I read it as "un-wrappable" and not "unwrap-able"

eliaslfox:
Yeah the name isn't perfect.

PR #254, opened 2021-12-06
Make to-boolean handle proper generalized booleans

### LINKED ISSUE #253: to-boolean only works on values that are already strictly boolean
`to-boolean`'s docstring says it converts a Lisp generalized boolean to a Coalton boolean. Based on [the definition of "generalized boolean"](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_g.htm) in the ANSI spec, `to-boolean` should therefore work on any input value, but it only works on Lisp booleans. Since Coalton booleans and Lisp booleans are defined the same way right now, `to-boolean` is basically a no-op.

I should be able to write, for instance,

```
(coalton-toplevel
  (define (my-member e xs)
    (lisp Boolean (e xs)
      (to-boolean (cl:member e xs)))))
```

since `cl:member` [returns `cl:nil` on failure and a nonempty list on success](http://www.lispworks.com/documentation/HyperSpec/Body/f_mem_m.htm). This isn't possible yet.

The codebase already uses `to-boolean` in a few places, but all the expressions it's used on are calls to functions that return Lisp booleans anyway, so it works by sheer happenstance.

Culprit: the current definition of `to-boolean` uses `cl:the` as though it were a typecast, whereas it's really [an assertion that the argument is already of a given type](http://www.lispworks.com/documentation/HyperSpec/Body/s_the.htm).

danielkoning:
That is, Lisp type T. Fixes #253.

### DIFF
--- a/src/library/macros.lisp
+++ b/src/library/macros.lisp
@@ -91,10 +91,9 @@ to
     (list-helper forms)))
 
 (cl:defmacro to-boolean (expr)
-  "Convert generalized boolean lisp expression EXPR to coalton boolean"
-  `(cl:if (cl:the cl:boolean ,expr)
-          True
-          False))
+  "Convert the Lisp expression EXPR, representing a generalized boolean, to a
+Coalton boolean."
+  `(cl:not (cl:eql ,expr cl:nil)))
 
 (cl:defmacro do (cl:&rest forms)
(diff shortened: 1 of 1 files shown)


eliaslfox (approved):
Thanks for the fix

PR #263, opened 2021-12-07
remove unsafe/, add Quantizable class, add division convenience functions

stylewarning:
This commit adds some functionality to make division more convenient.

First, it removes `unsafe/` which was nice in theory but too cumbersome in practice. Now `/` may signal an error at run-time. `safe/` is provided if needed.

Second, it adds a Quantizable class for rounding number types to integers. This is used to implement floor, ceiling, and round in a generic way.

Third, some convenience functions are added for supporting integer and float division. These are common, monomorphic functions that make use of the generic machinery.

Fourth, the introduction is improved to discuss division and how to work with it.

Fifth, the reference was updated.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -494,11 +494,11 @@ establishes that division of two `Single-Float`s can result in a `Single-Float`.
 
 Note that `Dividable` does *not* establish a default result type; you must constrain the result type yourself.
 
-See also: `/`
+The function / is partial, and will error produce a run-time error if the divisor is zero.
 
 
 Methods:
-- `UNSAFE/ :: (:A → :A → :B)`
+- `/ :: (:A → :A → :B)`
 
 <details>
 <summary>Instances</summary>
@@ -764,7 +764,7 @@ Is N even?
 #### `EXPT` <sup><sub>[FUNCTION]</sub></sup><a name="EXPT"></a>
 `(INTEGER → INTEGER → INTEGER)`
 
-Exponentiate BASE to the POWER.
+Exponentiate BASE to a non-negative POWER.
 
 
 ***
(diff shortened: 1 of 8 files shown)


eliaslfox:
Can we have integer divison?

stylewarning:
> Can we have integer divison? 

I don't think there's any "safe" (in the political sense) choice; every language disagrees on what the default ought to be. Some languages always floor, others always round, etc. Common Lisp always produces a rational number and you must be explicit about what you want.

stylewarning:
Just as an example, see [this comparison between C and Python](https://www.evanjones.ca/integer-division.html), but take the author's opining about "mathy languages" etc. with a grain of salt.

eliaslfox:
That makes sense. I didn't realize different languages implemented divison differently :/

stylewarning:
> That makes sense. I didn't realize different languages implemented divison differently :/

Huge annoyance when porting or QUOTE "transpiling" UNQUOTE code.

docs/reference.md
```diff
@@ -1339,6 +1405,101 @@ Print a line to *STANDARD-OUTPUT* in the form "{STR}: {ITEM}"
...
+
+Return the least integer greater than or equal to X.
+
+
+***
+
+#### `DOUBLE/` <sup><sub>[FUNCTION]</sub></sup><a name="DOUBLE/"></a>
+`(DOUBLE-FLOAT → DOUBLE-FLOAT → DOUBLE-FLOAT)`
+
+Compute the quotient of single-precision floats A and B as a single-precision float.
```
eliaslfox, line 1471: double-precision floats

eliaslfox (approved):
This looks good

PR #264, opened 2021-12-07
fix expt which could have returned rationals

stylewarning:
`expt` could return a rational if the power was negative. this fixes that

also catches a div by 0 in `mod`


stylewarning:
(This demonstrates the dangers of lisp bindings.)

eliaslfox (approved):
👍

PR #268, opened 2021-12-08
Add ecl support

eliaslfox:


stylewarning:
19 minute builds... Maybe we should support CCL instead? :/

ECL also doesn't have tail call elimination, does it?

eliaslfox:
I'll look into CCL

PR #272, opened 2021-12-08
Remove conversion functions superceeded by into

eliaslfox:


colescott (approved):

stylewarning:
I'm not sure I'm a fan of this change. I'm fine implementing monomorphic variants with polymorphism, but now we've made something that was simple and understandable for common conversion cases into something that's harder to search for, discover, and (in some cases) type without writing type declarations.

We shouldn't monomorphize _everything_, but if it's common, why not?

The benefit though is this "forces" the use of more abstract type system features, where we know bugs lie.

PR #281, opened 2021-12-10
Add `enable-coalton-toplevel-file`

colescott:
This adds a function `coalton:enable-coalton-toplevel-file` which enables reader macros to allow all toplevel coalton forms to show up without needing a `coalton-toplevel`.

`named-readtables` was added as a dependency to allow SLIME to use this read table.

Currently working:
- Toplevel coalton forms in file without coalton-toplevel
- Non-toplevel coalton forms are left alone
- `C-c C-l` works
- `C-c C-k` works
- `C-c C-c` works
- `compile-file` works

Known issues:
- ~`C-c C-k` does not work without first running `C-c C-l` or `C-c C-c` on the `enable-coalton-toplevel-file` form~ This was fixed in 7b7c3cf

An example of this in action:
```lisp
(cl:in-package #:coalton-user)

;; Enable the read table
(coalton:enable-coalton-toplevel-file)

;; Recursive definitions work
(define-type Red
  (Red Black))
(define-type Black
  (Black Red))

(define (f x)
  (+ 1 x))
(define (g y)
  (+ 2 (f (+ 3 y))))

;; Common lisp can be embedded seamlessly
(cl:defvar lol (f 5))

;; back to coalton
(define x 6)
```

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -11,7 +11,8 @@
                #:fset
                #:float-features
                #:split-sequence
-               #:uiop)
+               #:uiop
+               #:named-readtables)
   :in-order-to ((asdf:test-op (asdf:test-op #:coalton/tests)))
   :around-compile (lambda (compile)
                     (let (#+sbcl (sb-ext:*derive-function-types* t))
--- a/src/debug.lisp
+++ b/src/debug.lisp
@@ -137,6 +137,6 @@ and wrap in a COALTON-TOPLEVEL block."
   (:merge :standard)
   (:macro-char #\( #'read-coalton-toplevel-form))
 
-(defun coalton:enable-coalton-toplevel-file ()
+(defmacro coalton:enable-coalton-toplevel-file ()
   "Enables the coalton readtable to read the file as a coalton toplevel form."
-  (named-readtables:in-readtable coalton:syntax))
+  `(named-readtables:in-readtable coalton:syntax))
--- a/src/debug.lisp
+++ b/src/debug.lisp
@@ -23,3 +23,120 @@
 (defun coalton:kind-of (symbol)
   "Lookup the kind of type SYMBOL in the global environment"
   (coalton-impl/typechecker::kind-of (coalton-impl/typechecker::type-entry-type (coalton-impl::lookup-type *global-environment* symbol))))
+
+;;
+;; Read table shenanigans below. You have been warned!
+;;
+
+;; Depth counter for read-coalton-toplevel-form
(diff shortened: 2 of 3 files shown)


gefjon:
I do not think this is the right way to do this.

As I understand your goals, they are:

- compile and evaluate top-level forms as Coalton forms in two contexts:
  - when using `asdf` to `compile-file`
  - when using emacs with `sly` or `slime` to compile/load/eval the region/buffer

I would very much like a more general solution, which would do all of the following:
- compile and evaluate top-level forms as Coalton forms in the above two contexts
  - this requires separate code for asdf and sly/slime, but not very much of either
- treat Coalton symbols as case-sensitive
  -  and Common Lisp symbols as always all-caps
- allow Coalton to have its own reader macros, e.g. to treat `#(foo)`as a Coalton vector rather than a read-time error

The short version of my proposal is:

- define a function `coalton:compile-file` which:
  - binds dynamic variables e.g. `*package*` and `*readtable*` as appropriate
  - reads, macroexpands and compiles a `.coalton` (? think about ext) file
  - produces a `.fasl` file, and/or loads said file into the current lisp image
    - this might (or might not) involve producing a macroexpanded `.lisp` file first
- define an asdf component (subclassing `lisp-source-file`) which arranges to call `coalton:compile-file`

I'm too busy tonight to write a serious pros/cons list for my proposal, and too busy during this semester's finals to implement anything. But I believe it would be pretty easy to implement the thing I describe.

colescott:
After some discussion with @eliaslfox, I think I agree with @gefjon on this.

This PR was an initial attempt at implementing this without much work but I do agree that the `coalton:compile-file`/asdf/slime route is probably better in the long term for the language. This would allow for much better debugging position tracking which was the primary end-goal of me playing with read tables.

In addition to the native vector syntax, this might be a good time to introduce some other abbreviated syntax which would otherwise not be allowed by the lisp reader (such as @eliaslfox 's `.1` to read the first element of a sequence). It would also be good to take this opportunity to figure out if we want to make coalton case-insensitive and think about what that means for Common Lisp interop.

We also need to consider if we want to continue supporting embedding coalton inside lisp files (I think this is a good thing to have) and what features would still work within these forms.

It seems like the `coalton:compile-file`/asdf/slime route will require some more effort but I am happy to look in to it and see if it is something that can be done in the next couple of weeks.

colescott:
Converted this to a draft so we know not to merge without more discussion.

PR #283, opened 2021-12-11
Fix type-variables on environment

### LINKED ISSUE #249: Incorrect type inference
```
                (define (f a)
                   (let ((g (fn (x) (Tuple x a))))
                     (g 5))))

COALTON-USER> (type-of 'f)
∀ :A :B. (:A → (TUPLE INTEGER :B))
```

eliaslfox:
Previously fset:domain was used to get the keys in a map. Fset:domain builds a set, which deduplicates values. Fset:compare was defined to `type-scheme=` which treated `forall. a` and `forall. b` as equal.

Fixes #249

### DIFF
--- a/src/typechecker/environment.lisp
+++ b/src/typechecker/environment.lisp
@@ -15,7 +15,11 @@
                 (immutable-map-data env))))
 
 (defmethod type-variables ((env value-environment))
-  (remove-duplicates (mapcan #'type-variables (fset:convert 'list (fset:range (immutable-map-data env)))) :test #'equalp))
+  (let ((out nil))
+    (fset:do-map (name type (immutable-map-data env))
+      (declare (ignore name))
+      (setf out (append (type-variables type) out)))
+    (remove-duplicates out :test #'equalp)))
 
 ;;;
 ;;; Type environments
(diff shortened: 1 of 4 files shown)


stylewarning:
Nice sleuthing!

src/typechecker/environment.lisp
```diff
@@ -15,7 +15,11 @@
                 (immutable-map-data env))))
 
 (defmethod type-variables ((env value-environment))
-  (remove-duplicates (mapcan #'type-variables (fset:convert 'list (fset:range (immutable-map-data env)))) :test #'equalp))
+  (let ((out nil))
+    (fset:do-map (name type (immutable-map-data env))
+      (declare (ignore name))
+      (setf out (append (type-variables type) out)))
```
stylewarning, line 21: PUSH

src/typechecker/environment.lisp
```diff
@@ -15,7 +15,11 @@
...
+    (remove-duplicates out :test #'equalp)))
```
stylewarning, line 22: Any chance at a more specific equality predicate?

stylewarning (changes requested):
These changes aren't a deal breaker and would be fine to approve.

src/typechecker/environment.lisp
eliaslfox: They aren't equivalent. `type-variables` returns a list.

stylewarning: i'm dumb

src/typechecker/environment.lisp
eliaslfox: We don't have one defined, and equalp is used everywhere

stylewarning (approved):

PR #292, opened 2021-12-11
Allow `progn` at top level

simendsjo:
This enables macros that define several toplevel forms.


stylewarning:
Can we just re-use `COALTON-TOPLEVEL`? Does that work?

eliaslfox:
I think it's better to have macros generate their own calls to coalton-toplevel. We do that already in the standard library.

simendsjo:
Using `coalton-toplevel` is probably better then even though it might trip cl users.

Issue #273, opened 2021-12-08
Failed to reduce context in class definition


eliaslfox:
```
(coalton-toplevel
    (define-class ((Into Integer :a) (Eq :a) => (Num :a))
    "Types which have numeric operations defined."
    (+ (:a -> :a -> :a))
    (- (:a -> :a -> :a))
    (* (:a -> :a -> :a))
    (fromInt (Integer -> :a))))
```

Gives the following error

```
Failed to reduce context for INTO INTEGER #T43240311
in method +
in class definition (DEFINE-CLASS ((INTO INTEGER A) (EQ A) => (NUM A))
                   Types which have numeric operations defined.
                   (+ (A -> A -> A))
                   (- (A -> A -> A))
                   (* (A -> A -> A))
                   (FROMINT (INTEGER -> A)))
```

PR #293, opened 2021-12-11
Make reduce-context aware of current substitutions

### LINKED ISSUE #69: (fromSome "" (vector-index (into idx) ...))

colescott:
This fixes #69 and #273 which were caused by context reduction bugs. `reduce-context` now applies substitutions internally and will no longer signal a context reduction error in cases where the predicates should be deferred to the enclosing context. Additionally, explicit type annotations with `the` had a check for predicates which is no longer valid and has been removed.


eliaslfox (approved):
Looks good

PR #301, opened 2021-12-12
add LET*, add docs around variable binding

stylewarning:
This adds `let*`, which is a "more traditional" `let`-like construct that people are used to.

Also documents this `let` v `let*` business.


colescott:
I guess I'm not entirely sure what the use case for having non-recursive let bindings is... Do you have an example of the usage?

I think with a lot of these new language features we are adding as macros it is going to get harder for the Co. alton compiler to give meaningful error messages. @eliaslfox and I have discussed moving several of the existing macros into the compiler for this reason and I think this might also be a good case for that if we decide to include it.
I'm also a bit concerned that `let` and `let*` have opposite effects of their counterparts in Common Lisp.

PR #303, opened 2021-12-12
add Symbol type

stylewarning:


colescott:
What is the use-case for having Coalton-accessible symbols? I feel like they end up being just another wrapper over strings and I'm not sure if it adds much type safety.

(As a note, the current implementation results in boxed symbols in debug mode which I do not think is what we want.)

PR #305, opened 2021-12-13
Refactor toplevel codewalker

danielkoning:
The entry point to the compiler, COLLECT-TOPLEVEL-FORMS, seems to have grown in interface complexity to the point that it’s difficult to keep straight. At this point, the function claims to return three values, lists five, and actually returns six!

These commits go a long way toward detangling and simplifying the code, mainly by attaching information about Coalton operators to the operator symbols themselves. It now reads pretty much top to bottom without recursing or returning from the middle. A new macro, INSTALL-OPERATOR-METADATA, makes this technique self-documenting and easily extensible with new operators or constraints. COLLECT-TOPLEVEL-FORMS now returns a single plist instead of six values, eliminating the demand on the caller to handle the results positionally (and allowing the function’s behavior to change further while keeping its interface frozen).

Also included is a new test suite to make sure the rewritten codewalker does what it promises.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -121,6 +121,7 @@
   :serial t
   :components ((:file "package")
                (:file "utilities")
+               (:file "syntax-tests")
                (:file "free-variables-tests")
                (:file "tarjan-scc-tests")
                (:file "type-inference-tests")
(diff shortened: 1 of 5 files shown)


colescott (approved):
Thanks! This is definitely a lot cleaner than the existing implementation :smile:

PR #306, opened 2021-12-13
[WIP] Add Lazy evaluated value

simendsjo:


src/library/lazy.lisp
```diff
@@ -0,0 +1,57 @@
...
+      (lazy (into (f (into ma))))))
+
+  ;; TODO: Into should be implemented for all Applicative by using pure
+  (define-instance (Into :a (Lazy :a))
+    (define into pure))
+
+  (define-instance (Into (Lazy :a) :a)
+    (define into force))
+
+  (declare force ((Lazy :a) -> :a))
```
colescott, line 39: This should be called `lazy-force`.

src/library/lazy.lisp
```diff
@@ -0,0 +1,57 @@
+(in-package #:coalton-library)
+
+(cl:defmacro lazy (body)
```
colescott, line 3: This doesn't need to be a macro... A Coalton function would probably be better suited so it can be typechecked

src/library/lazy.lisp
```diff
@@ -0,0 +1,57 @@
...
+    (define (pure v)
+      (LazyCell (make-cell (Value v))))
+    (define (liftA2 f fa fb)
+      (lazy (f (into fa) (into fb)))))
+
+  (define-instance (Monad Lazy)
+    (define (>>= ma f)
+      (lazy (into (f (into ma))))))
+
+  ;; TODO: Into should be implemented for all Applicative by using pure
```
colescott, line 32: Please remove this comment. Also due to overlapping instance rules, I don't think this is possible.

colescott (changes requested):
This looks like a great addition to the standard library. Aside from the other comments, the documentation should be rebuilt using `make cleanall && make docs`

src/library/lazy.lisp
simendsjo: If it's not a macro, the user has to wrap it in a function themselves to avoid evaluating the expression, which feels a bit unergonomic.

src/library/lazy.lisp
simendsjo: My thought was that an instance can be written for all Applicative because they can use `pure`

src/package.lisp
```diff
@@ -657,6 +657,11 @@
    #:cell-swap
    #:cell-write
    #:cell-update)
+  ;; Lazy
+  (:export
+   #:Lazy
```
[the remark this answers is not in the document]
simendsjo: Does that mean I cannot both export the type Lazy and the macro lazy? Or is both automatically exported?

simendsjo:
> This looks like a great addition to the standard library. Aside from the other comments, the documentation should be rebuilt using `make cleanall && make docs`

Looks like `make docs` reorderes some stuff, so it's a big messy diff. Cannot see the documentation for lazy here either.

src/library/lazy.lisp
colescott: Unfortunately with the current multi-parameter typeclass system it is not possible to define instances on non-concrete types since it will cause an issue with instance overlapping with the other define instances. This is because the instance overlapping system sees this as `Into :a (:b :a)` which will overlap any other instance with that shape.

colescott:
> > This looks like a great addition to the standard library. Aside from the other comments, the documentation should be rebuilt using `make cleanall && make docs`
> 
> Looks like `make docs` reorderes some stuff, so it's a big messy diff. Cannot see the documentation for lazy here either.

There is a known issue with the doc generator reordering. Running `make cleanall` before `make docs` is the workaround to avoid it.

src/package.lisp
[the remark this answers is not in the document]
colescott: They are both exported when you list one because they use the same symbol

src/library/lazy.lisp
colescott: I would prefer not to have macros which automatically wrap code in lambdas because it obscures control flow. I also think we should keep macros to a minimum because they make Coalton's errors more difficult to understand

simendsjo:
>

simendsjo:
> There is a known issue with the doc generator reordering. Running `make cleanall` before `make docs` is the workaround to avoid it.

Looks like it changes some existing documentation (created another PR), but it doesn't create any documentation for these types. Do I need to register it somewhere?

src/library/lazy.lisp
simendsjo: Too bad, could have removed some boilerplate.

colescott:
> Looks like it changes some existing documentation (created another PR), but it doesn't create any documentation for these types. Do I need to register it somewhere?

It looks like `lazy.lisp` isn't in the `coalton.asd` file so it won't be loaded. After that is added the doc generator should be able to pick it up

src/library/lazy.lisp
simendsjo: > I would prefer not to have macros which automatically wrap code in lambdas because it obscures control flow. I also think we should keep macros to a minimum because they make Coalton's errors more difficult to understand

I agree on both points, but I think `lazy` is a good use-case. Supplying an anonymous function is boilerplate and expose implementation details. `lazy` as a language-level construct should already be familiar to people coming from scheme (`delay`) and F# (`lazy`), and probably others (guess `clazy` for CL does the same thing). C# needs to supply a lambda to construct a `Lazy<T>`.

simendsjo:
> Cannot see the documentation for lazy here either.

Never mind. Didn't include the file in the system.

src/library/lazy.lisp
simendsjo: > A Coalton function would probably be better suited so it can be typechecked

The typechecker still gives acceptable error messages

```lisp
  ;; Failed to unify types STRING and INTEGER
  ;; in unification of types (LAZY STRING) and (LAZY INTEGER)
  ;; in definition of X

  (declare x (Lazy String))
  (define x (lazy 1))
```

stylewarning:
I'm going to close this for now. I think laziness is a topic that we should revisit, but I don't think it's time to put this in now.

PR #309, opened 2021-12-14
Run `make docs`

simendsjo:
Looks like someone forgot to run it.


colescott (approved):
Good catch. 

I think in the future it would be nice to have a github action that automatically builds the docs and publishes them to the coalton-lang website and we can then remove the markdown version to avoid desync

Issue #304, opened 2021-12-12
Bad error message + invalid return value for non-instantiated class/type combo


stylewarning:
Given a class and a type (that does *NOT* implement the class):

```
(coalton-toplevel
  (define-class (Cls :a)
    (c (:a)))
  (define-type Bar Bar))
```

At the REPL:

```
COALTON-USER> (coalton (the Bar c))
Expression (THE BAR C)
    of type ∀ :A. (CLS BAR) => BAR
    has unresolved constraint (CLS BAR)
    add a type assertion with THE to resolve it
COMMON-LISP:NIL
```

There is no error, just this print-out and returning `nil`. I can't imagine this is intended.

stylewarning:
Putting this in a coalton-toplevel:

```
  (define (baz b)
    (progn
      (the Bar c)
      Unit))
```

and compiling gives

```
; processing (COALTON-TOPLEVEL (DEFINE-CLASS # ...) ...)

; file: /tmp/slimeuQE1yj
; in:
;      COALTON-TOPLEVEL (DEFINE-CLASS (CLS :A)
;                     (C (:A)))
;     (COALTON:COALTON-TOPLEVEL
;       (COALTON:DEFINE-CLASS (COALTON-USER::CLS :A)
;         (COALTON-USER::C (:A)))
;       (COALTON:DEFINE-TYPE COALTON-USER::BAR
;         COALTON-USER::BAR)
;       (COALTON:DECLARE COALTON-USER::BAZ
;                        (COALTON-USER::BAR COALTON:-> COALTON:UNIT))
;       (COALTON:DEFINE (COALTON-USER::BAZ COALTON-USER::B)
;         (COALTON-LIBRARY:PROGN (COALTON:THE COALTON-USER::BAR COALTON-USER::C)
;                                COALTON:UNIT)))
; 
; caught COMMON-LISP:ERROR:
;   (during macroexpansion of (COALTON-TOPLEVEL
;     (DEFINE-CLASS #
;       ...)
;     ...))
;   Internal coalton bug: Preds not expected. (CLS BAR)
; 
; compilation unit finished
;   caught 1 ERROR condition
```

stylewarning:
It all seems that it's _almost_ doing the right thing, just not directing errors in the right ways/places.

colescott:
With #308 the first example now signals an error with the following message:

```
Failed to reduce context for CLS BAR
in COALTON
[...]

PR #311, opened 2021-12-14
Add more specific errors when toplevel implicit bindings have unresolved predicates

colescott:
Fixes second example in #304


src/typechecker/derive-type.lisp
```diff
@@ -460,6 +462,15 @@ EXPL-DECLARATIONS is a HASH-TABLE from SYMBOL to SCHEME"
                  (multiple-value-bind (typed-node binding-preds new-subs)
                      (derive-binding-type name tvar expr env subs name-map)
                    (setf subs new-subs)
+
+                   (with-type-context ("definition of ~A" name)
+                     (when (not allow-deferred-predicates)
```
stylewarning, line 467: unless

stylewarning (approved):

src/typechecker/derive-type.lisp
colescott: I think this is a bit easier to understand because it is not the default codepath

PR #315, opened 2021-12-15
Add explicit binding support to let forms

### LINKED ISSUE #13: Explicit bindings in LET blocks

colescott:
Let forms can now use declare similar to toplevel forms

Resolves #13

### DIFF
--- a/src/ast/node.lisp
+++ b/src/ast/node.lisp
@@ -86,10 +86,11 @@
 (defstruct
     (node-let
      (:include node)
-     (:constructor node-let (unparsed bindings subexpr name-map)))
-  (bindings (required 'bindings) :type binding-list :read-only t)
-  (subexpr  (required 'subexpr)  :type node         :read-only t)
-  (name-map (required 'name-map) :type list         :read-only t))
+     (:constructor node-let (unparsed bindings declared-types subexpr name-map)))
+  (bindings       (required 'bindings)      :type binding-list :read-only t)
+  (declared-types (required 'declare-types) :type list         :read-only t)
+  (subexpr        (required 'subexpr)       :type node         :read-only t)
+  (name-map       (required 'name-map)      :type list         :read-only t))
 
 #+sbcl
 (declaim (sb-ext:freeze-type node-let))
--- a/src/ast/parse-form.lisp
+++ b/src/ast/parse-form.lisp
@@ -104,25 +104,60 @@ This does not attempt to do any sort of analysis whatsoever. It is suitable for 
      (parse-form subexpr new-m package)
      (invert-alist binding-local-names))))
 
-(defun parse-let (unparsed bindings subexpr m package)
+(defun parse-let (unparsed binding-forms subexpr m package)
   (declare (type t unparsed)
-           (type list bindings)
+           (type list binding-forms)
            (type t subexpr)
            (type immutable-map m)
            (type package package)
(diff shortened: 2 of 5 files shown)


stylewarning (commented):
Why not allow DECLARE in the "usual" Lisp spot in the body?

colescott:
> Why not allow DECLARE in the "usual" Lisp spot in the body?

Allowing `declare` forms in subexpressions makes it so we can't do context-free parsing and would require a major reworking of the parsing system.

I also am partial to this because it is similar to the way you declare types for top level forms and is the way it is done in Haskell.

stylewarning:
The syntax would be

```
<let-form> ::=
  (let (<binding>*)
    <decl>*
    <body>)

<binding> ::=
  (<symbol> <expr>)

<decl> ::=
  (declare <var> <type-expr>)
```

Critically, this is syntax. Nothing can "return" a `declare`, and `declare` cannot be mixed in with `<body>`. This is even what `alexandria:parse-body` parses out (except it looks for `cl:declare`). I don't understand what context dependence there is.

colescott:
I guess "context-free parsing" isn't quite what I mean. With the current structure of the parser this will be hard to implement. 

Aside from implementation, the main issue I have with putting declares in the body the same way that lisp does it is that it means that type annotations, which in languages like Coalton act as a form of documentation, are separated from their respective values.

An example of this causing a bit of confusion is in a `let` with many forms such as
```
(let ((a some-long-form...)
      (b another-long-form...)
      (c another-form...))
  (declare a (Type :signature))
  (declare b (Complicated (Type Signature)))
  ...)
```
where it would be more helpful to someone reading the code to have it similar to how top level forms are annotated. 
```
(let ((declare a (Type :signature))
      (a some-long-form...)

      (declare b (Complicated (Type Signature)))
      (b another-long-form...)

      (c another-form...))
  ...)
```

I do agree that we should aim for the most lispy syntax, however I think in this case it leads to harder to read code and leads to inconsistencies with the existing syntax.

stylewarning:
Alternative food 4 thought:

```
<binding> ::=
    (<var> <expr>)
  | (<var> :OF-TYPE <type-expr> := <expr>)
```

This would match CL:LOOP style and would make things align nicely.

Here are all the options splayed out:

```
(let ((declare x Integer)
      (x 2)
      (declare y (Num :a => :a -> :a))
      (y (+ (fromInt 1))))
  ...)

(let ((x 2)
      (y (+ (fromInt 1))))
  (declare x Integer)
  (declare y (Num :a => :a -> :a))
  ...)

(let ((x :of-type integer
         := 2)
        (y :of-type (Num :a => :a -> :a)
         := (+ (fromInt 1)))
  ...)
```

Upon inspection, I'm actually OK with any of them. Maybe I prefer the "Lisp" way most just because it's, well, the Lisp way.

gefjon:
another option:
```
(let (((the Integer x) 2)
      ((the (Num :a => :a -> :a) y) (+ (fromInt 1))))
  ...)
```

stylewarning:
Now that some options have aired out, I think I'm happy if @colescott makes the final decision.

stylewarning:
Oh, if there's still time to squeeze a question in: How might we imagine declaring `fn` parameters? Good idea? No?

eliaslfox (approved):

PR #322, opened 2021-12-18
Warn instead of printing in coalton macro

eliaslfox:


colescott (approved):
Looks good. This will hopefully make it a bit more obvious when this issue occurs when embedded in lisp expressions

PR #326, opened 2021-12-21
my opinionated, scheme-inspired style guide

gefjon:


eliaslfox:
I don't think we should adopt either `?` or `!` suffixes. They don't add information that isn't already visible a functions type signature. Predicates already return Boolean. Functions that mutate return `Unit` or similar while non-mutating functions return a new data structure. Also data structures have been split between immutable and mutable. There isn't a need to distinguish between mutating and non mutating sort operations on lists because all lists are immutable.

gefjon:
It would be nice for mutating functions to have return values other than `Unit`, e.g. so they could be meaningfully passed to `fold` or `reduce`. The Common Lisp convention of `setf` returning the new value is often quite useful. In my brief advent-of-code usage of Coalton, I redefined various cell and vector operators to return the updated container.

`?` as a suffix I'm less attached to as adding meaning, but I think it's pleasant to read, and I see very little cost. Note that most C-like languages prefix such functions with `is`, like `Option::is_none` in Rust. I am not familiar with a language where the convention is to use bare adjectives to name predicates (though such a language almost certainly exists.)

I will be more convinced by arguments about type signatures already encoding information when my Emacs responds meaningfully to `sly-describe-symbol` or `sly-edit-definition` in Coalton code. Plus, I believe more descriptive names are both a lower barrier to implement and a higher payoff in terms of accessibility to new users relative to these editor features.

eliaslfox:
I like `is_` as a prefix.

Adding type information to docstrings should be easy, I can work on that.

Would passing mutating functions to fold still be necessary if lists had `for-each`?

stylewarning:
Does the verb `is` need to match the count of the noun?

gefjon:
> I like is- as a prefix.

This is both longer and less expressive than the `?` suffix. If we want a naming convention for predicates, I really cannot see an argument for preferring `is-` over `?`. The reason Rust et.al. use `is` is because their syntax doesn't allow `?` in symbols. Coalton's does.

I contend that
```
(if (empty? vec)
  None 
  (Some (index vec 0)))
```
is more natural than
```
(if (is-empty vec)
  None
  (Some (index vec 0)))
```

From a linguistic perspective, verb phrases should be used to name operations; predicates are not operations, they're queries.

> Would passing mutating functions to fold still be necessary if lists had for-each?

It's not so much about what's necessary as what's convenient. It's always gonna be possible to write 
```
(fold (fn (cel elt)
        (progn (cell-update (func elt) cel)
               cel))
      cel
      lst)
```
I'd just rather be able to write
```
(fold (compose cell-update! func) cel lst)
```

This is a degenerate example because cells are uninteresting, but there are places in my AoC code where I found it more convenient to `fold`/`reduce` with a mutable vector or matrix as state rather than to `for-each` with a closure.

Also, I don't actually believe that returning `Unit` is a useful way to denote mutations in type signatures. Is `(const Unit)` side-effecting? Its type is `:a -> Unit`, so it should be, right? Is `read-line` side-effecting? Its type is `InputStream -> String`, which does not return `Unit`, so it shouldn't be. If you really want the type system to encode information about side effects, I'm game to discuss options, but I don't think a return type of `Unit` is the right way to do that.

eliaslfox (approved):
I've come around on this. It seems reasonable given Coalton's s-expression based syntax.

colescott (approved):
I missed out on the discussion but I do think using `!` and `?` suffixes work well with the way Coalton is written/how it looks. I look forward to updating the standard library to match these guidelines

colescott:
@gefjon If you are able to rebase your branch on top of `main` then I can go ahead and merge this

gefjon:
I can see two issues that my style guide does not yet explicate:

1. How to name conversion functions? `int-string` or `int->string`?
2. What suffix to use for side-effecting predicates? (rare, but iterators will get some)

My heart says:

1. Don't define conversion functions, define `Into` instances.
2. Side effecting predicates get the `!?` suffix, like `any!?` to test whether any element of an iterator matches a predicate, as analogous to list's `any?`.

gefjon:
Having thought more about it, I think I'd like to revise my previous point 2 to instead be:

2. Side effecting functions are not predicates, even if they return booleans. They get the `!` suffix, like `any!` to test whether any element of an iterator matches a predicate, analogous to `any?` on lists.

PR #328, opened 2021-12-22
StatefulComputation type and instances

chebert:
Created a new type to represent stateful computations.

Implemented a functor, applicative, and monadic instance for threading stateful computations together.
LiftA2 threads the state from left to right.

Three functions are provided for
 - running the stateful computations to produce a final state and value
 - setting the state of a stateful computation
 - returning the state of a stateful computation as a value

### DIFF
--- a/src/library/types.lisp
+++ b/src/library/types.lisp
@@ -91,9 +91,4 @@
     ;; This shouldn't be pattern matched against with user code.
     ;;
     ;; See fraction.lisp for more info.
-    (%Fraction Integer Integer))
-
-  (define-type (StatefulComputation :state :value)
-    "A computation of a value which may affect the state.
-Represented as a closure from initial state to updated state and value."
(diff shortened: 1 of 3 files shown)


src/library/types.lisp
```diff
@@ -91,4 +91,9 @@
     ;; This shouldn't be pattern matched against with user code.
     ;;
     ;; See fraction.lisp for more info.
-    (%Fraction Integer Integer)))
+    (%Fraction Integer Integer))
+
+  (define-type (StatefulComputation :state :value)
```
eliaslfox, line 96: This can be in stateful-computation.lisp. Types are only predeclared here when it's needed to avoid circular dependencies.

eliaslfox (changes requested):
Other than the type location, this looks great.

eliaslfox (approved):

PR #329, opened 2021-12-23
Remove graph from standard library

eliaslfox:
The graph code is unused and incomplete.


stylewarning:
Can you be more specific? This was used to build out a feature of QUILC. I'm supportive if there's more justification, like there being fundamental flaws.

eliaslfox:
The code isn't really flawed. But it is incomplete, and I don't think undirected graphs work. Also it feels a little too specific to include in the standard library. If you think the code is worth keeping then I can move it to another repository.

stylewarning:
Can we make it into an example program?

eliaslfox:
If you want to finish the code.

colescott:
I agree that the code is not complete and should probably not be included in the standard library. @stylewarning would you be okay with moving it into a separate repo/system/package like `coalton-graph`? I think this would be the best compromise to keeping it available while also not adding bloat to the standard library

stylewarning:
> I agree that the code is not complete and should probably not be included in the standard library. @stylewarning would you be okay with moving it into a separate repo/system/package like `coalton-graph`? I think this would be the best compromise to keeping it available while also not adding bloat to the standard library

I think it could be in the small-coalton-programs. It doesn't hurt to have more code loaded. It may be incomplete but I completely agree it's not ready to be standard.

colescott:
I think that if we want this to be a library that Coalton language users can use in their projects then I think having it as a separate project and allowing them to depend on it seems better than having them load it in as part of our example project code. We can still have tests in that repository but I don't think the compiler absolutely needs to have this as a test case (especially since we already have quite a few example Coalton projects which are loaded in the test suite).

stylewarning:
I'm not suggesting it as a usable library, just some additional test code that gets loaded and is something people can look at just to read Coalton code.

stylewarning:
In its current state I don't think it's a library people should use. But it's complicated enough that I think it's a decent test case to load as a part of tests; I suggest it goes into the small-coalton-programs folder or whatever, which additionally implies it shouldn't be used like a library.

PR #338, opened 2021-12-24
Allow redefining compiler-internal structs in development mode

gefjon:
Replace `*interaction-mode*` with a function `coalton-release-p` which inspects the
feature `:coalton-release`.

Lock all `freeze-type` declarations behind a reader-conditional on `coalton-release`, so
that compiling in development mode does not freeze internal structs.

Rationale: as a galaxy-brain Lisper, I get angry whenever I have to restart my Lisp process. This morning, I pulled a change which altered a `defstruct` form somewhere in the Coalton compiler. Because that struct had been `freeze-type`'d, SBCL rejected the change, and I found myself unable to continue without restarting.

We could have some discussion about whether `interaction-mode` is the correct knob to control how interactively-redefinable the Coalton compiler itself is, as that interface is intended for controlling the redefinability of user code, and most users will not care whether they can redefine the compiler's internal structs. As the compiler becomes more stable, it may be desirable to have a separate feature `:coalton-compiler-dev` which controls these `freeze-types`. For now, though, this seems fine.


stylewarning (approved):
I feel a little queasy not using a special variable because then it's harder to test/dynamically change. But at the same time it seems hard to stay consistent without read-level integration anyway.

gefjon:
> then it's harder to test/dynamically change

I don't believe it should be, or is, possible to dynamically change this. As far as I can tell, switching Coalton development modes requires clearing ASDF caches, forcibly recompiling and reloading, and when downgrading from release to development, some hackery to undo `freeze-type` declarations.

> not using a special variable
`*features*` is a special variable :P

gefjon:
does anyone object to this? if not, i'd rather get it merged so i can stop thinking about it, & if so i'd rather close it for the same reason

stylewarning:
I'm okay with it. I'd like @colescott or @eliaslfox to weigh in too.

eliaslfox (approved):
I think this is good. Long term we should determine if freezing structs has a meaningful performance impact.

stylewarning:
> I think this is good. Long term we should determine if freezing structs has a meaningful performance impact. 

I would even be ok just removing them and adding them back if we actually meter perf.

PR #344, opened 2021-12-30
`cl:fixnum` interface, revamped custom-hash hash tables, and bitwise integer ops

gefjon:
and as an added bonus, a `Bits` typeclass for bitwise ops on integers

This PR contains two loosely-coupled things:
1. a class `Bits` where `(Num :int) => (Bits :int)` with bitwise operations
2. a revamp of `hashtable.lisp` to use SBCL's `:hash-function` parameter with a class `(Hash :key)`

I hope we can all agree that the current implementation of hash tables in the standard library is both incorrect and stupid. (This is not meant as a slight; I assume whoever wrote it knew that it was a kludge, and you unfortunately can't do better within the CL standard.)

Unfortunately, GENHASH does not provide an interface that will work with Coalton; its requirement that tests and hash functions be declared at the top level is unworkable unless Coalton moves to compiling generics by monomorphizing (which it shouldn't).

SBCL's extensions do provide an interface that works nicely, and this PR implements Coalton hash tables in terms of it. The problem is that it's SBCL-specific.

Personally, I feel comfortable tying Coalton to SBCL. My employer hasn't purchased me Allegro or LispWorks licenses, so I don't care about supporting those platforms (though some coworkers do); I fear that CCL is unable to keep up with macOS/cocoa's rapid breaking changes & thus will not remain relevant; & none of the other CL impls feel worth mentioning.

In any case, you really should take my `Bits` typeclass at least. Or merge something like it. I don't even think there's a dependency between the two; I  I'm just feeling lazy at present, so I'm submitting this whole thing for review and discussion instead of splitting into two separate PRs.


src/library/classes.lisp
```diff
@@ -88,6 +88,18 @@
...
     (fromInt (Integer -> :a)))
 
+  ;;
+  ;; Bits
+  ;;
+
+  (define-class ((Num :int) => (Bits :int))
+    "Operations on the bits of twos-complement integers"
+    (bit-and (:int -> :int -> :int))
+    (bit-or (:int -> :int -> :int))
```
stylewarning, line 98: bit-ior 🥺?

src/library/classes.lisp
```diff
@@ -211,3 +223,23 @@ The fields are defined as follows:
...
+  (define-class ((Eq :a) => (Hash :a))
+    "Types which can be hashed for storage in hash tables.
+
+Invariant (== left right) implies (== (hash left) (hash right))."
+    (hash (:a -> Integer)))
+
+  (declare combine-hashes (Integer -> Integer -> Integer))
+  (define (combine-hashes left right)
+    (lisp Integer (left right)
+      (cl:logxor left right))))
```
stylewarning, line 237: Perhaps we can #+sbcl's hash mixer? I forgot what it's called. MIXF?

stylewarning (approved):
Exciting change!

src/library/hashtable.lisp
```diff
@@ -1,69 +1,88 @@
 (in-package #:coalton-library)
 
+#-sbcl (error "Coalton hash tables are only supported on SBCL!")
```
colescott, line 3: Although I think this is definitely a better interface over hash tables, I guess I'm not super comfortable with the change to have the standard library being supported on SBCL. I have looked in to support for different implementations in the past and they seemed to not require major changes, however a few were blocked on a backend rewrite for better codegen ordering. 

I would feel more comfortable merging this if we push for hash tables to be a non-prelude feature of the standard library and if there is a clear push forward on #331 such that the standard library does not sit broken on other implementations for too long.

src/package.lisp
```diff
@@ -324,6 +324,8 @@
    #:codegen-program
    #:compile-expression
    #:update-function-env
+   #:a1 #:a2 #:a3 #:a4 #:a5 #:a6 #:a7 #:a8 #:a9 #:a10
```
colescott, line 327: I'm not sure if this helps but there are functions `construct-function-entry` and `apply-function-entry` which could be wrapped with a macro to automatically select the correct function. I'm fine exposing the direct functions too but I think we should discuss what we want the external interface to be for this.

colescott (commented):
Looks great!

src/library/classes.lisp
stylewarning: Oh, the standard lib has BOOLEAN-OR so it wouldn't make sense to call it IOR.

gefjon:
I can see four ways to get a cross-implementation mutable hash table into Coalton:

1. Implement our own hash table (with blackjack and hookers). Realistically this is not terribly difficult, but I don't particularly want to do it.
2. Depend on GENHASH and subclass `hash-container` with a version that stores both its test and its hash-function, rather than storing only a test and implicitly associating it with a hash-function via a global resource (this is IMO a very stupid part of the design of GENHASH). This would be much less work, but would feel a little gross.
3. Do one of the above, but with a read-conditional so that SBCL would use its native (and I assume more efficient) implementation. This has all the downsides of the backing implementation, but also implies more maintenance work in exchange for added efficiency on SBCL.
4. Just bite the bullet and target SBCL. This will allow us to do fun things down the line like implement threading, synchronization and atomics in terms of SBCL's api, rather than bordeaux-threads' much worse interface, and I see very few downsides other than a moral desire to be portable. Can anyone name an implementation they would run Coalton on?

stylewarning:
Can we do 5 and just make a thin shim over hash tables conforming too the interface we expect and just have an SBCL implementation (for now)?

stylewarning:
(Sorry for double comment:) For number 4, a lot of the Quil stack is being ported to Allegro, so we would in principle not want to be antagonistic toward that.

gefjon:
> Can we do 5 and just make a thin shim over hash tables conforming too the interface we expect and just have an SBCL implementation (for now)?

Sure, I can do this.

gefjon:
See [defs.lisp](src/hashtable-shim/defs.lisp) for my proposed interface. It should be trivial to implement in terms of any CL extension that allows passing a `:hash-function` to `make-hash-table` at runtime.

stylewarning:
> See [defs.lisp](src/hashtable-shim/defs.lisp) for my proposed interface. It should be trivial to implement in terms of any CL extension that allows passing a `:hash-function` to `make-hash-table` at runtime.

I think a separate package would have been OK without the ceremony if a system, but a separate system is nice if we have a specific custom hash tables dep in the future like GENHASH. I'm fine to keep the sep system but I'm curious, what was your reasoning for it?

I approve this PR but we'll need to merge the fixnum stuff first, then rebase this, then merge this. So I'll hold off putting the green checkmark.

gefjon:
> I'm fine to keep the sep system but I'm curious, what was your reasoning for it?

I want to live in a world where we have many separate systems, more explicit `:depends-on`s, and fewer `:serial t`s. Ideally, `coalton/library` would be a separate system which `:depends-on` `coalton/compiler` and `coalton/hashtable-shim`, which would make it more clear that the compiler doesn't and shouldn't use the hashtable shim.

This code felt relatively decoupled from the rest, and as purely CL code it didn't feel like it belonged inside coalton-library. I started writing it there, but got fed up with prefixing every symbol with `cl:`, so I gave it its own package, subdirectory and subsystem all at once.

In any case, my personal style is to prefer writing subsystems rather than `:module` components. I wouldn't really mind making this a submodule of `coalton/library`, that just wasn't my first intuition as to how to write it. (I've always been sorta skeptical of the `:module` component class in general.)

gefjon:
> I approve this PR but we'll need to merge the fixnum stuff first, then rebase this, then merge this

I think it'd probably be easier to just make sure this branch includes all the fixnum stuff we want, and merge it all from here.

stylewarning:
> > I approve this PR but we'll need to merge the fixnum stuff first, then rebase this, then merge this
> 
> 
> 
> I think it'd probably be easier to just make sure this branch includes all the fixnum stuff we want, and merge it all from here.

Fine by me.

gefjon:
Seems to me that the last thing before merging all this is to decide on names for the `fixnum` and `(and fixnum unsigned-byte)` types. Current proposals are:

1. `Fixnum` and `Natnum`
2. `Fixnum` and `UFixnum`
3. `Fixnum` and `UnsignedFixnum`
4. `IFix` and `UFix`

I'm currently leaning towards 4, but it's not a strong preference.

gefjon:
I'll throw into the runnings:

5. `Int` and `UInt`

stylewarning:
> I'll throw into the runnings:
> 
> 
> 
> 5. `Int` and `UInt`

I vote 4 (I/UFix). I wrote a long screed in 2012 or so majorly embarrassing myself because at the base of my argument I (falsely) believed Int in Haskell was a bignum, so I can't seem hypocritical 10 years later. :)

stylewarning (approved):

PR #345, opened 2021-12-30
Fix codegen bug where deferred predicates generated functions

### LINKED ISSUE #343: Incorrect generalization of let-bound value
Minimum reproduction:
[issue-hash-predicate-not-shared.lisp.txt](https://github.com/coalton-lang/coalton/files/7790684/issue-hash-predicate-not-shared.lisp.txt)

Given a class `Key :key` and a type `Mapping :key :value`, and operators:
- `get` of type `(Key :key) => (Mapping :key :value) -> :key -> (Optional :value)`
- `set!` of type `(Key :key) => (Mapping :key :value) -> :key -> :value -> Unit`

the following definition of `swap!` generates erroneous code:
```
(declare swap! ((Key :key) => (Mapping :key :value) -> :key -> :value -> (Optional :value)))
  (define (swap! mapping key value)
    (progn
      (let old = (get mapping key))
      (set! mapping key value)
      old))
```
the generated code, with declarations and superfluous forms stripped away and gensyms renamed, looks something like:
```
(defun swap! (key-instance mapping key value)
  (let ((old (lambda (other-key-instance)
               (get other-key-instance mapping key))))
    (progn
      (set! key-instance mapping key value)
      old)))
```
a correct version would instead look like:
```
(defun swap! (key-instance mapping key value)
  (let ((old (get key-instance mapping key)))
    (progn
      (set! key-instance mapping key value)
      old)))
```
perhaps notably, this bug does not manifest if the result of `get` is not bound by `let`. that is, the following generates well-behaved code (it's just semantically different from the thing i want to do):
```
(coalton-toplevel
  (declare swap! ((Key :key) => (Mapping :key :value) -> :key -> :value -> (Optional :value)))
  (define (swap! mapping key value)
    (progn
      (set! mapping key value)
      (get mapping key))))
```

colescott:
Fixes #343

### DIFF
--- a/src/codegen/compile-expression.lisp
+++ b/src/codegen/compile-expression.lisp
@@ -219,7 +219,9 @@
                              (node (cdr binding))
                              (type (typed-node-type node))
 
-                             (preds (reduce-preds-for-codegen (scheme-predicates type) env))
+                             (preds
+                               (remove-if (lambda (p) (member p ctx :key #'car :test #'equalp))
+                                          (reduce-preds-for-codegen (scheme-predicates type) env)))
 
                              (dict-context (mapcar (lambda (pred) (cons pred (gensym))) preds)))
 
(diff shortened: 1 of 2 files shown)


stylewarning:
Can we add a test case for it?

colescott:
> Can we add a test case for it?

I'm not super sure how to add a test case for the codegen output... @eliaslfox do you have any ideas?

stylewarning:
We can merge just so bugs are fixed but we ought to figure out how to test these things

eliaslfox (approved):

PR #349, opened 2021-12-30
`Fixnum` and `Natnum` types

gefjon:
I'm open to discussion as to the name of `Natnum`, which corresponds to the CL type `(and fixnum unsigned-byte)`.

I was going to write a few updates on my hash-table PR (I have an idea about how to make it portable, at the cost of performing relatively poorly on non-SBCL platforms, but that's a different thread), when I was struck by the fact that Coalton didn't have a way to express the correct type of arguments to `sb-ext:mix`. This PR adds that type, `Natnum`, and defines `Fixnum` while I was at it since I'm sure I'll want it eventually.

I'd also like to raise: I'm not sure it's advisable to have `I64` and `U64` in the standard library. In my mind, these types imply to users a more efficient representation than `Integer`, which is not the case. At the very least, I think documentation should explicate that 64-bit integers may be heap-allocated when passed across full-call boundaries or stored in containers.

### DIFF
--- a/src/library/arith.lisp
+++ b/src/library/arith.lisp
@@ -21,10 +21,16 @@
                   ,(cl:format cl:nil "Signed value overflowed ~D bits." bits))
        (%unsigned->signed ,bits (cl:mod value ,(cl:expt 2 bits)))))))
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:defparameter +fixnum-bits+
+    #+sbcl sb-vm:n-fixnum-bits
+    #-sbcl (cl:1+ (cl:floor (cl:log cl:most-positive-fixnum 2)))))
+
 (%define-overflow-handler %handle-8bit-overflow 8)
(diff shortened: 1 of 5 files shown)


stylewarning:
How about unsigned-fixnum or ufixnum?

gefjon:
I'm good with `UFixnum`, but I'm struck by a possible alternative: for consistency with the other fixed-width integer types, we could adopt `IFix` and `UFix`. I'm not sure this is better, but it seems worth having on the table. Honestly, I still like `Natnum`, personally, though I see the problem that it doesn't obviously communicate its relationship with `Fixnum`.

stylewarning:
My objection to Natnum is that it feels like a synonym to UNSIGNED-BYTE.

IFix and UFix are interesting alternatives. Fits with the whole I and U integer nomenclature. Pays homage to the fixnum terminology. Symmetrical.

stylewarning:
These changes were folded in to the now-merged hash table PR.

PR #351, opened 2022-01-03
Improve documentation

eliaslfox:
Move definitions around and rewrite docstrings to improve the experience
when viewing as HTML.


src/library/list.lisp
```diff
@@ -176,7 +178,16 @@
...
-    "Returns a list containing the numbers from START to END inclusive."
+    "Returns a list containing the numbers from START to END inclusive.
+
+
+    ```
+    > COALTON-USER> (coalton (range 1 5))
+    (1 2 3 4 5)
+
+    > COALTON-USER> (coalton (range 5 2))
+    (5 4 2 2)
```
stylewarning, line 189: This doesn't look right.

src/library/classes.lisp
```diff
@@ -31,6 +31,19 @@
         ((Tuple (GT) (GT)) True)
         (_                 False))))
 
+  (define-instance (Ord Ord)
```
stylewarning, line 34: Seems a little odd if you ask me. Does Haskell or Rust do this?

src/library/builtin.lisp
```diff
@@ -17,6 +17,8 @@
   (declare fix (((:a -> :b) -> (:a -> :b)) -> (:a -> :b)))
   (define (fix f n)
     "The factorial function can be written
```
stylewarning, line 19: I think this function should have better documentation.

src/library/arith.lisp
```diff
@@ -485,3 +504,118 @@
...
+    (ceiling (exact/ a b)))
+
+  (declare round/ (Integer -> Integer -> Integer))
+  (define (round/ a b)
+    "Divide two integers and round the quotient."
+    (round (exact/ a b)))
+
+  (declare single/ (Single-Float -> Single-Float -> Single-Float))
+  (define (single/ a b)
+    "Compute the quotient of single-precision floats A and B as a single-precision float."
```
stylewarning, line 614: In other places you deleted argument names from the doc string. Should it happen here as well?

src/library/arith.lisp
```diff
@@ -485,3 +504,118 @@
 (define-sxhash-hasher UFix)
 (define-sxhash-hasher Single-Float)
 (define-sxhash-hasher Double-Float)
+
+
+;;; `Quantization' 
+
+(coalton-toplevel
+  (define-instance (Quantizable Integer)
```
stylewarning, line 512: FYI: I noticed the generated documentation for the Quantizable class didn't look good due to
mismatching autogenerated type var names and those that were mentioned in the docs.

stylewarning (changes requested):

src/library/classes.lisp
eliaslfox: [rust does](https://doc.rust-lang.org/stable/std/cmp/enum.Ordering.html#impl-Ord)

src/library/list.lisp
eliaslfox: thanks

src/library/arith.lisp
eliaslfox: yes

src/library/arith.lisp
eliaslfox: fixed

src/library/classes.lisp
eliaslfox: [haskell does too](https://hackage.haskell.org/package/base-4.16.0.0/docs/Prelude.html#t:Ordering)

src/library/builtin.lisp
eliaslfox: I don't really know how to describe it. And I don't want to copy the description from Haskell because I think the semantics might be a little different w/ strictness. If you have a better docstring I can add it.

stylewarning:
```
Compute the fixed point of a unary function. This is equivalent to the Y-combinator of the lambda calculus.

This combinator allows recursion without specific assignment of names. For example, ..."
```

stylewarning (approved):

PR #353, opened 2022-01-05
Reduce context when deriving function types

### LINKED ISSUE #350: Anonymous function with type class function inside not allowed in pipe
```
;; happy
(coalton-toplevel
  (define (bad-pipe x)
    (pipe
     x
     (fn (x)
       x))))

;; not happy
(coalton-toplevel
  (define (bad-pipe x)
    (pipe
     x
     (fn (x)
       (+ x 1)))))

;; happy
(coalton-toplevel
  (define add1 (+ 1))
  (define (bad-pipe x)
    (pipe
     x
     (fn (x)
       (add1 x)))))
;; happy
(coalton-toplevel
  (define add1 (+ 1))
  (define (bad-pipe x)
    (pipe
     x
     add1)))
```
the unhappy case produces this error:
```
  warning: 
    Derived type of
    (SB-C::%FUNCALL (SB-KERNEL:%COERCE-CALLABLE-FOR-CALL #:F4) #:G5) is
      (COMMON-LISP:VALUES COMMON-LISP:INTEGER COMMON-LISP:&OPTIONAL),
    conflicting with its asserted type
      (OR FUNCTION COALTON-IMPL/CODEGEN::FUNCTION-ENTRY).
    --> PROGN PROGN DEFUN PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN 
    --> SB-INT:NAMED-LAMBDA FUNCTION THE PROGN BLOCK 
    --> COALTON-IMPL/CODEGEN::A2 SB-C::INLINED BLOCK ETYPECASE LET COND IF 
    --> PROGN 
    ==>
      (COALTON-IMPL/CODEGEN::A1 (FUNCALL #:F4 #:G5) #:G6)
```
It looks to me that this has something to do with the "dict"ness of `+`. By "dict"ness, I mean that `+` is part of a type-class giving the implicit `dict` parameter argument in the sly minibuffer help.

eliaslfox:
Fixes #350

### DIFF
--- a/src/codegen/compile-expression.lisp
+++ b/src/codegen/compile-expression.lisp
@@ -44,8 +44,10 @@
 
   (:method ((expr typed-node-direct-application) ctx env)
     `(,(typed-node-direct-application-rator expr)
-      ,@(compile-typeclass-dicts (scheme-predicates
-                                  (typed-node-direct-application-rator-type expr))
+      ,@(compile-typeclass-dicts (remove-duplicates
+                                  (scheme-predicates
+                                   (typed-node-direct-application-rator-type expr))
+                                  :test #'equalp)
                                  ctx env)
       ,@(mapcar
          (lambda (expr) (compile-expression expr ctx env))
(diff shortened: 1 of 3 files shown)


stylewarning (approved):
@colescott should look too.

colescott (approved):

PR #357, opened 2022-01-06
Add hash instances for tuples

eliaslfox:


stylewarning (approved):
I'm fine with this but maybe we can hash in the hash of a string like `"TUPLE"` with it? This would be just to reduce collisions with other product types that are hashed equivalently... but it's not necessary.

stylewarning:
Ok just rebase and we can merge.

PR #368, opened 2022-01-10
basic complex arithmetic with Complex class

stylewarning:
This introduces the bare minimum needed to represent complex numbers with any participating data type. Functions like `sqrt` and `exp` and `abs` aren't implemented yet.

### DIFF
--- a/src/library/arith.lisp
+++ b/src/library/arith.lisp
@@ -417,6 +417,58 @@
   )
 
 (coalton-toplevel
+  (declare real-part ((Complex :a) -> :a))
+  (define (real-part z)
+    "The real part of a complex number."
+    (match z
+      ((Complex a _) a)))
+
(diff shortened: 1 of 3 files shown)


colescott (approved):
Looks good!

In the future we might want special compilation down to `(complex ...)` lisp types for some of these but I think that is not crucial for an initial implementation (cc @eliaslfox)

stylewarning:
> Looks good!
> 
> 
> 
> In the future we might want special compilation down to `(complex ...)` lisp types for some of these but I think that is not crucial for an initial implementation (cc @eliaslfox)

I was thinking this. I'd love to know more info about how we might have specialized representations for non-simple types (eg `(Complex Double-Float)`). If we crack that nut, then we'll also be able to do unboxed array representations (eg a fabled `(Array Single-Float)`).

PR #378, opened 2022-01-27
test hash tables, and fix various bugs

gefjon:
in no particular order:

- defines an (ugly) infrastructure (which reaches into fiasco internals) for writing tests
  natively in coalton.

- uses said infrastructure to verify that the three basic hashtable operators work

- fixes a typo where i passed args in the wrong order to `custom-hashtable-get`

- splits the sbcl hashtable shim into two, based on sbcl version
  - pre 2.1.12 (not yet released) requires getting our hands dirty to paper over a bug
  - post 2.1.12 (someday!) will use the clean, beautiful implementation i originally wrote

- changes the way `%make-hashtable` tricks the coalton compiler into constructing
  "monomorphic" specializations of `hash` and `==`, because the old way didn't actually do
  that, and was causing bugs where sbcl got an f3 when it wanted an f2.


eliaslfox (approved):
This looks great

PR #381, opened 2022-02-14
Codegen functions from the environment type

### LINKED ISSUE #377: Incorrect code generation with multiple constraints on type
When applying multiple constraints to a single type variable in a declaration, the definition generates code which reads from the wrong instance. For example:

```
(uiop:define-package :bug-multiple-predicates
  (:use :coalton-library :coalton))
(cl:in-package :bug-multiple-predicates)

(coalton-toplevel
  (declare range-increasing ((Num :num) (Ord :num) => :num -> :num -> :num -> (List :num)))
  (define (range-increasing step start end)
    (let ((inner
            (fn (current acc)
              (if (> current end)
                  acc
                  (inner (+ current step)
                         (Cons current acc))))))
       (inner start Nil))))
```

This compiles without error (as I would expect), but invoking the function in the repl signals a type-error:

```
BUG-MULTIPLE-PREDICATES> (coalton (range-increasing 1 0 10))
; Debugger entered on #<COMMON-LISP:TYPE-ERROR expected-type: COALTON-LIBRARY::CLASS/NUM
;                          datum: #<COALTON-LIBRARY::CLASS/ORD {7006FDCC63}>>
```

eliaslfox:
Previously Coalton looked at the type on a typed-node. This is incorrect
because the type of a typed-node does not always match the inferred type
for the definiton that generated that typed-node.

Fixes #377

### DIFF
--- a/src/coalton.lisp
+++ b/src/coalton.lisp
@@ -127,6 +127,12 @@ in FORMS that begin with that operator."
   `(let ((*emit-type-annotations* t))
      (process-coalton-toplevel ',toplevel-forms *global-environment*)))
 
+(defmacro coalton:coalton-codegen-ast (&body toplevel-forms)
+  "Prints the AST of the typechecked coalton code. Intended for debugging."
+  `(let ((*coalton-dump-ast* t))
+     (process-coalton-toplevel ',toplevel-forms *global-environment*)
+     (values)))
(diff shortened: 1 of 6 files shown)


stylewarning (approved):
looks great

PR #382, opened 2022-02-15
`Iterator` interface

gefjon:
also includes some operators not immediately related to iterators but required for some of
the definitions

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -99,6 +99,7 @@
                              (:file "slice")
                              (:file "hashtable")
                              (:file "monad/state")
+                             (:file "iterator")
                              (:file "prelude")))
                (:file "toplevel-environment")))
 
@@ -169,4 +170,5 @@
                (:file "environment-persist-tests")
                (:file "slice-tests")
                (:file "quantize-tests")
-               (:file "hashtable-tests")))
+               (:file "hashtable-tests")
+               (:file "iterator-tests")))
--- a/src/language-macros.lisp
+++ b/src/language-macros.lisp
@@ -196,3 +196,23 @@ Coalton boolean."
                          `(coalton:seq
                            ,@forms)))))
     (process forms)))
+
+(cl:defmacro assert (datum cl:&optional (format-string "") cl:&rest format-data)
+  "Signal an error unless DATUM is `True'.
+If the assertion fails, the signaled error will apply the FORMAT-DATA to the FORMAT-STRING via `cl:format' to
+produce an error message."
+  ;; OPTIMIZE: lazily evaluate the FORMAT-DATA only when the assertion fails
+  (cl:check-type format-string cl:string)
+  (cl:let* ((datum-temp (cl:gensym "ASSERT-DATUM-"))
+            (format-data-temps (alexandria:make-gensym-list (cl:length format-data)
+                                                            "ASSERT-FORMAT-DATUM-")))
+    `(let ((,datum-temp ,datum)
+           ,@(cl:mapcar #'cl:list format-data-temps format-data))
+       (progn
+         (lisp :any (,datum-temp ,@format-data-temps)
+           (cl:assert ,datum-temp ()
+                      ,(cl:format cl:nil
+                                  "Assertion ~a failed: ~a"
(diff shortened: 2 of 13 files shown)


gefjon:
Don't know why those builds are failing. Builds fine and tests pass on my machine.

stylewarning:
@gefjon @eliaslfox does this have to do with the new defaulting and overloaded integer literals?

eliaslfox:
I don't think so, the literal changes haven't been merged into this branch. I just checked and the tests pass locally for me too.

gefjon:
After merging `main`, I can reproduce the error locally. Not sure why it was manifesting here before that merge.

stylewarning:
> For the life of me I cannot tell what caused those errors, or why the CI caught them before merging, but after a few minutes of pounding my head on my keyboard, I've found formulations of `string-chars` and `count-forever` which compile.

Now that integers are generic, you may not need so many fromInt calls.

eliaslfox:
I'd prefer for iterators to maintain their own size information. Rust exposes this as [size_hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint). I think the easiest way to encode this would be to add another field to the iter definition.

```
(define-type (Iterator :a)
  (%Iterator (Unit -> (Optional :a)) (Unit -> (Tuple UFix (Optional UFix)))))
```

eliaslfox:
Also how are the following for applicative and monad instances

```
  (define-instance (Applicative Iterator)
    (define (pure x)
      (progn
        (let yielded? = (make-cell False))
        (Iterator
          (if (cell-read yielded?)
              None
              (progn
                (cell-write! yielded? True)
                (Some x))))))

    (define (liftA2 f a b)
      (Iterator
        (liftA2 f (next! a) (next! b)))))

  (define-instance (Monad Iterator)
    (define (>>= m f)
      (flatten! (map f m)))))
```

gefjon:
> Now that integers are generic, you may not need so many fromInt calls.

I suspect that there are many places where I can now remove `fromInt`, but I haven't gotten around to it yet.

> I'd prefer for iterators to maintain their own size information.

I'm disinclined to do this for a few reasons. In rough order of importance to me:
1. Most importantly, I really want `Iterator` to be a single-constructor single-field newtype, so that the release mode compiler can optimize it away. Rust solves this problem by making `Iterator` a trait/typeclass and monomorphizing, which we cannot do.
2. I want it to be easy to use the `Iterator` macro to define new iterators and to construct anonymous iterators without having to grok the range tuple. Rust solves this problem by defaulting `size_hint` to no information.
3. I don't believe tracking size adds useful information; many iterators' `size_hint`s will necessarily return the unbounded range, even without defaulting for user iterators.
4. Relatedly, very few `Iterator` operators care about a size hint; only a subset of the collecting operators would observe it at all, and for them it's only an optimization to avoid reallocing/resizing. The `collect-foo-size-hint!` operators permit users to do that optimization, but I think it's enough of a corner case that I don't feel the need to automate it.

> how are the following for applicative and monad instances

I think the only unsurprising `Applicative` and `Monad` instance is analogous to the one Haskell puts on `[]`, with `liftA2` being a Cartesian product. I wouldn't be opposed to including that one if someone asked for it, but I'm not champing at the bit to write it before I hear that someone wants it.

eliaslfox (approved):
> I really want Iterator to be a single-constructor single-field newtype, so that the release mode compiler can optimize it away.

Yeah that's probably more important.

I felt like `liftA2` on iterators being a zipper was intuitive. But I see how that could be confusing.

eliaslfox:
@gefjon I rewrote the commit history to avoid merge conflicts with #392. I changed a couple things.

* custom iterators didn't seem like they would be needed frequently enough to justify a macro, they can instead be created with `coalton-library/iterator:new`

* using bit representation for odd/even seemed weird. Haskell has [integral](https://hackage.haskell.org/package/base-4.16.0.0/docs/Prelude.html#t:Integral) and we probably should too.

* I removed `fn*`, I don't want to extend the base language syntax with more macros. Implicit progn seemed like a good idea, #398 will add it for all functions.

PR #384, opened 2022-02-15
Remove scheme and arguments from constructor-entry

eliaslfox:
Both of these fields store information duplicate type information. A
constructor-entry's scheme is equivalent to it's associated value's
scheme. A constructor-entry's arguments can be computed from that
scheme.

### DIFF
--- a/src/toplevel-define-type.lisp
+++ b/src/toplevel-define-type.lisp
@@ -28,48 +28,50 @@
                 :newtype (type-definition-newtype parsed-deftype)
                 :docstring (type-definition-docstring parsed-deftype))))
 
-        (dolist (ctor (type-definition-constructors parsed-deftype))
-          ;; Add the constructors to the constructor environment
-          (setf env
-                (set-constructor
-                 env
-                 (constructor-entry-name ctor) ctor))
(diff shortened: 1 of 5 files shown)


src/toplevel-define-type.lisp
```diff
@@ -28,48 +28,50 @@
...
+                        env
+                        (constructor-entry-name ctor)
+                        (make-name-entry
+                         :name (constructor-entry-name ctor)
+                         :type :constructor
+                         :docstring nil
+                         :location (or *compile-file-pathname*
+                                       *load-truename*))))
 
-              ;; TODO: This does not seem right... Is this to ensure
```
eliaslfox, line 67: @gefjon I deleted your TODO comment. Your assumption about redefined constructors is correct.

stylewarning:
Would it be valuable to still have the functions constructor-entry-{scheme, arguments} defined, just implemented in the way you suggest?

eliaslfox:
`constructor-entry-scheme` returned duplicate information. I added `constructor-arguments` as a replacement for `constructor-entry-arguments`.

stylewarning (approved):

PR #392, opened 2022-02-18
Reorganize Standard Library

eliaslfox:
* splits the standard library into multiple packages (`coalton-library/vector`, `coalton-library/cell`, ...)
* adds the package `coalton-prelude` to be used in Coalton source files
* removes the package `coalton-library`
* add package locks to the standard library
* minor compiler changes to generate symbols in the current package
* rename some functions to match style guide

Still needs updates to the documentation generator.


eliaslfox:
One thing I learned was that shadowing a symbol macro from a use'd package causes a lock violation. The following code is invalid

```lisp
(defpackage #:test-package 
  (:use #:coalton #:coalton-prelude))

(coalton-toplevel
  (define x
    (let ((hash 5))
      (lisp Integer (hash) hash))))
```

Local variables are renamed, so this an only an issue with `lisp` forms.

colescott (approved):
:shipit:

PR #394, opened 2022-02-18
Fix subtraction for Complex

aijony:
Fix a typo where complex numbers are added instead of subtracted.


eliaslfox (approved):
Looks great!

PR #398, opened 2022-02-21
Fn syntax improvements

eliaslfox:
* add nullary fn syntax `(fn () 5)`
* add implicit fn progn

### DIFF
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -474,7 +474,18 @@ Coalton's `progn` can have flattened `let` syntax.
     (progn
       (let x_ = (into x))
       (let y_ = (into y))
-      (concat-string x_ y_))))
+      (<> x_ y_))))
+```
+
+Function defintions create an implicit `progn` block
+
+```lisp
+(coalton-toplevel
+  (declare f (Integer -> Integer -> String))
+  (define (f x y)
+    (let x_ = (into x))
+    (let y_ = (into y))
+    (<> x_ y_)))
 ```
 
 ## Unless and When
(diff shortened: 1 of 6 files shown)


eliaslfox:
Eventually I think `(let x = y)` forms in progn blocks should expand to non-recursive bindings.

gefjon:
This may be out-of-scope for now, but I'd very much like to change all binding forms (`let`, `fn` and the function form of `define`, plus possibly others I'm forgetting) to have:

- implicit `progn` in the body (inner scope), and
- implicit destructuring (irrefutable matching) on bound names.

I understand this second point is somewhat controversial since we don't have exhaustiveness checking, but I have found that a majority of `fn`s and function-form `define`s I write begin immediately with a single-arm `match`. If it were up to me, we'd edit the built-in `fn` to behave like the `fn*` macro that I wasn't able to sneak by you in the iterators pr, and edit `let` and the `progn (let ... = ...)` the same way.

eliaslfox:
> implicit progn in the body (inner scope), and
> implicit destructuring (irrefutable matching) on bound names.

Yeah that does seem like the right direction for the language syntax. I'm less worried about irrefutable patterns than I am about the error messages `fn*` would have generated. A function like

````
(define (fn* f (Some x) (Some y) (Some z))
  ...)
``` 

could generate a trace like

```
Unable to unify String and Integer
In branch on (Some z)
In match on G405 
In branch on (Some y)
In match on G404
In branch on (Some x)
In match on G403
In definition of f
In Coalton Toplevel
```

Implementing the restructuring in the compiler will avoid error messages mentioning variables not in a users code. It'll also avoid the error trace mentioning other match branches.

FWIW this pr does add progn to function style defines. `(define (f a) y)` is rewritten into `(define f (fn (a) y))` early on in parsing.

colescott (approved):
Looks great!

stylewarning (approved):

PR #405, opened 2022-02-23
Add `(REPR :TRANSPARENT)` and `(REPR :NATIVE)`

### LINKED ISSUE #395: Repr Newtype
Add `(repr :newtype)` or `(repr :transparent)`. This would force the newtype optimization in debug mode.

eliaslfox:
Fixes #395 

`(repr :transparent)` forces the newtype optimization in development mode. 

`(repr :native x)` makes a types runtime representation `x`. This is useful for interop hacks.

### DIFF
--- a/src/typechecker/parse-type-definition.lisp
+++ b/src/typechecker/parse-type-definition.lisp
@@ -163,21 +163,9 @@ Returns TYPE-DEFINITIONS"
                     :constructor-types ctor-types
                     :docstring docstring))
 
-                  ((and enum-type
-                        (coalton-impl:coalton-release-p))
-                   (let ((parsed-ctors (mapcar #'rewrite-ctor parsed-ctors)))
-                     (make-type-definition
-                      :name tycon-name
-                      :type tcon
-                      :runtime-type `(member ,@(mapcar #'constructor-entry-compressed-repr parsed-ctors))
-                      :enum-repr t
-                      :newtype nil
-                      :constructors parsed-ctors
-                      :constructor-types ctor-types
-                      :docstring docstring)))
-
-                  ((and newtype
-                        (coalton-impl:coalton-release-p))
+                  ((or (and newtype (eql repr :transparent))
+                    (and newtype
+                            (coalton-impl:coalton-release-p)))
                    (let (;; The runtime type of a newtype is the runtime type of it's only constructor's only argument
                          (runtime-type (function-type-from
                                         (qualified-ty-type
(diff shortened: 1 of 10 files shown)


src/typechecker/parse-type-definition.lisp
```diff
@@ -192,6 +195,23 @@ Returns TYPE-DEFINITIONS"
                       :constructor-types ctor-types
                       :docstring docstring)))
 
+                  ((and (eql repr :transparent) (not newtype))
+                   (error "Type ~A cannot be repr transparent" tycon-name))
```
colescott, line 199: Is this the error shown to the end user when a type cannot be made transparent? If so it should include some explanation about how the type must be a newtype to be transparent.

colescott:
How does the compiler know how to use the native type repr? Is this meant for values that are simply handed around in Coalton?

eliaslfox:
The native repr doesn't really effect the compiler. The runtime repr is only used for type annotations in lisp. We could use it to do things like

```
(repr :native cl:boolean)
(define-type Boolean)

(define True (lisp Boolean () cl:t))
(define False (lisp Boolean () cl:nil))
```

stylewarning:
This doesn't seem unreasonable, but it seems that it creates a lot of non-obvious effects from a library user's perspective. Why force optimization in debug mode in the standard library? Should `:transparent` and `:native` instead be a private Coalton features? i.e. replace `:transparent` with `coalton-impl::transparent`, so as to dissuade the user from using these features?

eliaslfox:
I consider these to be similar to `Lisp-Object`. Most users won't care about them, but people writing interop code probably will.

> Why force optimization in debug mode in the standard library?

These optimizations are off for user code because they break redefining types. I don't think we're going to redefine vectors or hashtables frequently. Also it means I can remove all of the code that looks like

```
(match v
  ((%Vector v) (lisp (Vector :a) (v) (%Vector (... actual code starts here...)))))
```

eliaslfox:
This also avoids extra wrapper calls from the lisp side.

```
CL-USER> (coalton-library/vector:index-unsafe 1 (coalton-library/vector::%vector #(1 2 3)))
```

stylewarning:
So I agree completely with you on the standard library, but wouldn't that behoove us to make the specific REPR command symbol internal for our purposes, and thus if people use it for their library code, "here be dragons"?

eliaslfox:
I don't think we should use internal symbols to indicate when a part of the language can be misused to violate static guarantees. It's already trivial to sidestep the type system with `lisp`. Repr has a well defined semantics. I want to encourage its use for interop.

colescott:
I would be fine to leave it public but I agree that the semantics should be clearly documented somewhere. Would it make sense to start up an advanced guide on how to do fancier interop with Common Lisp?

eliaslfox:
Doesn't that exist already? [https://github.com/coalton-lang/coalton/blob/main/docs/coalton-lisp-interop.md](https://github.com/coalton-lang/coalton/blob/main/docs/coalton-lisp-interop.md)

stylewarning (approved):

PR #414, opened 2022-03-08
Expression-oriented `Optional` redefinition of `when` and `unless`

gefjon:
Redefine `when` and `unless` to return an `Optional`, the result from the body is returned as `Some`, and returning `None` if the body is not evaluated.

Also, rename the `Optional` operators to have scheme-ey names; `fromSome` => `from-some`, `isSome` => `some?`, `isNone` => `none?`.


eliaslfox:
I think the current definitions of when and unless are useful and shouldn't be changed. 

The proposed versions in this pr seem potentially useful but I'm not sure that I've ever wanted them. @gefjon could you share some examples of their usage?

Also they should probably use `pure`/`empty` instead of `Some`/`None`.

gefjon:
> I think the current definitions of when and unless are useful and shouldn't be changed.

Can you think of an example of a usage of the `Unit`-returning `when` and `unless` which:
1. no longer compiles with the `Optional`-based definitions,
2. cannot be trivially transformed into an expression-oriented use of `if` or `cond` the way this PR redefines `slice:new` or `%mkFraction`, and
3. are not obviously symptoms of inefficient and unidiomatic procedural code, like the current definitions of `==` for `Slice` and `Vector`?
 
(Note that both of these in 3 can be elegantly and more efficiently redefined in terms of iterators, but I haven't gotten around to re-jiggering the dependencies between library components to make that happen. They will eventually look like:
```
(define-instance ((Eq :elt) => (Eq (Vector :elt)))
  (define (== v1 v2)
    (and (== (length v1) (length v2))
         (every! (uncurry ==)
                 (zip! (vector-iter v1)
                       (vector-iter v2))))))
```

There's also a tail-recursive formulation without iterators which essentially amounts to inlining `every!` by hand, but I don't want to write that version.)

Independent of the answers to those questions, my reasoning is as follows:
1. The current versions of `when` and `unless` privilege statement-oriented, side-effecting procedual code and encode support for it in the language; I would rather privilege expression-oriented functional code.
2. Coming from CL, my intuition about `when` and `unless` is that they return the value of their body when it is evaluated, and return nothing when it is not evaluated. The current versions of `when` and `unless` do not do that, and I was initially confused by the difference.

> could you share some examples?

This is the most recent example, but I have written the `(if COND (Some EXPR) None)` pattern before, and I know I will again.

I've been playing around with a Coalton interface for interacting with s-expressions, which will start with a typeclass `LispType :coalton-type` which, after some `Proxy`-based magic to fool the type-inferrer, will basically have a method `is-a?` of type `Lisp-Object -> Boolean`. In terms of that, there's a function `try-coerce` of type `(LispType :coalton-type) => Lisp-Object -> (Optional :coalton-type)` (again, after `Proxy` magic). Stripping away the `Proxy` crap, its body wants to look like:

```
(define (try-coerce obj)
  (when (is-a? obj)
    (unsafe-coerce obj)))
```

I think I also ran into a case once while writing a function that mapped integers in the range 0..=9 to their ascii representations, which wanted to be:

```
(define (ascii-digit n)
  (when (and (plus? n) (<= n 9))
    (code-char (+ n (char-code #\0)))))
```

> Also the proposed ones should probably use pure/empty instead of Some/None.

I am concerned that this would worsen the type inference problem, which is the whole reason this change isn't just an obvious slam dunk. But if we can show that there aren't useful cases where the `Optional`-only version typechecks but the more general version doesn't, then I'm all for it.

eliaslfox:
Current uses of `when` could definitely be rewritten to use `cond` or `if`, but I find the reduced nesting of `when` easier to read.

> The current versions of when and unless privilege statement-oriented, side-effecting procedural code and encode support for it in the language

Coalton needs to support procedural for performance. I don't want to make writing that code more difficult for ideological reasons.

> my intuition about when and unless is that they return the value of their body when it is evaluated, and return nothing when it is not evaluated

In haskell when/unless [don't return their body](https://hackage.haskell.org/package/base-4.16.0.0/docs/Control-Monad.html#v:when).

> I am concerned that this would worsen the type inference problem
Alternative is a single parameter type class, so it should be about the same.

These seem useful. I'm happy to add either both of them or maybe just a version of `when` under different names.

gefjon:
Could we come to a compromise where the current definitions of `when` and `unless` are not exported from the `coalton` package, but from some library package dedicated to procedural patterns? I don't particularly object to having them exported from the prelude (though I'd prefer to give the expression-oriented Common Lisp-analogous versions the names `when` and `unless`, personally), but championing them as part of the language both feels like it encourages a style that I don't want to be the default, and makes interacting with the `coalton` package more difficult in terms of shadowing.

eliaslfox:
>  but championing them as part of the language both feels like it encourages a style that I don't want to be the default
I want to make it easier to write imperative code not harder. Coalton borrows heavily from Haskell's pure functional style, but the one of projects goals is to create a pragmatic language not necessarily a purely functional one. 

> and makes interacting with the coalton package more difficult in terms of shadowing
What does it make more difficult?

I consider the existing macros to be language syntax. That they are expanded by CL and not Coalton is merely an implementation detail.  Eventually I would like most of them to create regular ast nodes which are then desugared after type checking. I think it makes sense to keep them in `coalton`.

PR #432, opened 2022-03-23
Fix doc generation

eliaslfox:


src/faux-macros.lisp
```diff
@@ -51,6 +51,3 @@
 
 (define-coalton-editor-macro coalton:lisp (type vars &body lisp-expr)
   "An escape from Coalton into the Lisp world.")
-
```
stylewarning, line 54: was this intended?

stylewarning (approved):

src/faux-macros.lisp
eliaslfox: Yeah. It's already a macro in `language-macros.lisp` so it doesn't need an editor macro.

PR #440, opened 2022-03-24
Add Foldable and Traversable

eliaslfox:
Also switches the type signature of fold from

`(:a -> :b -> :b) -> :b -> (List :a) -> :b`

to

`(:b -> :a -> :b) -> :b -> (List :a) -> :b`

The argument order now matches the order of the fold. This is also now consistent with Haskell and `iterator:fold!`.

### DIFF
--- a/src/library/classes.lisp
+++ b/src/library/classes.lisp
@@ -19,6 +19,9 @@
    #:>>
    #:MonadFail #:fail
    #:Alternative #:alt #:empty
+   #:Foldable #:fold #:foldr
+   #:Traversable #:traverse
+   #:sequence
    #:Into
    #:TryInto
    #:Iso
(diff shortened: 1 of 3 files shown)


eliaslfox:
```
COALTON-USER> (coalton (sequence (make-list (Some 1) (Some 2) (Some 3))))
#.(SOME (1 2 3))
```

stylewarning:
this will be a small Lisper foot gun since REDUCE has the original arg order. Maybe we define `reduce` to be a flipped fold?

stylewarning (approved):

eliaslfox:
`reduce` is a right fold by default though, it should have the same argument order as `foldr`.

stylewarning:
> `reduce` is a right fold by default though, it has the same argument order as `foldr`. 

Isn't CL:REDUCE a left fold? I thought it's only a right fold if you pass :FROM-END T to it.

eliaslfox:
Yeah you're right. I added it in #446.

PR #443, opened 2022-03-24
Sort lists with `cl:sort`

eliaslfox:


src/library/list.lisp
```diff
@@ -410,15 +411,15 @@
...
     (sortBy <=> xs))
 
-  ;; NOTE: This is a very inefficient implementation and can be
-  ;;       replaced with a faster stable sorting algorithm.
   (declare sortBy ((:a -> :a -> Ord) -> (List :a) -> (List :a)))
   (define (sortBy cmp xs)
     "Generic version of sort"
-    (foldr (insertBy cmp) Nil xs))
+    (lisp (List :a) (cmp xs)
+      (cl:sort xs (cl:lambda (a b)
```
stylewarning, line 421: SORT is destructive. Is that OK?

stylewarning (changes requested):

src/library/list.lisp
eliaslfox: oops

stylewarning (approved):

PR #453, opened 2022-03-27
add eq instance for unit

gefjon:
I had cause to use this instance recently, and it seems like a fundamentally sane rule-abiding instance.


stylewarning (approved):
Doh!

PR #459, opened 2022-03-29
Unwrappable ext

gefjon:
> rewrite unwrappable
>
> to have a single operator, `unwrap-or-else`, which takes a `fail` continuation as an
> argument; define `with-default`, `unwrap` and `expect` in terms of `unwrap-or-else`.
>
> implementing `unwrap` and `expect` in classes.lisp requires moving `error` from
> builtin.lisp to classes.lisp. i think that possibly `error` should be an actual builtin,
> exported from the `coalton` package, but for now it remains in the standard library and is
> re-exported from builtin.lisp, as it feels more at home there.

`Unwrappable :container` now has a single method, `unwrap-or-else`, with the signature `(Unit -> :elt) -> (:container :elt) -> :elt`. Its first argument is a continuation to be taken on failure.

`unwrap` and `with-default` (renamed for Scheme-ey goodness) are now functions which invoke `unwrap-or-else` internally:
- `unwrap` (and its new friend `expect`) passes a continuation which uses `error` bail out.
- `with-default` passes a continuation which returns the default value.

## Motivation

I wanted to write `expect`, an operator from Rust, which is like `unwrap` but with a user-supplied error message rather than a message generated based on the type. The old definition of `Unwrappable` didn't allow that; this new one does.


eliaslfox (approved):
:shipit:

PR #469, opened 2022-03-31
Add dict hoisting

eliaslfox:


eliaslfox:
Code to bench

This hoists `(Eq (Optional Int))` to the top level. This is a best case scenario. 
```
(coalton-toplevel
  (define (f x)
    (== (Some (Some 1)) (Some (Some 1)))))
```

This hoists `(Eq (Optional :a))` to inside the definition of `f`. This is a more realistic scenario. 
```
(coalton-codegen
  (define (f x)
    (== (Some (Some x)) (Some (Some x)))
    (== (Some (Some x)) (Some (Some x)))))
```

stylewarning (approved):

PR #470, opened 2022-03-31
Extract native test infrastructure

### LINKED ISSUE #468: Extract testing infrastructure for downstream consumers

eliaslfox:
Fixes #468

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -116,6 +116,14 @@
                (:file "iterator")
                (:file "prelude")))
 
+(asdf:defsystem #:coalton/testing
+  :depends-on (#:coalton
+               #:fiasco)
+  :pathname "src/testing/"
+  :serial t
+  :components ((:file "package")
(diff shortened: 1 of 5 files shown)


stylewarning (approved):
one small tip.

This PR seems a little odd though. What about defining a macro called
```
(defmacro coalton-test (define-test name args &body body)
  `(,define-test ,name ,args (coalton (progn ,@body))))
```

then you can do the normal fiasco test definition stuff and do

```
(coalton-testing:coalton-test my-test:define-test ()
  (is (== 2 (+ 1 1)))
```

or so.

Alternatively, maybe just have no special macro at all? 🤷‍♂️

eliaslfox:
It seems like fiasco uses `*package*` by default to determine where to put the test. Without a macro we would need to add `(name :in coalton-tests:fiasco-suite)` to every test.

PR #471, opened 2022-03-31
Name all optimize flags

### LINKED ISSUE #462: name all DECLARE/DECLAIM OPTIMIZEs
Replace all

```
(declare (optimize ...))
```

to

```
(defvar *X* '(...))

(declare (optimize #.*X*))
```

where `*X*` is given a nice name and defined in an early config-like file.

`*X*` can be a generic name like `*fast-unsafe*`, or a specific name like `*coalton-funcall-policy*`.

I'm suggesting this because of lessons learned with previous Lisp codebases, not because I have any specific issue now.

eliaslfox:
Fixes #462

### DIFF
--- a/src/codegen/function-entry.lisp
+++ b/src/codegen/function-entry.lisp
@@ -48,7 +48,7 @@
                  ;; OPTIMIZE flags to remove type checks.
                  `(progn
                     (defun ,application-sym (,applied-function-sym ,@arg-syms)
-                      (declare (optimize (speed 3) (safety 3))
+                      (declare #.coalton-impl:*coalton-optimize*
                                (type (or function function-entry) ,applied-function-sym)
                                (values t))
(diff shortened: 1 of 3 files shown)


src/codegen/function-entry.lisp
```diff
@@ -48,7 +48,7 @@
                  ;; OPTIMIZE flags to remove type checks.
                  `(progn
                     (defun ,application-sym (,applied-function-sym ,@arg-syms)
-                      (declare (optimize (speed 3) (safety 3))
```
stylewarning, line 51: this was safety 3; do you want it 0?

stylewarning (approved):

src/codegen/function-entry.lisp
eliaslfox: Yeah that was a mistake.

PR #472, opened 2022-03-31
Add `(Num :a => Num (Optional :a))`

### LINKED ISSUE #461: The following should work
`(coalton (+ (Some 1) (Some 2))`

eliaslfox:
Fixes #461

### DIFF
--- a/library/optional.lisp
+++ b/library/optional.lisp
@@ -63,6 +63,12 @@
            ((Some _) LT)
            ((None) EQ))))))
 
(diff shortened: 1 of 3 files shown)


stylewarning:
Do other languages do this? I'm surprised if so.

stylewarning (approved):
I think it's a little loony but I could see why it's convenient.

eliaslfox:
It looks like neither Haskell nor Rust do. I really thought they would.

PR #473, opened 2022-03-31
Detect when code is loaded in another mode

### LINKED ISSUE #206: Detect when code is being loaded for two different modes
Right now, we don't allow (as a matter of principle) Coalton code compiled in `development` and `release` modes to interoperate. Probably in the future we can allow it, but we aren't doing it now.

However, we also don't explicitly error. We ought to be able to emit assertions in the compiled Coalton code that check that the mode is what it was compiled for so there's no mode-mismatching. It may be as simple as adding toplevel forms which do
```
(assert *mode* :<MODE-COMPILED-FOR>)
```
or so.

eliaslfox:
Fixes #206

### DIFF
--- a/src/coalton.lisp
+++ b/src/coalton.lisp
@@ -251,6 +251,11 @@ in FORMS that begin with that operator."
                  (if *coalton-skip-update*
                      program
                      `(progn
+                        (eval-when (:load-toplevel)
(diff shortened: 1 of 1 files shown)


stylewarning (approved):
i think you can use eq not equalp

PR #475, opened 2022-04-01
Add early returns

### LINKED ISSUE #439: Add early returns

eliaslfox:
Fixes #439


tests/type-inference-tests.lisp
```diff
@@ -433,3 +433,16 @@
        (coalton:let a_ coalton:= (coalton-prelude:+ a 1))
        a_))
    nil))
+
```
stylewarning, line 436: can we add a test for a return without a surrounding function?

stylewarning (commented):
by and large LGTM; one req.

stylewarning (approved):

PR #479, opened 2022-04-02
Change division symbol

aijony:
Implements #458.

Note: I removed `single/` and `double/` as they are now redundant. However, I can put them back if people want.


eliaslfox (approved):
LGMT

PR #482, opened 2022-04-04
Block compile `coalton-toplevel` forms

### LINKED ISSUE #481: Add block compilation to `coalton-toplevel`
Optionally wrap the output in `start-block/end-block` tags.

eliaslfox:
Fixes #481

Waiting on #469

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -98,9 +98,9 @@
   :author "Coalton contributors (https://github.com/coalton-lang/coalton)"
   :license "MIT"
   :version (:read-file-form "VERSION.txt")
-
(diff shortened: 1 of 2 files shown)


stylewarning:
I think it should happen only in release mode.

Can you rebase and then I'll look?

eliaslfox:
I don't think this needs to be gated on release mode. Block compilation shouldn't break redefinitions because the definitions are already compiled together. It would only break redefinition if a definition was moved to a second `coalton-toplevel` and then the first `coalton-toplevel` wasn't recompiled.

stylewarning (approved):

PR #488, opened 2022-04-06
Avoid package lock violations

eliaslfox:
This avoid package lock violations when reloading the asdf system.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -29,6 +29,7 @@
   :pathname "src/"
   :serial t
   :components ((:file "package")
+               (:file "unlock-package" :if-feature :sb-package-locks)
                (:file "settings")
                (:file "utilities")
                (:file "global-lexical")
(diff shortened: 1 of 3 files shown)


src/coalton.lisp
```diff
@@ -115,33 +115,105 @@ in FORMS that begin with that operator."
...
+                          '*global-environment*)
+                        ,program))
+                 env)))))))))
+
+;;;
 ;;; Coalton Macros
-(defmacro coalton:coalton-toplevel (&body toplevel-forms)
+;;;
+
+(in-package #:coalton)
```
stylewarning, line 190: I don't understand why you're not just in-package coalton-impl and qualifying them coalton symbols as done already

src/debug.lisp
```diff
@@ -1,25 +1,25 @@
-(in-package #:coalton-impl)
+(in-package #:coalton)
 
-(defun print-value-db (&optional package)
```
stylewarning, line 3: same here

this looks very bizarre; there must be a better way than to use this SBCL-specific extension

stylewarning (changes requested):

stylewarning (commented):
I'm happy to approve this, but what do you think about making an explicit files "lock-package" and "unlock-package" both with :if-feature sb-package-locks?

eliaslfox:
I like that

stylewarning (approved):

PR #500, opened 2022-05-04
Replace usages of `ensure-symbol (gensym)` with `gentemp`

eliaslfox:
`*gensym-counter*` is not strictly increasing. Coalton was generating
identifiers with coliding names. This was causing runtime failures.

### DIFF
--- a/src/codegen/optimizer.lisp
+++ b/src/codegen/optimizer.lisp
@@ -113,7 +113,6 @@
 
       (inline-methods env)
 
-      #+this-is-broken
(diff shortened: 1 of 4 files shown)


stylewarning (approved):

src/codegen/optimizer.lisp
```diff
@@ -113,7 +113,6 @@
 
       (inline-methods env)
 
-      #+this-is-broken
```
stylewarning, line 116: most satisfying line @eliaslfox

PR #502, opened 2022-05-06
Fix typeclass context reduction

eliaslfox:


stylewarning:
@eliaslfox Im surprised at the removal of the HNF function and checking. Can you explain that a bit?

stylewarning (approved):

PR #505, opened 2022-05-07
Change predicate defaulting behavior

aijony:
This is my first foray into Coalton's internals, so feedback is appreciated. Here are the changes:  

 - Prevent defaults in let bindings from conflicting with declarations.
  The following failed to type check before:
```
  (declare a (Num :a => :a -> :a))
  (define (a x)
    (let y = 2)
    (+ x y))
```
but
```
  (define (b x)
    (let y = 2)
    (+ x y))
```
still has `(type-of 'b)` = `Integer -> Integer`.

 - Constrain defaults to only integer literals with `defaultInt` instead of `fromInt`.
  `(define (c x) (+ x x))` now has `(type-of 'c)` = `Num :a => :a -> :a`
  `(define (d x) (+ x 1))` still has `(type-of 'd)` = `Integer -> Integer`.
 
 - Prevent defaulting when missing an instance.
    `(define (e x) (/ x 1))` is now valid ~but `(type-of 'e)` = ` ∀ A. (Dividable A A) (DefaultInt A) ⇒ (A → A)`~.

-  Allow defaulting of values in explicitly declared forms.
This now type checks:
```
  (declare f (:a -> :a))
  (define (f x)
    (traceObject "2" 2)
    x)
```

### DIFF
--- a/library/classes.lisp
+++ b/library/classes.lisp
@@ -186,6 +186,12 @@
     (* (:a -> :a -> :a))
     (fromInt (Integer -> :a)))
 
+  (define-class (DefaultInt :a)
+    "Internal class to mark implicit defaulting to integers."
+    (defaultInt (:a -> :a)))
+
+  (define-instance (Num :a => DefaultInt :a)
+    (define (defaultInt x) x))
 
   ;;
   ;; Haskell
--- a/src/ast/parse-form.lisp
+++ b/src/ast/parse-form.lisp
@@ -262,17 +262,25 @@ This does not attempt to do any sort of analysis whatsoever. It is suitable for 
        (pattern-constructor ctor (mapcar #'parse-pattern args))))))
 
 (defun parse-atom (atom)
-  ;; Convert integer literals into fromInt calls. This allows for
+  ;; Convert integer literals into defaultInt calls. This allows for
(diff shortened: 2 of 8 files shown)


eliaslfox (commented):
The compiler changes look good. Could you elaborate on the purpose  of `DefaultInt`?

aijony:
> The compiler changes look good. Could you elaborate on the purpose of `DefaultInt`?

I updated `defaultInt` to speed up the benchmark. Anyways, function `defaultInt` is meant to tag integer literals during parsing with the class `DefaultInt`. Now instead of defaulting `Num :a` to `Integer` we _only_ default `DefaultInt :a` to `Integer`.

Another example in of where this is useful is if one wants to do `(^ x 2)` where `^` takes an `Integral`, because `Integral` is a child of `Num`, it doesn't default 2 to an integer and thus cannot codegen. However, now functions constrained on superclasses of `Num` will default to integers.

Also, I now remove the `DefaultInt` constraint after type checking, so `(type-of 'e)` is now  `∀ A. (DIVIDABLE A A) (NUM A) ⇒ (A → A)`.

aijony:
This is closed until defaulting is better designed #516

PR #506, opened 2022-05-07
Integral class

aijony:
Adds an `Integral` like Haskell, however further defines a `Remainder` which may be something like Gaussian integers. Also takes `mod`, `even?`, `odd?`, `gcd`, and `lcm` out of `coalton-library/arith` and into `coalton-library/integral`. I'm open to any suggestions.

Depends on #505.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -119,6 +119,7 @@
                (:file "bits")
                (:file "arith")
                (:file "complex")
+               (:file "integral")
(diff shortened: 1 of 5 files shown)


stylewarning:
bike shed: could we maybe call it `LikeInteger` or `Integerlike`?

eliaslfox (approved):

PR #508, opened 2022-05-07
Stdlib Improvements

### LINKED ISSUE #503: Swap order of arguments to `list:index`
They're currently the opposite of `vector:index`.

### LINKED ISSUE #507: `reciprocal` type should be `∀ :A. Dividable :A :A ⇒ (:A → :A)`
With #491 we can now specialize generic functions, meaning we can use the existing `reciprocal :: Fraction → Fraction` implementation for fractions while remaining generic.

eliaslfox:
* generalize reciprocal to `:a -> :a` (fixes #507)
* reverse argument order of `list:index` to match `vector:index` (fixes
  #503)
* add `Foldable` instanes for `Vector` and `Slice`

### DIFF
--- a/library/arith.lisp
+++ b/library/arith.lisp
@@ -454,8 +454,15 @@ The fields are defined as follows:
     (lisp Integer (q)
       (cl:denominator q)))
 
-  (declare reciprocal (Fraction -> Fraction))
-  (define (reciprocal q)
+  (declare reciprocal ((Dividable :a :a) (Num :a) => :a -> :a))
+  (define (reciprocal x)
+    "The multiplicative of a number."
+    (/ 1 x))
+
+  (specialize reciprocal reciprocal-frac (Fraction -> Fraction))
+
+  (declare reciprocal-frac (Fraction -> Fraction))
+  (define (reciprocal-frac q)
     "The reciprocal of a fraction."
     (lisp Fraction (q)
       (cl:/ q)))
--- a/library/arith.lisp
+++ b/library/arith.lisp
@@ -456,7 +456,7 @@ The fields are defined as follows:
 
   (declare reciprocal ((Dividable :a :a) (Num :a) => :a -> :a))
   (define (reciprocal x)
-    "The multiplicative of a number."
+    "The multiplicative inverse of a number."
     (/ 1 x))
 
   (specialize reciprocal reciprocal-frac (Fraction -> Fraction))
--- a/library/slice.lisp
+++ b/library/slice.lisp
@@ -164,6 +164,23 @@
              s1 s2)
             (cell:read out)))))
 
+  (define-instance (Foldable Slice)
+    (define (fold f init s)
+      (lisp :a (f init s)
+        (cl:reduce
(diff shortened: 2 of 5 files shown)


library/slice.lisp
```diff
@@ -33,9 +33,8 @@
   ;; Slice
   ;;
 
-  (repr :transparent)
-  (define-type (Slice :a)
-    (%Slice Lisp-Object))
+  (repr :native (cl:array cl:t (cl:*)))
```
eliaslfox, line 36: @stylewarning can you confirm that this is the correct type annotation for 1 dimensional potentially overlapping arrays of type `t`?

library/arith.lisp
```diff
@@ -459,8 +459,15 @@ The fields are defined as follows:
     (lisp Integer (q)
       (cl:denominator q)))
 
-  (declare reciprocal (Fraction -> Fraction))
-  (define (reciprocal q)
+  (declare reciprocal ((Dividable :a :a) (Num :a) => :a -> :a))
+  (define (reciprocal x)
+    "The multiplicative of a number."
```
stylewarning, line 464: "reciprocal" or "multiplicative inverse"

library/slice.lisp
stylewarning: yes this is correct, though you can simplify it to:

    (vector t)

which would be equivalent.

If you **know** it overlaps, you could write:

    (and (vector t) (not simple-vector))

To be clear, though, if the array/vector could possibly be specialized (like specialized single-float arrays), then you'll want

    (vector *)

or simply

     vector

This is different than `t`. 

`t` says "array of boxed pointers" (roughly)

`*` says "any kind of array/vector, possibly specialized, possibly not"

library/vector.lisp
```diff
@@ -46,9 +46,8 @@
   ;; Vector
   ;;
 
-  (repr :transparent)
-  (define-type (Vector :a)
-    (%Vector Lisp-Object))
+  (repr :native (cl:array cl:t (cl:*)))
```
stylewarning, line 49: see other comment, but this is equivalent to `(cl:vector cl:t)`

We would get speed benefits if we could guarantee this to be a `simple-array` (no displacement, no fill pointers, etc)

library/vector.lisp
```diff
@@ -151,54 +150,54 @@
...
-          (cl:let ((pos (cl:position-if
-                         #'(cl:lambda (x)
-                             (cl:equalp True (coalton-impl/codegen::A1 test x)))
-                         v)))
-            (cl:if pos
-                   (Some pos)
-                   None)))))
+      (lisp (Optional Integer) (v test)
+        (cl:let ((pos (cl:position-if
+                       #'(cl:lambda (x)
```
stylewarning, line 155: no `#'` needed

library/vector.lisp
eliaslfox: We need fill pointers for push and pop. Vectors will never be displaced, but they can be overlapped by slices.

stylewarning (approved):

PR #510, opened 2022-05-10
Add Fractional class

eliaslfox:


stylewarning:
I don't actually see the definition of Fractional in the diff.

library/arith.lisp
```diff
@@ -478,22 +479,39 @@ The fields are defined as follows:
     (define (fromInt z)
       (lisp Fraction (z) z))))
 
+(coalton-toplevel
```
eliaslfox, line 482: @stylewarning  It's here

stylewarning (changes requested):
I'm open to not making this change, but as a math-code writer, I prefer things to be more specific.

eliaslfox:
Closing in favor of the type class hierarchy in #533

PR #526, opened 2022-05-18
Rewrite defaulting

### LINKED ISSUE #512: Add ambiguous instance check
The following should fail

```
(define (x)
  (into (into x))
```

### LINKED ISSUE #516: Make defaulting behave depend on ambiguity
See Typing Haskell in Haskell - 11.5.1

eliaslfox:
Use the defaulting strategy described in THIH. This allow typing more
programs, and allows for defaulting in explicit bindings. Also now
correctly detects ambigious constraints.

Fixes #516
Fixes #512

Defaulting no longer applies to predicate sets like `(Num :a, Num [:a])`
or `(Num :a, Eq [:a])`. This could be improved by extending context
reduction to remove the other predicates before defaulting.


eliaslfox:
@aijony did you intend for `^` to be typed as `(Num :a) (Integral :b) => :a -> :b -> :a`?

aijony:
> @aijony did you intend for `^` to be typed as `(Num :a) (Integral :b) => :a -> :b -> :a`?

Yep, here was my original implementation:
 https://github.com/aijony/coalton/blob/80c7e969653046d83aba21170d948f00b722a86c/library/integral.lisp

aijony:
Defaulting in the coalton macro only seems to work on `Num`:

```
(coalton (+ 2 2))
 ; => 4 (3 bits, #x4, #o4, #b100)

(coalton (even? 2))
 ; => UNABLE-TO-CODEGEN
```
It also doesn't carry through let bindings:
```
(coalton-toplevel
  (define x
    (let ((y (even? 2))) y)))
```
Errors with `Ambigious constraint INTEGRAL #T21564`.

aijony:
You probably also want `(fn (x) (+ x x)` to not be of type `Integer -> Integer`. I imagine Haskell gets away with lazy tricks, but you may only want to default on restricted bindings (or some other notion of valuation): https://github.com/coalton-lang/coalton/commit/a8d79cab57c88cc5ed033efc2aafb8334d1be8b6

I think that the current hard-coded way for defaulting is good enough to merge. While I think having a syntax for declaring defaults would be cool, that is probably a separate issue. However, a compromise I played around with was detecting if a class was a child of `Num`:

```
(defun in-class-tree (env source-class target-class)
  "Checks if the a SOURCE-CLASS is equal to or is a subclass of TARGET-CLASS."
  (cond
    ((eq source-class target-class) t)
    ((null (ty-class-superclasses (lookup-class env source-class))) nil)
    (t (find-if
        (lambda (next-class)
          (in-class-tree env source-class next-class))
        (mapcar #'ty-predicate-class
                (ty-class-superclasses (lookup-class env target-class)))))))
```

eliaslfox:
This seems like it's working now. Adjusting defaulting for restricted bindings was definitely the right call.

Haskell does hardcode a list of default-able numeric classes. Haskell also requires that all predicates containing a variable which will be defaulted are defined in the standard library. I didn't implement this because it seemed overly restrictive.

For example in Haskell:
* `(Eq :a, Show :a, Integral :a)` => `:a -> Integer`
* `(Eq :a, Show :a, Integral :a, MyCustomClass :a)` => no defaulting

eliaslfox:
The following isn't defaulting correctly
```
(coalton (mod 1 2))
```

src/typechecker/derive-type.lisp
```diff
@@ -768,8 +767,15 @@ which have default instances. Currently only resolves Num :a -> Num Integer"
                                              (not (entail env expr-preds p)))
                                            (apply-substitution local-subs preds))))
 
+        ;; NOTE: This is where defaulting happens
+
+        (unless allow-deferred-predicates
+          (setf local-subs (compose-substitution-lists (default-subs env nil reduced-preds) local-subs))
+          (setf reduced-preds (apply-substitution local-subs reduced-preds))
+          (setf expr-types (apply-substitution local-subs expr-type)))
```
aijony, line 775: Should be `expr-type`, `expr-types` is undefined here.

aijony (approved):
Just fix the typo and I think it is good to go! (Although the fact is was working with the typo does concern me).

src/typechecker/derive-type.lisp
eliaslfox: Good catch. It looks like the assignment was unnecessary anyway.

PR #527, opened 2022-05-18
Adds upwards numerical casts

### LINKED ISSUE #348: Add `Into` for smaller -> larger integer types
Add `Into` instances from smaller integer types (like U8) to larger integer types (like U32 or I32). (Note that you can't convert a Un to an In! It must be Un to I(2n) or larger.)

rvs314:
Fixes #348 by adding conversions from smaller numerical types to larger ones


eliaslfox (approved):
Looks great

PR #529, opened 2022-05-18
Implement chain!

### LINKED ISSUE #474: Add `iterator:chain!`

rvs314:
Adds a chain function to fix #474. Let me know if the side-effects of this are implemented as intended.

### DIFF
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -45,6 +45,7 @@
    #:filter!
    #:take!
    #:flatten!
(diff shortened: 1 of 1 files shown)


eliaslfox:
Thanks for adding this. This seems like the right implementation, it's also what [rust does](https://doc.rust-lang.org/stable/src/core/iter/adapters/chain.rs.html#76-81).

eliaslfox (approved):

PR #533, opened 2022-05-20
Implement Trigonometric, Exponentiable and Float

### LINKED ISSUE #366: sqrt and some transcendental functions not implemented
sqrt, exponentiation, exp, log, sin, cos, tan, their inverses, etc. aren't implemented. Binding to Lisp isn't the hard part, but coming up with a good standard library design is.

rvs314:
Implements trigonometric functions, exponentiation and floating points. Attempts to fix #366.
Let me know if there's anything I should change with the implementation. 
Pinging @stylewarning and @aijony to ensure that I got everything right.

### DIFF
--- a/library/arith.lisp
+++ b/library/arith.lisp
@@ -6,7 +6,8 @@
   (:use
      #:coalton
      #:coalton-library/builtin
-     #:coalton-library/classes)
+     #:coalton-library/classes
+     #:coalton-library/utils)
   (:local-nicknames
    (#:bits #:coalton-library/bits))
   (:export
@@ -22,12 +23,24 @@
    #:negate
    #:abs
    #:sign
-   #:expt
    #:ash
    #:mkFraction
    #:numerator
    #:denominator
    #:reciprocal
+   #:Trigonometric
+   #:sin #:cos #:tan
+   #:sinh #:cosh #:tanh
+   #:asin #:acos #:atan
+   #:asinh #:acosh #:atanh
+   #:sincos
+   #:Exponentiable
+   #:log #:expt #:sqrt #:exp #:ln
+   #:Float
+   #:pi #:ee
+   #:RealFloat
+   #:atan2
+   #:rationalize
    #:floor
    #:ceiling
    #:round
@@ -412,13 +425,6 @@ The fields are defined as follows:
         -1
         1))
 
-  (declare expt (Integer -> Integer -> Integer))
-  (define (expt base power)
-    "Exponentiate BASE to a non-negative POWER."
-    (if (< power 0)
-        (error "Can't exponentiate with a negative exponent.")
-        (lisp Integer (base power) (cl:expt base power))))
-
   (declare ash (Integer -> Integer -> Integer))
   (define (ash x n)
     "Compute the \"arithmetic shift\" of X by N. "
@@ -501,6 +507,178 @@ The fields are defined as follows:
     (define (general/ x y)
       (lisp Double-Float (x y)
         (cl:coerce (cl:/ x y) 'cl:double-float)))))
+
+(coalton-toplevel
+  (define-class (Trigonometric :a)
+    (sin   (:a -> :a))
+    (cos   (:a -> :a))
+    (tan   (:a -> :a))
+    (asin  (:a -> :a))
+    (acos  (:a -> :a))
+    (atan  (:a -> :a)))
+
+  (declare sincos (Trigonometric :a => :a -> (Tuple :a :a)))
+  (define (sincos x)
+    (Tuple (sin x) (cos x)))
+
+  (define-class (Exponentiable :a)
+    (expt (:a -> :a -> :a))
+    (log  (:a -> :a -> :a)))
+
+  (declare sqrt ((Into Fraction :a) (Exponentiable :a) => :a -> :a))
+  (define (sqrt x)
+    "Returns the square root of the input"
+    (expt x (into 1/2)))
+
+  (define-class ((Num :a) (Into Fraction :a) (Dividable :a :a) (Trigonometric :a) (Exponentiable :a)
+                 => Float :a)
+    "A finite, floating-point approximation of a real or imaginary number"
+    (ee :a)
+    (pi :a))
+
+  (declare exp ((Float :f) => :f -> :f))
+  (define exp
+    "Raise ee to the power of X"
+    (expt ee))
+
+  (declare ln ((Float :f) => :f -> :f))
+  (define ln
+    "The natural logarithm of a given value"
+    (log ee))
+
+  ;; See http://clhs.lisp.se/Body/f_sinh_.htm
+
+  (declare sinh ((Float :f) => :f -> :f))
+  (define (sinh x)
+    (/ (- (exp x) (exp (negate x))) 2))
+
+  (declare cosh ((Float :f) => :f -> :f))
+  (define (cosh x)
+    (/ (+ (exp x) (exp (negate x))) 2))
+
+  (declare tanh ((Float :f) => :f -> :f))
+  (define (tanh x)
+    (/ (sinh x) (cosh x)))
+
+  (declare asinh ((Float :f) => :f -> :f))
+  (define (asinh x)
+    (ln (+ x (sqrt (+ 1 (expt x 2))))))
+
+  (declare acosh ((Float :f) => :f -> :f))
+  (define (acosh x)
+    (* 2 (ln (+ (sqrt (/ (+ x 1) 2)) (sqrt (/ (- x 1) 2))))))
+
+  (declare atanh ((Float :f) => :f -> :f))
+  (define (atanh x)
+    (/ (- (ln (+ 1 x)) (ln (- 1 x))) (fromInt 2)))
+
+  (define-class ((Float :f) (Ord :f) => RealFloat :f)
+    "Finite, floating-point approximations of real numbers"
+    (rationalize (:f -> Fraction)))
+
+  (declare atan2 ((RealFloat :f) => :f -> :f -> :f))
+  (define (atan2 y x)
+    "Computes the two-argument arctangent of y and x, which is roughly the same as (atan (/ y x)) when defined and accounting for the quadrant of the point (x, y)."
+    (match (Tuple (<=> x 0) (<=> y 0))
+      ((Tuple (GT) _)    (atan (/ y x)))
+      ((Tuple (LT) (LT)) (- (atan (/ y x)) pi))
+      ((Tuple (LT) _)    (+ (atan (/ y x)) pi))
+      ((Tuple (EQ) (GT)) (/ pi (fromint 2)))
+      ((Tuple (EQ) (LT)) (/ pi (fromint -2)))
+      ((Tuple (EQ) (EQ)) 0))))
+
+
+(cl:defmacro %define-real-float-instance (coalton-type underlying-type)
+  "Defines the float instances for a lisp floating-point type"
+  `(coalton-toplevel
+     (define-instance (Trigonometric ,coalton-type)
+       ,(generate-unary-wrapper coalton-type 'sin  'cl:sin)
+       ,(generate-unary-wrapper coalton-type 'cos  'cl:cos)
+       ,(generate-unary-wrapper coalton-type 'tan  'cl:tan)
+       ,(generate-unary-wrapper coalton-type 'asin 'cl:asin :domain '(cl:lambda (x) (cl:<= -1 x 1)))
+       ,(generate-unary-wrapper coalton-type 'acos 'cl:acos :domain '(cl:lambda (x) (cl:<= -1 x 1)))
+       ,(generate-unary-wrapper coalton-type 'atan 'cl:atan))
+
+     (define-instance (Exponentiable ,coalton-type)
+       (define (expt base power)
+         (lisp ,coalton-type (base power)
+           (cl:when (cl:minusp base)
+             (cl:error "Cannot exponentiate with a negative base ~a" base))
+           (cl:unless (cl:or (cl:plusp base) (cl:plusp power))
+             (cl:error "Either the base (~a) or the power (~a) must be positive" base power))
+           (cl:expt base power)))
+       (define (log base number)
+         (lisp ,coalton-type (base number)
+           (cl:when (cl:minusp base)
+             (cl:error "Cannot take a logarithm with the non-positive real base ~a" base))
+           (cl:when (cl:minusp number)
+             (cl:error "Cannot take the logarithm of the non-positive real number ~a" number))
+           (cl:log number base))))
+
+     (define-instance (Into Fraction ,coalton-type)
+       (define (into x)
+         (lisp ,coalton-type (x)
+           (cl:coerce x ',underlying-type))))
+     
+     (define-instance (Float ,coalton-type)
+       (define ee
+         (lisp ,coalton-type ()
+           (cl:exp (cl:coerce 1 ',underlying-type))))
+         
+       (define pi
+         (lisp ,coalton-type ()
+           (cl:coerce cl:pi ',underlying-type))))
+
+     (define-instance (RealFloat ,coalton-type)
+       (define (rationalize x)
+         (lisp Fraction (x)
+           (cl:rationalize x))))))
+
+(%define-real-float-instance Single-Float cl:single-float)
+(%define-real-float-instance Double-Float cl:double-float)
+
+(cl:defmacro %define-integer-expt-instance (coalton-type)
+  `(coalton-toplevel
+     (define-instance (Exponentiable ,coalton-type)
+       (define (expt base power)
+         (lisp ,coalton-type (base power)
+           (cl:when (cl:minusp power)
+             (cl:error "Cannot exponentiate to the power of the negative integer ~a" power))
+           (cl:expt base power)))
+       (define (log b x)
+         ;; The floor of the logarithm with base B > 1 of X >= 1.
+         ;; See GHC's wordLogBase#
+         (let quot =
+           (fn (n m)
+             (lisp ,coalton-type (n m)
+               (cl:values (cl:truncate n m)))))
+         (let ilog-rec =
+           (fn (y)
+             (if (< x y)
+                 (the (Tuple :a :a) (Tuple x 0))
+                 (match (ilog-rec (* y y))
+                   ((Tuple a b)
+                    (if (< a y)
+                        (Tuple a (* 2 b))
+                        (Tuple (quot a y) (+ (* 2 b) 1))))))))
+         (cond
+           ((== x 1) 0)
+           ((< x 1)  (error "Power of ILOG must be greater than or equal to 1."))
+           ((<= b 1) (error "Base of ILOG must be greater than 1."))
+           ((== b 2) (- (lisp ,coalton-type (x) (cl:integer-length x)) 1))
+           (True     (match (ilog-rec b) ((Tuple _ b) b))))))))
+
+(%define-integer-expt-instance Integer)
+(%define-integer-expt-instance U8) 
+(%define-integer-expt-instance I8) 
+(%define-integer-expt-instance U16) 
+(%define-integer-expt-instance I16) 
+(%define-integer-expt-instance U32) 
+(%define-integer-expt-instance I32) 
+(%define-integer-expt-instance U64) 
+(%define-integer-expt-instance I64) 
+(%define-integer-expt-instance Ifix) 
+(%define-integer-expt-instance Ufix) 
 
 (coalton-toplevel
   (define-instance (Into Integer String)
--- a/library/complex.lisp
+++ b/library/complex.lisp
@@ -2,10 +2,11 @@
 ;;;;
 ;;;; Complex numbers
 
-(coalton-library/utils::defstdlib-package #:coalton-library/complex
+(coalton-library/utils:defstdlib-package #:coalton-library/complex
     (:use #:coalton
           #:coalton-library/classes
-          #:coalton-library/arith)
+          #:coalton-library/arith
+          #:coalton-library/utils)
   (:local-nicknames
    (#:arith #:coalton-library/arith))
   (:export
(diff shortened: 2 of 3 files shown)


library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    (**      (:a -> :a -> :a))
+    (logBase (:a -> :a -> :a)))
+
+  (declare sqrt ((Dividable Integer :a) (Exponentiable :a) => :a -> :a))
+  (define (sqrt x)
+    "Returns the square root of the input"
+    (** x (general/ (the integer 1) (the integer 2))))
+
+  (define-class ((Num :a) (Dividable :a :a) (Exponentiable :a) => Float :a)
+    (e  :a)
```
stylewarning, line 508: could we call it `ee` in line with calling the imaginary unit `ii`?

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    (asinh (:a -> :a))
+    (acosh (:a -> :a))
+    (atanh (:a -> :a)))
+
+  (define-class (Exponentiable :a)
+    (**      (:a -> :a -> :a))
+    (logBase (:a -> :a -> :a)))
+
+  (declare sqrt ((Dividable Integer :a) (Exponentiable :a) => :a -> :a))
+  (define (sqrt x)
```
stylewarning, line 503: can we add `sqrt` to the `exponentiable` class? I know it's technically redundant but sqrt is used an awful lot and can be specialized.

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    "Returns the square root of the input"
+    (** x (general/ (the integer 1) (the integer 2))))
+
+  (define-class ((Num :a) (Dividable :a :a) (Exponentiable :a) => Float :a)
+    (e  :a)
+    (pi :a))
+
+  (declare exp ((Float :f) (Exponentiable :f) => :f -> :f))
```
stylewarning, line 511: i would personally add `exp` and `log` to the exponentiable class, again, understanding there can be redunancy

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+  (define (exp x)
+    (** e x))
+
+  (declare log ((Float :f) (Exponentiable :f) => :f -> :f))
+  (define (log x)
+    (logBase e x))
+
+  (define-class ((Float :f) => RealFloat :f)
```
stylewarning, line 519: document this; what's its purpose?

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    (rationalize (:f -> Fraction))))
+
+(cl:defun %define-unary-wrapper (return-type function cl:&optional (package :cl))
+  "defines a function which is a unary wrapper around a corresponding CL function"
+  (cl:let ((x      (cl:gensym))
```
stylewarning, line 524: I don't think you need this gensym stuff or the error checking. This is for local use only anyway,

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    (cl:unless cl-fun
+      (cl:error "Could not find symbol ~s in package ~s" function package))
+    `(define (,function ,x)
+       (lisp ,return-type (,x)
+         (,cl-fun ,x)))))
+
+(cl:defmacro %define-float-instances (coalton-type cl:&optional (package :cl))
+  "Defines the float instances for a lisp floating-point type"
+  (cl:let ((underlying-type (cl:find-symbol (cl:symbol-name coalton-type) package)))
+    (cl:unless underlying-type
```
stylewarning, line 535: see above about not needing error checking

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+           (lisp ,coalton-type (base power)
+             (cl:expt base power)))
+         (define (logBase base number)
+           (lisp ,coalton-type (base number)
+             (cl:log number base))))
+
+       (define-instance (Float ,coalton-type)
+         (define e
+           (lisp ,coalton-type ()
+             (cl:coerce (cl:exp 1) ',underlying-type)))
```
stylewarning, line 563: I would `coerce` the `1`, not the result; you'll lose precision

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+      (cl:error "Could not find symbol ~s in package ~s" coalton-type package))
+    `(coalton-toplevel
+       (define-instance (Trigonometric ,coalton-type)
+         ,(%define-unary-wrapper coalton-type 'sin   package)
+         ,(%define-unary-wrapper coalton-type 'cos   package)
+         ,(%define-unary-wrapper coalton-type 'tan   package)
+         ,(%define-unary-wrapper coalton-type 'sinh  package)
+         ,(%define-unary-wrapper coalton-type 'cosh  package)
+         ,(%define-unary-wrapper coalton-type 'tanh  package)
+         ,(%define-unary-wrapper coalton-type 'asin  package)
```
stylewarning, line 545: beware of domain/range issues...

```
* (asin -2.0)
#C(-1.5707964 1.3169578)
```

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+         (define pi
+           (lisp ,coalton-type ()
+             (cl:coerce cl:pi ',underlying-type))))
+
+       (define-instance (RealFloat ,coalton-type)
+         (define (rationalize x)
+           (cl:coerce (cl:rationalize x) ',underlying-type))))))
+
+(%define-float-instances single-float)
+(%define-float-instances double-float)
```
stylewarning, line 574: complex floats?

library/arith.lisp
```diff
@@ -480,6 +480,99 @@ The fields are defined as follows:
...
+    (tan   (:a -> :a))
+    (sinh  (:a -> :a))
+    (cosh  (:a -> :a))
+    (tanh  (:a -> :a))
+    (asin  (:a -> :a))
+    (acos  (:a -> :a))
+    (atan  (:a -> :a))
+    (asinh (:a -> :a))
+    (acosh (:a -> :a))
+    (atanh (:a -> :a)))
```
stylewarning, line 496: I would also include `atan2` and `sincos`

stylewarning (changes requested):
some comments/thoughts/etc.

library/arith.lisp
rvs314: Would it make sense to use the `specialize` form there?

library/arith.lisp
rvs314: done

library/arith.lisp
stylewarning: I mean the RealFloat class.

library/arith.lisp
rvs314: What would be the best way of solving this? Wrapping the domain around $[-1, 1]$ works for `asin` and `acos`, but not for `asinh` or `acosh`.

library/arith.lisp
rvs314: done

library/arith.lisp
stylewarning: I suppose we can "just" rely on specializing here.

library/arith.lisp
rvs314: I've removed it

library/arith.lisp
rvs314: Done

library/arith.lisp
rvs314: done

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+
+  (declare sincos (Trigonometric :a => :a -> (Tuple :a :a)))
+  (define (sincos x)
+    (tuple (sin x) (cos x)))
```
stylewarning, line 513: nit: `Tuple`

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+
+  (define-class (Exponentiable :a)
+    (**      (:a -> :a -> :a))
```
stylewarning, line 516: you may want to re-use Common Lisp's names `expt` and `log`

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+  (define-class ((Num :a) (Dividable :a :a) (Trigonometric :a) (Exponentiable :a) => Float :a)
+    (ee :a)
+    (pi :a))
+
+  (declare exp ((Float :f) => :f -> :f))
+  (define (exp x)
+    "Raise ee to the power of X"
+    (** ee x))
+
+  (declare log ((Float :f) => :f -> :f))
```
stylewarning, line 533: if you name the one before `log`, you can name this `ln`

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+  (define (log x)
+    "The natural logarithm of a given value"
+    (logBase ee x))
+
+  (define-class ((Float :f) (Ord :f) => RealFloat :f)
+    "Real Floating-Point Numbers"
```
stylewarning, line 539: How about "Floating point numbers that admit an exact rational representation"

library/arith.lisp
stylewarning: only objection is that in Lisp `log` is value then base, not base then value.

I much prefer base then value, as you have

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+    (ee :a)
+    (pi :a))
+
+  (declare exp ((Float :f) => :f -> :f))
+  (define (exp x)
+    "Raise ee to the power of X"
+    (** ee x))
+
+  (declare log ((Float :f) => :f -> :f))
+  (define (log x)
```
stylewarning, line 534: can we be cute and do `(define ln (log ee))` (however the names end up)

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+  (define (sqrt x)
+    "Returns the square root of the input"
+    (** x (general/ (the integer 1) (the integer 2))))
+
+  (define-class ((Num :a) (Dividable :a :a) (Trigonometric :a) (Exponentiable :a) => Float :a)
+    (ee :a)
+    (pi :a))
+
+  (declare exp ((Float :f) => :f -> :f))
+  (define (exp x)
```
stylewarning, line 529: can we be cute and do `(define exp (expt ee))` (or however the names end up)

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+    "The natural logarithm of a given value"
+    (logBase ee x))
+
+  (define-class ((Float :f) (Ord :f) => RealFloat :f)
+    "Real Floating-Point Numbers"
+    (rationalize (:f -> Fraction)))
+
+  (declare atan2 ((RealFloat :f) => :f -> :f -> :f))
+  (define (atan2 y x)
+    "computes the arc tangent of y/x,
```
stylewarning, line 544: Write in capitalized and complete sentences, no line breaks.

> Computes the two-argument arctangent of y and x, which is roughly the same as `(atan (/ y x))` when defined and accounting for the quadrant of the point (x, y).

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+      ((tuple (LT) (LT)) (- (atan (/ y x)) pi))
+      ((tuple (LT) _)    (+ (atan (/ y x)) pi))
+      ((tuple (EQ) (GT)) (/ pi (fromint 2)))
+      ((tuple (EQ) (LT)) (/ pi (fromint -2)))
+      ((tuple (EQ) (EQ)) 0))))
+
+(cl:eval-when (:compile-toplevel :load-toplevel :execute)
+  (cl:defun %define-unary-wrapper
+      (return-type function cl:&key
+       (package :cl)
```
stylewarning, line 557: i would nix these arguments; just double up with two arguments like `'sin 'cl:sin`, no need to match it like this

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
...
+  `(coalton-toplevel
+     (define-instance (Trigonometric ,coalton-type)
+       ,(%define-unary-wrapper coalton-type 'sin   :package package)
+       ,(%define-unary-wrapper coalton-type 'cos   :package package)
+       ,(%define-unary-wrapper coalton-type 'tan   :package package)
+       ,(%define-unary-wrapper coalton-type 'sinh  :package package)
+       ,(%define-unary-wrapper coalton-type 'cosh  :package package)
+       ,(%define-unary-wrapper coalton-type 'tanh  :package package)
+       ,(%define-unary-wrapper coalton-type 'asin  :package package
+                                                   :domain '(cl:<= -1 x 1))
```
stylewarning, line 585: this domain is only valid for real arguments; if you account for complex ones you'll have to define things separately

library/arith.lisp
```diff
@@ -480,6 +493,136 @@ The fields are defined as follows:
       (lisp Double-Float (x y)
         (cl:coerce (cl:/ x y) 'cl:double-float)))))
 
+(coalton-toplevel
+  (define-class (Trigonometric :a)
+    (sin   (:a -> :a))
+    (cos   (:a -> :a))
+    (tan   (:a -> :a))
+    (sinh  (:a -> :a))
```
stylewarning, line 501: all of the hyperbolic functions are "just" different combinations of `(exp x)`. Can we just `define` them and have them constrained on `Exponentiable :t => :t -> :t`? We can `specialize` if we want to call the native Common Lisp variant.

To me, a smaller type class is better, if it's reasonable.

The big question, though, would be whether you want to define `tan` as a functon instead of just

```
(define (tan x)
  (/ (sin x) (cos x)))
```

I personally think it's OK for `tan` to be in the typeclass since you'll want `atan` there too...

stylewarning (changes requested):

library/arith.lisp
rvs314: To your first point, I imagine it would make sense to implement the inverse hyperbolic functions outside the typeclass as well. I'll remove them and add them as functions.

One could argue that a reason that `tan` shouldn't be in the typeclass while `atan` should is because one can be computed "exactly" in terms of other methods in the typeclass (e.g. all implementations would have either `(/ (sin x) (cos x))` or an optimized version thereof for that method), while the other can't be (to my knowledge). In theory the only people who would be exposed to this idiosyncrasy would be people implementing the typeclass, for whom this would mean less work.

This is total bikeshedding, so feel free to tell me to go one way or the other and I'd oblige, but I think it makes sense to keep it outside the typeclass for now.

stylewarning: I'm fine either way.

library/arith.lisp
rvs314: Fixed

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+  (define-class (Exponentiable :a)
+    (expt (:a -> :a -> :a))
+    (log  (:a -> :a -> :a)))
+
+  (declare sqrt ((Into Fraction :a) (Exponentiable :a) => :a -> :a))
+  (define (sqrt x)
+    "Returns the square root of the input"
+    (expt x (into 1/2)))
+
+  (define-class ((Num :a) (Into Fraction :a) (Dividable :a :a) (Trigonometric :a) (Exponentiable :a) => Float :a)
```
stylewarning, line 533: document this class, be sure to say something about commonality / practicality, and that it actually has very little to do with the abstract definition of a float

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+    (ee :a)
+    (pi :a))
+
+  (declare exp ((Float :f) => :f -> :f)) (define exp
```
stylewarning, line 537: newline

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+                                           "Raise ee to the power of X"
+                                           (expt ee))
+
+  (declare ln ((Float :f) => :f -> :f))
+  (define ln
+    "The natural logarithm of a given value"
+    (log ee))
+
+  (declare sinh ((Float :f) => :f -> :f))
```
stylewarning, line 546: add a comment linking to http://clhs.lisp.se/Body/f_sinh_.htm

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+  (define-class ((Float :f) (Ord :f) => RealFloat :f)
+    "Finite, floating-point approximations of real numbers"
+    (rationalize (:f -> Fraction)))
+
+  (declare atan2 ((RealFloat :f) => :f -> :f -> :f))
+  (define (atan2 y x)
+    "Computes the two-argument arctangent of y and x, which is roughly the same as (atan (/ y x)) when defined and accounting for the quadrant of the point (x, y)."
+    (match (tuple (<=> x 0) (<=> y 0))
+      ((tuple (GT) _)    (atan (/ y x)))
+      ((tuple (LT) (LT)) (- (atan (/ y x)) pi))
```
stylewarning, line 579: capital T in Tuple, by convention.

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+         (lisp ,coalton-type ()
+           (cl:exp (cl:coerce 1 ',underlying-type))))
+         
+       (define pi
+         (lisp ,coalton-type ()
+           (cl:coerce cl:pi ',underlying-type))))
+
+     (define-instance (RealFloat ,coalton-type)
+       (define (rationalize x)
+         (lisp fraction (x)
```
stylewarning, line 630: capital F

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+           (cl:coerce (cl:rationalize x) 'cl:ratio))))))
```
stylewarning, line 631: i don't think this coercion is correct (or needed)

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+           (cl:when (cl:< power 0)
+             (cl:error "Cannot exponentiate to the power of the negative integer ~a" power))
+           (cl:expt base power)))
+       (define (log base number)
+         (lisp ,coalton-type (base number)
+           (cl:when (cl:<= base 0)
+             (cl:error "Cannot take a logarithm with the non-positive real base ~a" base))
+           (cl:when (cl:<= number 0)
+             (cl:error "Cannot take the logarithm of the non-positive real number ~a" number))
+           (cl:values (cl:floor (cl:log number base))))))))
```
stylewarning, line 650: this will be error prone for large values... we'll need a more clever algorithm (probably using integer-length somehow)

library/arith.lisp
```diff
@@ -502,6 +508,159 @@ The fields are defined as follows:
...
+
+(%define-integer-expt-instance integer)
```
stylewarning, line 652: capital first letter in type names

library/complex.lisp
```diff
@@ -5,7 +5,8 @@
 (coalton-library/utils::defstdlib-package #:coalton-library/complex
```
stylewarning, line 5: single colon

library/utils.lisp
```diff
@@ -11,3 +12,19 @@
      (cl:defpackage ,name ,@args)
      #+sb-package-locks
      (sb-ext:lock-package ',name)))
+
+(cl:eval-when (:compile-toplevel :load-toplevel :execute)
```
stylewarning, line 16: no need for eval when here since it'll get compiled and loaded earlier

stylewarning (changes requested):

library/arith.lisp
stylewarning: i'm ashamed to admit i don't remember the difference between ^ ^^ **

aijony: - `^ x n` multiplies any `Num` `x` by itself `n` times (`n` is an integer)
- `^^ x n` multiplies or divides any `Dividable` `x` by itself `n` times (`n` is an integer)
- `** x y` exponentiates any algebriac/transcendental numbers `x` and `y`
- `expt` (in common lisp) does all of the above

library/arith.lisp
rvs314: I've implemented Aidan's algorithm

stylewarning (changes requested):

library/arith.lisp
```diff
@@ -502,6 +508,179 @@ The fields are defined as follows:
...
+(cl:defmacro %define-real-float-instance
+    (coalton-type cl:&key
+     (underlying-type (cl:find-symbol (cl:symbol-name coalton-type) :cl)))
+  "Defines the float instances for a lisp floating-point type"
+  `(coalton-toplevel
+     (define-instance (Trigonometric ,coalton-type)
+       ,(generate-unary-wrapper coalton-type 'sin)
+       ,(generate-unary-wrapper coalton-type 'cos)
+       ,(generate-unary-wrapper coalton-type 'tan)
+       ,(generate-unary-wrapper coalton-type 'asin :domain '(cl:lambda (x) (cl:<= -1 x 1)))
```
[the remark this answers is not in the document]
rvs314: I believe it does:

```
CL-USER> (disassemble #'(lambda () (funcall (lambda (x) (fn-call x)) 2)))
; in: DISASSEMBLE #'(LAMBDA () (FUNCALL (LAMBDA (X) (FN-CALL X)) 2))
;     (FN-CALL X)
; 
; caught STYLE-WARNING:
;   undefined function: COMMON-LISP-USER::FN-CALL
; 
; compilation unit finished
;   Undefined function:
;     FN-CALL
;   caught 1 STYLE-WARNING condition
; disassembly for (LAMBDA ())
; Size: 30 bytes. Origin: #x53B22A1C                          ; (LAMBDA ())
; 1C:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; 20:       488945F8         MOV [RBP-8], RAX
; 24:       BA04000000       MOV EDX, 4
; 29:       B902000000       MOV ECX, 2
; 2E:       FF7508           PUSH QWORD PTR [RBP+8]
; 31:       B8E27D4450       MOV EAX, #x50447DE2              ; #<FDEFN FN-CALL>
; 36:       FFE0             JMP RAX
; 38:       CC10             INT3 16                          ; Invalid argument count trap
```

stylewarning (approved):
Rebase into a single commit and we should be good to go!

aijony (approved):

PR #535, opened 2022-05-21
make reverse, append, and concatMap tail-recursive

### LINKED ISSUE #534: Exhausting control stack on compiled function
On SBCL 2.1.9, the following exhausts the default control stack size: `(concatMap perms (combsOf 5 (make-list 1 2 3 4 5 6 7 8)))`.

I think `concatMap` is the culprit here, when I re-evaluate its definition the problem seems to be fixed.

stylewarning:
fixes #534

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -140,14 +140,16 @@
         Nil
         (Cons x (repeat (- n 1) x))))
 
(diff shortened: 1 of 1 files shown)


aijony (approved):
Works like a charm

PR #538, opened 2022-05-21
add new COALTON-LIBRARY/SYSTEM package and gc, time, sleep functions

stylewarning:


eliaslfox (approved):
This looks good. Personally I don't think these are worth importing in the prelude.

stylewarning:
> This looks good. Personally I don't think these are worth importing in the prelude. Especially gc and sleep.

True, I agree, I'll remove them.

stylewarning:
I kept TIME because I find it so useful in day-to-day development, but duly removed GC and SLEEP.

library/prelude.lisp
```diff
@@ -156,7 +156,17 @@
...
+   #:Iterator)
+  (:export
+   #:Iterator)
+
+  (:import-from
+   #:coalton-library/system
+   #:time)
+  (:export
+   #:gc
+   #:time
```
eliaslfox, line 168: I think `gc` and `sleep` shouldn't be exported here

PR #553, opened 2022-05-23
add primes example

stylewarning:
add an example of computing primes using lazy streams


eliaslfox:
Could this be written more idiomatically using iterators? I feel like we already have some of these combinators defined

eliaslfox:
I'm also fine with having this as a self contained example

stylewarning:
It might be nice to have both. This program was just an example to help people understand syntax of Coalton and the like.

eliaslfox (approved):

eliaslfox (approved):

Issue #528, opened 2022-05-18
Remove `Lisp-Object` type


eliaslfox:
It has been superseded by `:repr :native`

eliaslfox:
Before we had repr decelerations it was necessary to embed a `Lisp-Object` in an ADT and let the compiler optimize away the outer ADT in release mode. Now you can instead write

```
(repr :native cl:MyType)
(define-type MyType)
```

This is clearer, and Coalton will generate lisp type annotations for you. If you really want `cl:t` as your type that's still possible with `(repr :native cl:t)`.

PR #556, opened 2022-05-24
file/char-stream i/o interface

gefjon:
A long-awaited interface for reading from and writing to character streams in Coalton!

Features:
- `Input` and `Output` types which wrap `cl:stream`
- `Show` typeclass for printing stuff to `Output` streams, with instances for `String`, `Char` and the various integers
- The four `Encoding`s (external-formats) I cared about while developing, with utf-8 as a default
- `with-input-file!` and `with-output-file!` macros
- Iterators over the `chars!` and `lines!` of an `Input`
- Reading and locally binding (but not setting) `*standard-input*` and `*standard-output*`
- Docstrings
- One (1) test

Still to do:
- External documentation
- More tests
- Filesystem APIs, like `file-exists?` and `generate-temp-file-name`
- More `Show` instances
- Formatting functions, like printing integers as binary or hex, floats in scientific notation, etc

Open questions:
- Is there a better way to encode write errors than returning `(Optional Unit)`?
- Do we care about any other CL i/o functions, like `read-sequence`?
- What errors should we transform into returning `None` (currently: `cl:error`), and what should we leave as unhandled conditions (currently: none, afaik)?
  - Part of me wants to handle only `file-error`.
  - Another part of me wants to handle closed stream errors, which are non-standard and therefore not of type `file-error`.
    - This part is currently winning.
- Byte streams?

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -133,6 +133,7 @@
                (:file "hashtable")
                (:file "monad/state")
                (:file "iterator")
+               (:file "char-io")
                (:file "file")
                (:file "system")
                (:file "prelude")))
--- a/coalton.asd
+++ b/coalton.asd
@@ -133,6 +133,7 @@
                (:file "hashtable")
                (:file "monad/state")
                (:file "iterator")
+               (:file "resource")
                (:file "char-io")
                (:file "file")
                (:file "system")
--- a/coalton.asd
+++ b/coalton.asd
@@ -133,6 +133,7 @@
                (:file "hashtable")
                (:file "monad/state")
                (:file "iterator")
+               (:file "file")
                (:file "system")
                (:file "prelude")))
 
@@ -231,4 +232,5 @@
                (:file "slice-tests")
                (:file "quantize-tests")
                (:file "hashtable-tests")
-               (:file "iterator-tests")))
+               (:file "iterator-tests")
+               (:file "file-tests")))
--- a/library/char-io.lisp
+++ b/library/char-io.lisp
@@ -1,0 +1,211 @@
+(coalton-library/utils:defstdlib-package #:coalton-library/char-io
+    (:use
+     #:coalton
+     #:coalton-library/classes
+     #:coalton-library/builtin)
+  (:local-nicknames (#:iter #:coalton-library/iterator))
+  (:export
+
+   ;; error handling
+   #:UnknownStreamError
+   #:StreamError #:Closed #:EndOfFile #:Unknown
+
+   ;; input files
+   #:Input
+   #:close-input!
+   #:open-input?
+   #:read-char!
+   #:read-line!
+   #:iterator!
+   #:chars!
+   #:lines!
+   #:standard-input
+
+   ;; output files
+   #:Output
+   #:close-output!
+   #:open-output?
+   #:write-char!
+   #:write-string!
+   #:write-line!
+   #:newline!
+   #:standard-output))
+(cl:in-package #:coalton-library/char-io)
+
+;; errors
+
+(coalton-toplevel
+  (repr :native cl:stream-error)
+  (define-type UnknownStreamError)
+  
+  (define-type StreamError
+    Closed
+    EndOfFile
+    (Unknown UnknownStreamError)))
+
+(cl:defmacro with-converting-stream-errors (cl:&body forms)
+  `(cl:handler-case
+       (cl:progn ,@forms)
+     (cl:end-of-file () (Err EndOfFile))
+     #+sbcl (sb-int:closed-stream-error () (Err Closed))
+     (cl:stream-error (e) (Err (Unknown e)))
+     (:no-error (res cl:&rest ignored) ; `cl:read-line' returns an extra value that we don't care about, so ignore it
+       (cl:declare (cl:ignore ignored))
+       (Ok res))))
+
+;; char input streams
+
+(coalton-toplevel
+  (repr :native (cl:and cl:stream (cl:satisfies cl:input-stream-p)))
+  (define-type Input
+    "An input stream, from which characters can be read.")
+
+  (declare open-input? (Input -> Boolean))
+  (define (open-input? in)
+    "Is IN open? Returns `False' if `close-input!' has been called on IN in the past.
+
+If `False', reads from IN will fail."
+    (lisp Boolean (in)
+      (cl:open-stream-p in)))
+
+  (declare close-input! (Input -> Unit))
+  (define (close-input! in)
+    "Close IN, freeing any corresponding resources. Future reads from IN will fail.
+
+It is safe to call `close-input!' on an `Input' that is already closed."
+    (when (open-input? in)
+      (lisp :any (in)
+        (cl:close in))
+      Unit))
+
+  (declare try-close-input! (Input -> (Result StreamError Unit)))
+  (define (try-close-input! in)
+    (if (open-input? in)
+        (Ok (close-input! in))
+        (Err Closed)))
+
+  (declare read-char! (Input -> (Result StreamError Char)))
+  (define (read-char! in)
+    "Read a character from IN.
+
+Blocks if no data is available from IN."
+    (lisp (Result StreamError Char) (in)
+      (with-converting-stream-errors
+        (cl:read-char in cl:t cl:nil cl:nil))))
+
+  (declare read-line! (Input -> (Result StreamError String)))
+  (define (read-line! in)
+    "Read a line up to a newline from IN, discard the newline and return the line.
+
+Blocks if no data is available from IN."
+    (lisp (Result StreamError String) (in)
+      (with-converting-stream-errors
+          (cl:read-line in cl:t cl:nil cl:nil))))
+
+  (declare iterator! ((Input -> (Result StreamError :elt))
+                      -> Input
+                      -> (iter:Iterator (Result StreamError :elt))))
+  (define (iterator! read-elt! in)
+    "Returns an iterator over elements of IN by READ-ELT!"
+    (iter:new
+     (fn ()
+       (if (not (open-input? in))
+           None
+           (match (read-elt! in)
+             ((Ok elt) (Some (Ok elt)))
+             ((Err (EndOfFile)) None)
+             ((Err e) (Some (Err e))))))))
+
+  (declare chars! (Input -> (iter:Iterator (Result StreamError Char))))
+  (define (chars! in)
+    "Returns an iterator over the characters of IN, calling `read-char!' successively.
+
+`next!' may block if no data is available on IN."
+    (iterator! read-char! in))
+
+  (declare lines! (Input -> (iter:Iterator (Result StreamError String))))
+  (define (lines! in)
+    "Returns an iterator over the lines of IN, calling `read-line!' successively.
+
+`next!' may block if no data is available on IN."
+    (iterator! read-line! in))
+
+  (declare standard-input (Unit -> Input))
+  (define (standard-input)
+    "Returns the current standard input stream, i.e. the value of `cl:*standard-input*'."
+    (lisp Input () cl:*standard-input*)))
+
+;; output files
+
+(coalton-toplevel
+  (repr :native (cl:and cl:stream (cl:satisfies cl:output-stream-p)))
+  (define-type Output
+    "An output file, to which characters can be written.")
+
+  (declare open-output? (Output -> Boolean))
+  (define (open-output? out)
+    "Is OUT open? Returns `False' if `close-output!' has been called on OUT in the past.
+
+If `False', writes to OUT will fail."
+    (lisp Boolean (out)
+      (cl:open-stream-p out)))
+  
+  (declare close-output! (Output -> Unit))
+  (define (close-output! out)
+    "Close OUT, forcing its writes to complete and then freeing corresponding resources.
+
+Future writes to OUT will fail.
+
+It is safe to call `close-output!' on an `Output' that has already been closed."
+    (when (open-output? out)
+      (lisp :any (out)
+        (cl:close out)))
+    Unit)
+
+  (declare try-close-output! (Output -> (Result StreamError Unit)))
+  (define (try-close-output! out)
+    "Close OUT, forcing its writes to complete and then freeing corresponding resources.
+
+Returns (Err Closed) if OUT was already closed.
+
+Future writes to OUT will fail."
+    (if (open-output? out)
+        (Ok (close-output! out))
+        (Err Closed)))
+
+  (declare write-char! (Output -> Char -> (Result StreamError Unit)))
+  (define (write-char! out c)
+    "Write C to OUT."
+    (lisp (Result StreamError Unit) (out c)
+      (with-converting-stream-errors
+        (cl:write-char c out)
+        Unit)))
+
+  (declare write-string! (Output -> String -> (Result StreamError Unit)))
+  (define (write-string! out str)
+    "Write STR to OUT."
+    (lisp (Result StreamError Unit) (out str)
+      (with-converting-stream-errors
+        (cl:write-string str out)
+        Unit)))
+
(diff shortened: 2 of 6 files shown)


gefjon:
In response to @stylewarning 's questions on Discord, which I'd like to preserve here:

> Also comments on why streams are a type and not a class?

My goal here was to present an interface for working with CL streams, and specifically character streams as the minimum product. (More on byte streams later.) What I wanted was a way to describe to Coalton, "the type of a CL input stream," and ways of asking Coalton to invoke CL input/output ops on those streams. That necessitates that `file:Input` and `file:Output` both correspond to actual `stream` objects, not to arbitrary implementors of some typeclass.

I wouldn't be opposed to adding a typeclass for character inputs and outputs in the future (or even arbitrary inputs and outputs), but the question then becomes, what is the type that implements `generic-stream:Input` (working title) for I/O to CL streams? `file:Input` is that type.

There's also a world where we could use Gray streams to pass implementors of `generic-stream:Input` back to CL code, but it would necessitate doing some type erasure in order to construct an object that behaves like a CL stream without being passed an instance. And at that point, what you've got is an object that can be used as a `file:Input`.

> do you have any thoughts or comments as to why a stream doesn't encode its underlying element type as a part of the type?

I assume this question to mean, "Why don't we have both `(Input Char)` and `(Input U8)`?" I'll ignore the CL spec's nonsense about subtypes of `character`, because those have never seemed useful to me, and I don't believe there's any plan to offer a Coalton counterpart to `base-char`.

I asked myself this question, and I'm not entirely sure I made the right call, but my thinking was: I've always found it obnoxious in CL that byte-streams and char-streams are interchangable (and indistinguishable to the type system).

I think there are very few operations that are actually generic over the element-type of a stream. Notably, exactly zero of the operations defined by this PR other than `open-*!`, `open-*?` and `close-*!` could apply to a byte stream. In the future, we might define a few operations that could apply to both byte streams and character streams, but these would be either diagnostic, like `read-would-block?`, or somewhat niche ops, like `read-sequence`. In my mind, it feels better to have separate types for byte and char streams, so as to say "byte and char streams behave differently and are not interchangeable," but then to define a typeclass that encodes these shared operations, so as to say, "but they do have these things in common."

I also just really can't imagine ever writing user code that wanted to be generic over the element-type of a stream. Has anyone ever done this?

All that said, I think that even if you agree with all that reasoning, there's a strong case to be made that `file` is a bad name for this package. Probably the package name should include the fact that this is specifically for `Char` streams, like `char-io`. That way, we can later add a package named something like `byte-io`, and then `char-io:Input` and `byte-io:Input` can coexist.

eliaslfox:
A few thoughts

I think Coalton should have an abstraction for file paths instead of just using Strings; probably drawing inspiration from [std::path](https://doc.rust-lang.org/stable/std/path/index.html). I'm unsure whether we should directly wrap lisp paths, or just use byte arrays internally. From the documentation I've been able to find, SBCL paths seem fairly reasonable. However, I'm not sure about other lisp implementations. What I am sure of is that only allowing UTF-8 paths would be a mistake.

Readable and writable files should be the same type. I'm don't think writing to a read only file will be a frequent enough bug to warrant to the increase in API complexity. Also the proposed design doesn't support bidirectional streams. And adding a third set of open/close and a second set of read/write functions seems overarchitected.

The file/stream interface needs be a lot more explicit about error handling. Failure to open a file should return a `Result` with useful information. Closing an already closed file is more likely to be a bug than intentional, and should signal an error. File and stream related functions should only catch signals related to files and streams, not `cl:error`.

As a general rule, functions shouldn't take or return bare lisp objects. Also the lisp object type is scheduled for removal anyway (see https://github.com/coalton-lang/coalton/issues/528). The encoding keyword symbols could instead be represented as

```
(repr :native '(cl:member :ascii :utf-8 :utf-16 :latin-1))
(define-type Encoding-Lisp)
```

I don't want to have macros for opening and closing files. Also I think the bracket pattern functions could be better expressed as a single function.

```
with-open-file! :: (Into FileError :e => Result FileError Stream -> (Stream -> Result :e :a) -> Result :e :a)

;; example usage

(with-open-file! (file:open! "./test.txt")
  (fn (file)
   (trace (file:read-to-string! file)))))
```

The opening functions should have a short version and a long version that takes a config object. 

```
open! :: Into :a FilePath => :a -> Result FileError Stream
open-with! :: Into :a FilePath => :a -> FileOptions -> Result FileError Stream
```

This would give a place to override the defaults for `:if-exists` and `:if-not-exists`.

I think reading`*standard-out*` is fine. But I don't want to support dynamically rebinding it. Coalton currently avoids exposing dyamic scoping entirely and I would like to keep it that way. If desired, similar behavior can be expressed using the reader monad. 

The file interface should expose functions for flusing streams, and calling fsync. The only lisp function related to this I could find was `finish-output`. Calling fsync or the windows equivalent is required for not losing data in the event of a system crash, even if it has to be SBCL specific.

I would prefer to have Show split into another pr. I think it has merits, but it's worth discussing them seperately.

eliaslfox:
> I think finish-output should do the thing you want on Windows. You're right, both finish-output and force-output should be exposed.

I don't think it will. `finish-output` for files appears to be defined [here](https://github.com/sbcl/sbcl/blob/master/src/code/fd-stream.lisp#L288-L339). It does flush, but does not issue an fsync call. I want something more like:

```lisp
(defun flush-file (file)
  (flush-output file)

   #+(and unix (not darwin))
  (sb-posix:fsync (sb-ext:fd-stream-file file))

  #+darwin
  ;; send F_FULLFSYNC ioctl

  #+win32
  ;; do whatever windows does
  )
```

> Working within CL, paths are always going to have to be composed of cl:character elements. There's no way around that. How the underlying implementation translates those to OS encodings is not our concern.

It does seem that this is unfortunately true. I thought SBCL had some way to convert byte arrays into paths, but it turns out I misread the docs. This would seem to imply that files with filenames unrepresentable in UTF-8 (invalid code points, invalid UTF-16 surrogate pairs, ...) are un-openable in SBCL.

> In short, input streams and output streams are different sorts of objects which support different operations. Encoding those differences is what types are for.

They support a disjoint set of operations. If encoding the read/write status of files in the type system is necessary then it can be done with type states.

```
(repr :native 'cl:stream)
(define-type (File :a))

(define-type Read)
(define-type Write)
(define-type BiDirectional)

(define-instance (FileRead Read))
(define-instance (FileRead BiDirectional))

read! :: FileRead :a => File :a -> Result IOError String
```
But that also adds api complexity, and makes it impossible to have a list containing both readonly and bidirectional streams. FWIW rust has blanket implementations of both [`Read`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#impl-Read) and [`Write`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#impl-Write) for files.

> I find it's quite common to say, "I have a stream here, and I want to make sure it's closed, even though it might be already."

This behavior can be easily attained with

```
(when (file:open? file)
  (file:close! file))
```

> Does this work? Are symbols (incl. keywords) allowed to appear as literal objects in Coalton code?

Not at the value level. This is just to more directly express the return type of `encoding-name`. 

> There's no way to do this portably, because the CL spec underdefines the errors signaled by the various stream functions. If we want to restrict ourselves to SBCL, I'm comfortable with it, but I want to document that as a choice we're making.

That's fine. Hash-tables are already SBCL specific. 

> Bidirectional streams do not exist in nature, and I do not believe they are useful.

Sockets are bidirectional streams. 

> Coming from CL, I find writing the fn (file) form to be unnecessary syntactic overhead for no benefit, and would like to support the version that elides that syntactic overhead.

The benefit is giving the compiler a correct parse tree for the purpose of generating useful error messages.

> Also, your proposed version of with-open-file! has different semantics in the presence of currying

Yeah it seems error prone. The following is harder to misuse. This also appears to be how Haskell structures [`withFile`](https://hackage.haskell.org/package/base-4.16.1.0/docs/System-IO.html#v:withFile).

```
open! : Into IOError :e => Path -> FileOptions -> (File -> Result :e :a) -> Result :e :a
``` 

> To make CL interop practical, I need to have some way of using Coalton to invoke CL code with the standard streams locally rebound.

Rebind the streams in lisp.

gefjon:
I'll divide this comment into two sections.

## The juicy issue: bidirectional streams and types for them

> Sockets are bidirectional streams.

I would argue that it's better to model a socket as a pair of (input-stream, output-stream).

> makes it impossible to have a list containing both readonly and bidirectional streams.

It's quite possible to do this, if we provide a function `split-bidirectional-stream` of type `InputOutput -> (Tuple Input Output)`. We could implement this as

```
(repr :native (cl:and cl:stream (cl:satisfies cl:input-stream-p) (cl:satisfies cl:output-stream-p)))
(define-type InputOutput)

(define-instance (Into InputOutput Input)
  (define (into io) (lisp Input (io) io)))

(define-instance (Into InputOutput Output)
  (define (into io) (lisp Output (io) io)))

(declare split-input-output (InputOutput -> (Tuple Input Output)))
(define (split-input-output io)
  (Tuple (into io) (into io)))
```

> FWIW rust has blanket implementations of both [Read](https://doc.rust-lang.org/stable/std/fs/struct.File.html#impl-Read) and [Write](https://doc.rust-lang.org/stable/std/fs/struct.File.html#impl-Write) for files.

Rust is working under a very different set of design constraints than we are. Rust's preference for static dispatch and monomorphization, combined with its object model that equates type with size and privileges pass-by-value means:

- Different sorts of streams (file streams, string streams, buffered streams, sockets, user-defined streams...) must be of different types.
- To provide consistent stream APIs on those different types, stream ops must be defined on typeclasses/traits, not on types.
- Concern with low-level representation makes it easy and practical to represent all file handles as the same type.

That said, I have in Rust had to debug code where library code passed a write-only `File` where user code expected a readable `File` (and vice-versa), and I have wished that the compiler was able to detect it instead of making me run the code and trace down a runtime error.

Because we get boxing and dynamic dispatch from CL, I see no reason to introduce a typeclass where a type is sufficient. And, as I've said before, the ability to pass streams back into CL (and therefore to be able to implement formatting functions in terms of `cl:format`, for example) gives us a strong incentive to represent streams as a type. If we were going to use typeclasses, I think you'd agree that `Input :stream` and `Output :stream` should be separate classes. So it seems to me that, if we want to use types, `Input` and `Output` should be separate types.

Note that `Input` is not the type "a file that has been opened for reading," it's the type "an object from which characters can be read." (Renaming this package from `file` to `char-io` will make this more clear, I think.) Files are one such object, and it's possible to open a file in such a way that you get either an `Input` or an `Output` from it, or both, but I don't think that means we should instead have a single type `Stream` that means "an object from which you might be able to read characters, or to which you might be able to write characters."

If you want a type that means "a file that might be readable or writeable," I can add a `FileHandle` type. If we do that, I still strongly believe that it should not on its own be readable or writeable; it should provide fallible methods which check its mode and convert it into a stream of the appropriate type.

Here's a partial proposal, with a few interesting signatures, and a couple implementations that I find illuminating:

```
(define-type FileHandle)
(define-type Input)
(define-type Output)
(define-type FileOptions)
(define-type Path) ; possibly string, possibly not

(declare with-path (FileOptions -> Path -> FileOptions)) ; should paths be passed separately to `open`?
(declare with-read (FileOptions -> FileOptions))
(declare with-write (FileOptions -> FileOptions))
(declare with-read-write (FileOptions -> FileOptions))
(define (with-read-write opts)
  (with-read (with-write opts)))

(declare file-input (FileHandle -> (Optional Input))
(declare file-output (FileHandle -> (Optional Output))

(open! (FileOptions -> (Result SomeError FileHandle)))
(open? (FileHandle -> Boolean))
(close! (FileHandle -> (Result SomeError Unit))) ; see discussion in next section

;; convenience openers for read-only and write-only files
(open-input! (FileOptions -> (Result SomeError Input)))
(define (open-input! opts)
  (map (compose unwrap file-input) (open! (with-read opts))))
(open-output! (FileOptions -> (Result SomeError Output)))
(define (open-output! opts)
  (map (compose unwrap file-output) (open! (with-write opts))))
```

This adds some complexity to the implementation, but I think makes for a nice API:
- Users who want specifically an `Input` or an `Output` can ask for it.
- Users who want to open a file specifically as input or output don't have to interact with `FileHandle`.
- Users who want to interact with `FileHandle` can see exactly where they're invoking a runtime check for its mode.
- Non-file stream APIs which are only ever either readable or writeable can specify in their type which.
- `Input` and `Output` are concrete types.
  - No need to put typeclass bounds in declarations.
  - Can be passed back to CL.
  - We can still implement `Read` and `Write` typeclasses later, once we work out how to use Gray streams to construct a stream object. Hashtables with custom tests and hash functions do pretty much the same thing.

The one problem I can think of is that it's no longer safe to close an `Input` at the end of its iterator because a corresponding `Output` might still be open, but honestly, doing that was probably wrong anyway.

## Other stuff

> This behavior can be easily attained with
> 
> (when (file:open? file)
>  (file:close! file))

I think that pattern is common enough to put in the library, and one could make the argument going the other way is also trivial, but I can see why we'd want both. What if we had:

```
(declare try-close[ | -input | -output]! ([FileHandle | Input | Output] -> (Result SomeError Unit))
(declare close[ | -input | -output]! ([FileHandle | Input | Output] -> Unit))
(define (close[ | -input | -output]! stream)
  (when (open[ | -input | -output]? stream)
    (unwrap (try-close[ | -input | -output]! stream))))
```

I think it's reasonable to make the infallible version the shorter name and the fallible error-reporting version slightly more verbose, because the code that actually handles the error is going to overshadow the slightly longer name in terms of ergonomics anyway.

> The benefit [to using functions instead of macros] is giving the compiler a correct parse tree for the purpose of generating useful error messages.

Yeah, sure, fair enough. Willco.

> Rebind the streams in lisp.

Sure, I guess. Willco.

> finish-output for files [...] does flush, but does not issue an fsync call. I want [...]

Unfortunately, SB-UNIX does not seem to expose a wrapper around fsync. And I really do not know what Windows does, or how I would find out. If you don't mind, I think it would be best to leave figuring out how to properly sync files as future work.

stylewarning:
`SB-POSIX:FSYNC` exists

stylewarning:
(Side comment: FWIW, `SB-POSIX:FSYNC` exists.)

eliaslfox:
The current character stream implementation seems unintuitive. That multiple input and output streams can be created which alias the same underlying file, with a shared file pointer and open/closed status is unfortunate. Especially when said streams are used to build lazy iterators. Either character streams should have independent file pointers, or they should be discarded entirely. 

Also `FileOptions` should include an ADT for the mode to open a file. Opening a file without reading or writing doesn't make sense. SBCL [just opens it in read mode anyway](https://github.com/sbcl/sbcl/blob/master/src/code/fd-stream.lisp#L2427).

> should paths be passed separately to `open`?

yes

> I think that pattern is common enough to put in the library, and one could make the argument going the other way is also trivial, but I can see why we'd want both.

Either version is *trivial*, but users should need to explicitly opt-in to the version that throws away errors. Closing an already closed file is a programmer bug, not an io error, so it should throw. Also `cl:close` can do file io and errors from that should be returned as a result. See the full implementation [here](https://github.com/sbcl/sbcl/blob/master/src/code/fd-stream.lisp#L2014).

> Unfortunately, SB-UNIX does not seem to expose a wrapper around fsync

[SB-POSIX does](https://koji-kojiro.github.io/sb-docs/build/html/sb-posix/function/FSYNC.html). 

> And I really do not know what Windows does, or how I would find out

Windows uses [`FlushFileBuffers`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/windows/fs.rs#L308). And darwin needs an [`F_FULLFSYNC`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/fs.rs#L905) ioctl.

stylewarning:
There's lots of good discussion, but it doesn't appear this specific PR is going anywhere. Happy to re-open if so.

PR #558, opened 2022-05-24
Add intersperse function on interators

colescott:
Adds the `intersperse!` function which returns alternating elements from iterators


colescott:
Oops. Somehow used the wrong name

stylewarning:
why was this closed / what's wrong with it

stylewarning:
> why was this closed / what's wrong with it 

oh i see interleave was merged

PR #573, opened 2022-05-27
add basic support for Big-Float numbers

stylewarning:
Big-Float is an arbitrary precision floating point number, where the
user may select (or change) the precision with a SET-PRECISION!
call. This unfortunately breaks some expectations and invariants of
classes like the Float class.

This library is built on SB-MPFR, and this library is experimental and
incomplete.

### DIFF
--- a/library/big-float.lisp
+++ b/library/big-float.lisp
@@ -1,0 +1,250 @@
+;;;; big-float.lisp
+;;;;
+;;;; Arbitrary precision float's using SBCL's MPFR library.
+
+(coalton-library/utils:defstdlib-package #:coalton-library/big-float
+  (:use #:coalton
+        #:coalton-library/classes
+        #:coalton-library/arith)
+
+  (:export
+   #:RoundingMode
+   #:rndna
+   #:rndn
+   #:rndz
+   #:rndu
+   #:rndd
+   #:rnda
+   #:rndf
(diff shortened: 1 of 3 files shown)


eliaslfox (changes requested):

library/big-float.lisp
```diff
@@ -0,0 +1,201 @@
+;;;; big-float.lisp
+;;;;
+;;;; Arbitrary precision float's using SBCL's MPFR library.
+
+(coalton-library/utils:defstdlib-package #:coalton-library/big-float
+  (:use #:coalton)
+  (:local-nicknames
+   (#:cls #:coalton-library/classes)
```
eliaslfox, line 8: `:use` these to maintain consistency w/ the rest of the standard library.

eliaslfox (changes requested):

eliaslfox (approved):

stylewarning:
I want to ensure test runners go through OK before merging... they were failing before.

stylewarning:
The issue in the test runners is that it's failing to load libmpfr.so, probably because it's not found.

eliaslfox:
You should be able to `apt-install` it in the build script

stylewarning:
> You should be able to `apt-install` it in the build script

I have done that, and there's still an issue... investigating

PR #579, opened 2022-05-31
Add Rational and Reciprocable classes

aijony:
- Added a `Rational` class
    - The numerator and denominator functions are fairly boring on floats, but would be helpful for more obviously rational types (e.g. fixed-point or dyadics)
    - I left out integers and such out for now, but we can add them if they seem useful
  
-  Added a `Reciprocable` class
    - Simplifies constraints for `/`
    - Easily polymorphic reciprocation
    - Replaces the usage of `RealFloat :a => Into Fraction :a` which could result in floating-point overflows.


library/arith.lisp
```diff
@@ -119,8 +131,20 @@ The fields are defined as follows:
...
+  ;;
+  ;; Rational
+  ;;
+
+  (define-class ((Quantizable :a) (Into :a Fraction) => Rational :a)
+    "Any real number that can be exactly represented by a fraction of integers."
+    (numerator (:a -> Integer))
+    (denominator (:a -> Integer)))
+
+  (declare rational->fraction (Rational :a => :a -> Fraction))
```
eliaslfox, line 145: Is this necessary?

aijony: No, it just seemed like a common enough operation (especially when it was `toFraction`)
Now that I take a closer look, it only saves a single space:

`(the Fraction (into _))`
`(rational->fraction _)`

I will remove it.

PR #588, opened 2022-06-03
move Big-Floats to system coalton/library/big-float

stylewarning:
The introduction of Big-Float has caused issues with many people,
usually because of either of two reasons:

1. The user does not have MPFR installed.

1. The user does have it installed, but paths aren't set up
correctly. Setting up paths, surprisingly, has been sufficiently
non-trivial.

In addition, Big-Float in its current implementation is hopelessly
SBCL specific.

As such, we just move the dependency to a separate system.


eliaslfox (approved):
seems reasonable

gefjon:
I think ideally it should be possible to build docs and run tests on a system that doesn't have MPFR installed. 
I don't think solving that problem should block merging this PR, I just want to get some thoughts about it down.

Unfortunately, I don't see a clean way to do that, as neither SBCL nor SB-MPFR offers an interface to ask, "does this shared object exist in a place SBCL can find it?" at `defsystem` time for a system that depends on SB-MPFR. We could have some packages do `:defsystem-depends-on "sb-mpfr"` and then branch on the `:sb-mpfr` feature, but that feels kinda gross. We could unload docs and tests for library/big-float into their own system, but that seems like a PITA. A real conundrum.

Also, a couple of notes, since I'm only now actually looking at big-float.lisp:

- It would be nice if the read-time error if there's no SB-MPFR feature bound a restart for "I just installed/linked/fixed the MPFR library, reload SB-MPFR, check the feature again and then continue compiling."
- I don't actually think (write-documentation-for-packages](src/doc/generate-documentation.lisp#61) is generating docs for big-float.
- I think `Big-Float` should probably be parameterized by its precision, and have all its method bodies wrapped in a `with-precision`.
- The type should be called `BigFloat`, right?

All that said, I think definitely merge this.

stylewarning:
@gefjon

1. I thought about the `defsystem-depends-on`. I agree that it feels a little gross. In general it would be nice if ASDF had a way to state system prereqs somehow.
2. I think the restart should be added to sb-mpfr proper. I completely agree.
3. I'm sort of divided on how to deal with precision. I definitely don't like this ragtag `set-precision!` or `with-precision` stuff too much. But parameterizing it also has issues: you (and the user) need to decide how to deal with $n$-ary operations (choose min precision? choose max precision? both are valid approaches). The best would be to have a truly computable reals library instead of floats, imho; or to have precision a type-level thing (which is massively inconvenient in Coalton).
4.  It should be, but just keeping in line with the `Single-Float` etc. nomenclature. If/when that changes, `Big-Float` should too.

gefjon:
I have the restart working, and I'll submit a PR on Monday. It's not a big change, and really I'd prefer it if SB-MPFR offered a different interface, but at least there's now a way to interactively see the error, go install MPFR, and continue compilation.

My vision for floats parameterized with precision is just to have `(Big-Float :precision)` be a type (where `:precision` is one of a few types with no members defined in the library, or maybe an implementer of a typeclass so that users can define their own), and to offer binary ops only on `(Big-Float :precision) -> (Big-Float :precision) -> (Big-Float :precision)`. And probably to offer `(Precision :a) (Precision :b) => (Into (Big-Float :a) (Big-Float :b))`.

PR #592, opened 2022-06-04
Error when a constructor is used in multiple types

### LINKED ISSUE #520: Coalton should detect different ADTs with the same constructor
The following should be invalid
```
(coalton-toplevel
 (define-type T1 X)

  (define-type T2 x))
```

eliaslfox:
Fixes #520

### DIFF
--- a/src/typechecker/parse-type-definition.lisp
+++ b/src/typechecker/parse-type-definition.lisp
@@ -163,17 +163,33 @@
                            :for args_ := (apply-ksubstitution ksubs_ args)
                            :for type := (make-function-type* args_ applied-type)
                            :for scheme := (quantify-using-tvar-order (mapcar #'tvar-tyvar tyvar-types) (qualify nil type))
-                           :collect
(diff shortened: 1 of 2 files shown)


src/typechecker/parse-type-definition.lisp
```diff
@@ -163,17 +163,33 @@
...
+                           :collect (cons scheme entry))))
      env)))
 
+(defun check-constructor (ty-name ctor-name env)
+  "Verify that a given constructor isn't used by another type"
+  (declare (type symbol ty-name)
+           (type symbol ctor-name)
+           (type environment env))
+  
+   ;(with-type-context ("define-type of ~S" ty-name)
```
stylewarning, line 187: dead code

stylewarning (approved):

src/typechecker/parse-type-definition.lisp
eliaslfox: thanks

PR #594, opened 2022-06-04
Move arith, integral, and complex to a new math package

aijony:
Rationale:
- `arith.lisp` is getting quite large (I will be breaking it up further in a later PR).
-  Keeps mathematical code into one folder yet separate files
- Internally allows for managing some concept of dependencies, but gives a nice external interface as well

I'm open to naming the package whatever (e.g. `coalton-library/math`, `coalton-library/num`, `coalton-library/arith`). I went with numeric since things like `Functor` and `Semigroup` seem mathematical as well, but it has stuff beyond simple arithmetic.

I believe I remember hearing @eliaslfox  thinking about something like this, so let me know if I am in line with your plans.

### DIFF
--- a/library/prelude.lisp
+++ b/library/prelude.lisp
@@ -1,3 +1,13 @@
+;;;; prelude.lisp
+;;;;
+;;;; Collections of packages
+
+(uiop:define-package #:coalton-library/math
+  (:use-reexport
+   #:coalton-library/math/arith
+   #:coalton-library/math/integral
+   #:coalton-library/math/complex))
+
 (uiop:define-package #:coalton-prelude
   (:use-reexport
    #:coalton-library/classes
@@ -5,7 +15,7 @@
    #:coalton-library/functions)
 
   (:import-from
-   #:coalton-library/arith
+   #:coalton-library/math/arith
    #:Dividable #:/
    #:Fraction
    #:negate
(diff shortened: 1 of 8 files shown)


aijony:
On further consideration - perhaps calling it `coalton-library/math` would be better and `math:` is a nice short prefix.

library/string.lisp
```diff
@@ -122,7 +122,20 @@ does not have that suffix."
...
+    (define (into z)
+      (lisp String (z)
+        (cl:format cl:nil "~D" z))))
+
+  (define-instance (TryInto String Integer)
+    (define (tryInto s)
+      (lisp (Result String Integer) (s)
+        (cl:let ((z (cl:ignore-errors (cl:parse-integer s))))
+          (cl:if (cl:null z)
+                 (Err "String doesn't have integer syntax.")
```
eliaslfox, line 137: Include `s` in the error message

eliaslfox (commented):
I like this. Also I think `math:` is a good prefix.

library/prelude.lisp
```diff
@@ -193,3 +201,8 @@
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-PRELUDE")
+
+(cl:in-package #:coalton-library/numeric)
```
eliaslfox, line 205: Is this needed?

eliaslfox (approved):

PR #595, opened 2022-06-06
Add a restart around the error if SB-MPFR failed to load

gefjon:
Useful for interactive use: this provides me a way to see the error, go "crap, i forgot to
install MPFR," install it, and tell coalton "okay it's there now, look again."

### DIFF
--- a/library/big-float/big-float.lisp
+++ b/library/big-float/big-float.lisp
@@ -36,10 +36,18 @@
 (in-package #:coalton-library/big-float)
 
 #-sbcl (error "This file is hopelessly SBCL specific.")
-#-sb-mpfr (error "SB-MPFR failed to load, for some reason. ~
-                  This is probably due to the shared library ~
-                  not existing, or the system being unable ~
-                  to find it.")
+#+sbcl
+cl-user::
+(eval-when (:compile-toplevel :load-toplevel)
+  (loop :until (uiop:featurep :sb-mpfr)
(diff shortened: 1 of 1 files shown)


stylewarning:
@gefjon Do you think this ought to go upstream into SBCL? I don't mind it here necessarily, but at a glance it seems more appropriate in the SBCL contrib.

gefjon:
Yeah, that'd make sense. I'll go write a patch and submit it to the list.

gefjon:
That would be nice, but I don't believe it's a possibility. Really, SB-MPFR should signal a non-continuable error at load-time (or possibly at compile time?) if MPFR is not present or is too outdated for use. Unfortunately, as best as I can tell, SBCL's make.sh is set to compile, load and test all the contribs unconditionally. If SB-MPFR signaled a hard error, like it really should, when MPFR failed to load, then building SBCL from source on a system without MPFR installed would fail. I think that must be why SB-MPFR only warns and effectively fails silently. I, personally, am far too lazy to dive into SBCL's arcane build system to try and hack it into a state where loading SB-MPFR without MPFR installed is a hard error but won't break building SBCL itself.

Linked below is a patch to SB-MPFR that would do it, but try applying it to the SBCL repo, doing `brew unlink mpfr` (or otherwise making MPFR inaccesible to your system), and then doing `sh make.sh`. It breaks.

https://gist.github.com/gefjon/610fac43974ea996efde1b162c04430f

Issue #449, opened 2022-03-25
Add computable real type


stylewarning:
Neither Common Lisp nor Coalton have arbitrary precision numerical functions (`log`, `sin`, etc.). There are bigints, big rationals, but no big floats. SBCL has an add-on module allowing the use of a library called `mpfr`.

One approach is to add a "computable real" type, which is a type that allows approximation to arbitrary precision. The function `approx` is a hypothetical function that takes a computable real `CReal` and produces a rational that is guaranteed to be correct to a certain integer-number of bits (i.e., it's accurate to `1/2^k`).

```
approx :: CReal -> Integer -> (Fraction Integer)
```

Alternatively, you can just return an `Integer` and implicitly know that you need to divide it by `2^k`. It's slightly more efficient that way since you avoid rationals, avoid GCD, etc.

There's a library that I maintain called [`COMPUTABLE-REALS`](https://github.com/stylewarning/computable-reals) which implements this. Here's a short session.

Recall that `arctan(1)` is `pi/4`:

```
CL-USER> (* 4 (atan 1))
3.1415927
```

Lisp can only compute a floating point approximation as can be seen, which only has a few digits of precision. In the `computable-reals` library, we can use the `*r` and `atan-r` functions:

```
CL-USER> (defvar *pi (computable-reals:*r 4 (computable-reals:atan-r 1)))
*PI
CL-USER> *pi
+3.14159265358979323846...
```

The library tries hard to allow transparent use of Common Lisp rational/real numbers in the interface, so it's not necessary to "cast" numbers like `4` or `1` to a creal data type.

```
CL-USER> (type-of *pi)
COMPUTABLE-REALS::C-REAL
```

Now we can actually compute a rational number that is within `1/2^k` accuracy:

```
CL-USER> (computable-reals:rational-approx-r *pi 100)
124451306656115542615260972311/39614081257132168796771975168
```

Just as a sanity check, we can see that the above fraction is close to `pi`:

```
CL-USER> (float *)
3.1415925
```

As a party trick, let's compute `sin(pi)`, which is zero:

```
CL-USER> (defvar *zero (computable-reals:sin-r *pi))
*ZERO
```

The library conveniently prints a digital representation:

```
CL-USER> *zero
+0.00000000000000000000...
```

Note that this library doesn't "know" that `*zero` is actually zero. The only thing the library knows is that it can ask for an approximation to `k` bits. Let's compute 1000 bits of precision:

```
[...]

PR #597, opened 2022-06-06
Add math/real

aijony:
Creates another math package Real containing rounding/approximation facilities for Real-ish numbers.

- Redefine Quantization
  - Rounding only requires a `cl:truncate` operation instead of a `cl:floor` and `cl:ceiling`
- Use closest even rounding for `round` as that is what `cl:round` does.
- Add Rational and Real class with approximation methods for floats and what not
  - Remove RealFloat

Relevant: #449
Replaces: #579 
Requires: #594

### DIFF
--- a/library/math/real.lisp
+++ b/library/math/real.lisp
@@ -1,0 +1,255 @@
+;;;; real.lisp
+;;;;
+;;;; Numbers that exist on the real number line
+
+(coalton-library/utils::defstdlib-package #:coalton-library/math/real
+    (:use
+     #:coalton
+     #:coalton-library/math/arith
+     #:coalton-library/math/integral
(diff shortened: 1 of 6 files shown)


library/math/real.lisp
```diff
@@ -0,0 +1,256 @@
...
+
+       (define-instance (Rational ,type)
+         (define (to-fraction x)
+           (lisp Fraction (x)
+             (cl:rational x)))
+         (define (best-approx x)
+           (lisp Fraction (x)
+             (cl:rationalize x))))
+
+       #+coalton-release
```
eliaslfox, line 175: Why only on release?

eliaslfox (approved):

PR #599, opened 2022-06-07
Add matching to inline let constructs

rvs314:
This adds pattern matching support to coalton's inline `let` statements. I found this very useful when working with product types.

Example:
```lisp
(coalton 
  (progn
    (let t = (Tuple 1 2)) ; Normal bindings still work
    (let (Tuple a b) = t) ; Destructuring binding
    (+ a b)))
```

A few possible downsides:
- Incongruity between parenthesized `let` and inline `let`
- Only matches on a single case, so non-exhaustive for sum types:

```lisp
(coalton
  (progn
    (let l = (make-list))
    (let (Cons a b) = l)
    a)
```
results in
> Evaluation aborted on #<COMMON-LISP:SIMPLE-ERROR "Pattern match not exaustive error" {1009F4DD03}>.

### DIFF
--- a/src/language-macros.lisp
+++ b/src/language-macros.lisp
@@ -186,8 +186,12 @@ Coalton boolean."
                                   (cl:return-from process
                                     `(coalton::seq
                                       ,@(cl:reverse before-let)
-                                      (coalton:match  ,(cl:fourth form)
-                                        (,(cl:second form) ,(process (cl:nthcdr (cl:+ 1 (cl:length before-let)) forms))))))))
+                                      ,(cl:if (cl:symbolp (cl:second form))
+                                              `(coalton::bind ,(cl:second form) ,(cl:fourth form)
+                                                             ,(process (cl:nthcdr (cl:+ 1 (cl:length before-let)) forms)))
+                                              `(coalton:match ,(cl:fourth form)
+                                                 (,(cl:second form)
+                                                  ,(process (cl:nthcdr (cl:+ 1 (cl:length before-let)) forms)))))))))
                                (cl:t (cl:push form before-let)))))
 
                          ;; There was never a let generate a simple seq


eliaslfox:
I don't want to merge this directly, because functions defined with match expressions won't compile to `cl:labels`:

Current codegen:
```
(coalton-codegen
  (define (f x)
    (let id = (fn (x) x))
      (id x)))

(COMMON-LISP:DEFUN F (X-1)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-1))
   (COMMON-LISP:PROGN
    (COMMON-LISP:LABELS ((ID-2 (X-3) ; <- id defined w/ labels
                           (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-3))
                           X-3))
      (ID-2 X-1))) ; <- direct call to id function
```

Codegen if match was used instead:

```
(COMMON-LISP:DEFUN F (X-7)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-7))
   (TRIVIA.LEVEL2:MATCH (COALTON-IMPL/CODEGEN/FUNCTION-ENTRY:F1
                         (COMMON-LISP:LAMBDA (X-8)
                           (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-8))
                           (COMMON-LISP:BLOCK
                               COALTON-IMPL/CODEGEN/CODEGEN-EXPRESSION::@@LOCAL
                             X-8)))
     (ID-9 (COALTON-IMPL/CODEGEN/FUNCTION-ENTRY:A1 ID-9 X-7)) ; <- slow call to id
     (COALTON-IMPL/CODEGEN/CODEGEN-EXPRESSION::_              ;  currying cannot currently be removed statically
      (COMMON-LISP:ERROR "Pattern match not exaustive error"))))
```

This seems useful, but I think it needs to be implemented in the compiler. I opened #600 for tracking.

eliaslfox:
Yeah that should work

eliaslfox (approved):

PR #603, opened 2022-06-07
Monomorphize functions passed as arguments

### LINKED ISSUE #544: when monomorphizing, also monomorphize global function/value bindings
e.g. consider

```
(coalton-codegen
  (declare square (Num :t => :t -> :t))
  (define (square x) (* x x))

  (declare norm^2 ((Num :t) => (List :t) -> :t))
  (define (norm^2 x)
    (reduce + 0 (map square x)))

  (monomorphize)
  (declare norm^2-df ((List Double-Float) -> Double-Float))
  (define norm^2-df norm^2))
```

In this case, a full polymorphic call to `+` and `square` will still occur.

This would help make monomorphized functions faster.

eliaslfox:
Fixes #544

### DIFF
--- a/src/codegen/monomorphize.lisp
+++ b/src/codegen/monomorphize.lisp
@@ -141,7 +141,7 @@ recompilation, and also maintains a stack of uncompiled candidates."
            (values))
 
   (when (candidate-manager-get candidate-manager candidate)
-    (return-from candidate-manager-push))
+    (return-from candidate-manager-push nil))
 
   (let ((new-name
           (gentemp (concatenate 'string (symbol-name (compile-candidate-name candidate)) "#")
@@ -164,11 +164,12 @@ recompilation, and also maintains a stack of uncompiled candidates."
   (declare (type symbol name)
            (type node-list args)
            (type symbol-list bound-variables)
+           (type tc:environment env)
            (values (or compile-candidate null) &optional))
 
   ;; Only functions with known code are valid candidates
   (unless (tc:lookup-code env name :no-error t)
-    (return-from valid-candidate-p))
+    (return-from valid-candidate-p nil))
 
   (let ((new-args (loop :for node :in args
                     :if (and (node-can-be-propidated node env) (node-free-p node bound-variables))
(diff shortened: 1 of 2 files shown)


eliaslfox:
```
(coalton-codegen
  (declare square (Num :t => :t -> :t))
  (define (square x) (* x x))

  (monomorphize)
  (declare norm^2 (List Double-Float -> Double-Float))
  (define (norm^2 x)
    (reduce + 0 (map square x))))

(COMMON-LISP:DEFUN NORM^2 (X-2187)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-2187))
   (|REDUCE#2189| ARITH::|INSTANCE/NUM DOUBLE-FLOAT-+|
    (ARITH::|INSTANCE/NUM DOUBLE-FLOAT-FROMINT| 0)
    (LIST::|INSTANCE/FUNCTOR LIST-MAP| |SQUARE#2188| X-2187)))
```

src/codegen/monomorphize.lisp
```diff
@@ -182,8 +183,14 @@ recompilation, and also maintains a stack of uncompiled candidates."
 
     ;; Fully propigating arguments could change the ordering of side effects
     ;; Exit early to avoid this
+    ;;
+    ;; Unless the function is known to have more arguments than are being applied
     (when (every #'propagatable-p new-args)
-      (return-from valid-candidate-p))
+      (let ((function (tc:lookup-function env name :no-error t)))
+        (unless function
+          (return-from valid-candidate-p))
```
stylewarning, line 191: can you explicitly return NIL here and in the other cases ?

stylewarning (approved):
one small req

src/codegen/monomorphize.lisp
eliaslfox: Is that standard style in common lisp? I feel like it just add line noise

stylewarning: I think it's like `when`/`unless` vs `cond`/`if`. If the return value matters (in this case it does), them being explicit is better than implicit. If the return value doesn't matter (a function just for effect), then eliding the return value is OK.

A non-expert Lisp programmer might also return `(return-from f)` as `f` returning `(values)`, which is not correct, but not an unreasonable assumption.

PR #608, opened 2022-06-09
Add context clues to unification and print relevant source code

rvs314:
This adds a few changes to the type checking machinery to provide context for type errors. More specifically, it adds the unparsed source code as a type context, and adds hits to unification errors to the origins of the types being unified. There's definitely some things that could be improved, but these were a few less invasive changes that I've found to improve debug-ability.


eliaslfox (approved):
Awesome

PR #610, opened 2022-06-09
Limit type context stack depth

rvs314:
Currently, we implement the stack containing context results by nesting conditions, but this causes an error for larger programs, as there is a maximum depth to which conditions can be nested. The most immediate fix, to me, is to just limit the depth to which we store contexts - we generally don't have to show the entire definition each time a type error occurs, only the most relevant parts.

I've chosen an arbitrary constant for the context depth, so let me know if I should change it. SBCL's max error nesting depth is 10, so any value we choose has to be less than that.


eliaslfox:
Is there a way we could mark that frames were emitted in the trace, and always grab the frame specifying the toplevel definition?

```
Failed to unify Integer and String
in definition of g
in match on (Some x)
... (frames omitted)
in definition of my-whole-program
in COALTON-TOPLEVEL
```

rvs314:
After running some tests it turns out this approach to preventing overrun doesn't work. I will have another PR tomorrow that does.

PR #613, opened 2022-06-10
Keep error context in a separate stack

rvs314:
This fixes a bug that prevents coalton type errors from containing more than `SB-KERNEL:*MAXIMUM-ERROR-DEPTH*` context frames. It does away with managing contexts by wrapping errors and replaces it with an explicit `*type-error-context-stack*`, which means we can have a virtually unlimited stack depth. This bug was technically in coalton for a while, but my recent addition of source contexts means it's more apparent.

Example code that would fail previously, but doesn't fail (meaning it gives a correct type error) with this patch:
```lisp
(in-package :coalton)

(cl:defun ex (depth)
  (cl:if (cl:zerop depth)
    "Base"
    `(match None ((None) None) ((Some _) ,(ex (cl:1- depth)))))

(cl:eval `(coalton ,(ex 20)))
```


eliaslfox (approved):
nice

PR #617, opened 2022-06-12
add into string methods for single-float and double-floats

yaktender:


library/arith.lisp
```diff
@@ -692,6 +692,17 @@ The fields are defined as follows:
       (lisp String (z)
         (cl:format cl:nil "~D" z))))
 
+  (define-instance (Into Single-Float String)
+    (define (into z)
+      (lisp String (z)
+        (cl:format cl:nil "~D" z))))
```
stylewarning, line 698: `~D` is for (Decimal) printing of integers.

You might instead opt for `prin1-to-string`, which will try to write a `read`able string

stylewarning (changes requested):

library/arith.lisp
yaktender: ok updated

aijony (commented):
Make sure you are on the latest branch as `library/arith.lisp` has been moved. Also, into instances dealing with strings have been moved to `library/string.lisp`.

eliaslfox:
I had to open a new pr to rewrite the commit history. See https://github.com/coalton-lang/coalton/pull/633.

PR #620, opened 2022-06-13
Data letrec

gefjon:
A proof of concept for allowing recursive bindings of non-functions in `let` forms.

# E.g.

```
(coalton-toplevel
  (define-type (MyList :elt) MyNil (MyCons :elt (MyList :elt)))

  (declare bar (Unit -> (MyList IFix)))
  (define (bar)
    (let ((x (MyCons 1 x))) x)))
```

# How it works

Conveniently, the typechecker already supported this; it just stuck in an error after type-checking because codegen didn't support it. I stripped out that error, but note to follow because it still needs to be signaled in some cases.

With this change, `codegen-let` has an additional case for SCCs where every binding is a direct application of a constructor. In that case, it allocates "empty" objects (i.e. slots initialized to `nil`) for each of the bindings, then sets their slots with `setf`. For example, the `let` form in the example compiles to something like:

```
(let ((x (mycons nil nil)))
  (setf (slot-value x '_0) 1)
  (setf (slot-value x '_1) x)
  x)
```

# Current shortcomings

- Undocumented.
- Untested.
- Trying to print a cyclical object blows the stack, because our `print-object` methods don't respect `cl:*print-circle*`.
- Recursive `let` binding where the initform is not a direct constructor application gives an unhelpful error message at codegen-time; it should give a useful error message after typechecking, like it used to.
- No support for built-in or `repr :native` types, e.g. `List` and `Vector`, only for default-repr `define-type` constructors.
- No error checking for the above constraint.

It seems like, at the very least, it's important to support `(let ((x (Cons 1 x))) x)`; I had to define a hacky `MyList` type in order to make this work. Do any other `native`-repr types matter?


src/codegen/codegen-expression.lisp
```diff
@@ -22,204 +25,214 @@
...
-        ,@(mapcar
-           (lambda (node)
-             (codegen-expression node current-function env))
-           (node-application-rands node)))))
-
-  (:method ((node node-direct-application) current-function env)
-    (declare (type tc:environment env)
-             (type (or null symbol) current-function))
-    `(,(node-direct-application-rator node)
+(defgeneric codegen-expression (node current-function env))
```
eliaslfox, line 28: Leave this as a single `defgeneric`. That way it reads like a match expression.

src/codegen/compile-expression.lisp
```diff
@@ -126,159 +126,160 @@
...
-             (type tc:environment env)
-             (values node))
-    (let ((qual-ty (tc:fresh-inst (tc:typed-node-type expr))))
-      (assert (null (tc:qualified-ty-predicates qual-ty)))
-      (node-bind
-       (tc:qualified-ty-type qual-ty)
-       (tc:typed-node-bind-name expr)
-       (compile-expression (tc:typed-node-bind-expr expr) ctx env)
-       (compile-expression (tc:typed-node-bind-body expr) ctx env)))))
+(defgeneric compile-expression (expr ctx env))
```
eliaslfox, line 129: Same for this one

gefjon:
Ugh, what a pain; this fails on release mode for two reasons:

1. `struct-or-class` emits `:read-only t` when defining structs for constructors, so `setf` of `slot-value` fails. This is easy to fix.
2. In release mode, single-constructor single-payload types are compiled as "transparent," i.e. they have no actual `defstruct`, their constructor is `identity`, and their accessor is also `identity`. This means that the strategy I used here, of allocating an empty instance and then setf'ing its slots fails; I wind up "constructing" `nil`, and then trying to `(setf (slot-value nil '_0) actual-value)`. I don't see a good fix for this, and in general transparent types seem incompatible with letrec.

src/codegen/codegen-expression.lisp
gefjon: I can aesthetically appreciate wanting it to look like a match expression, but I don't think it's worth the downgrade in developer experience caused by breaking `grep defmethod foo` and `M-.` to jump to method definitions.

gefjon: That said, I don't think I actually ended up needing to edit any of the methods on either of these functions, so these changes don't belong in this PR, and I'm not going to bother making another PR just for this sort of change.

eliaslfox: Jump to definition still works for me with `defgeneric`
Screen Shot 2022-06-13 at 2 26 59 PM
.

gefjon: :shrug: I don't know what to tell you. With SLY on SBCL, the `:method` version always jumps to the top of the `defgeneric` form. But not at issue; I'll strip these changes out of this PR as irrelevant.

gefjon:
I see two ways forward for this:

1. disable the optimization that turns the constructor `Foo` of `(define-type Foo (Foo Bar))` into an identity, and make it always allocate unless some `repr` is specified.
2. be like Rust and introduce a wrapper type `(Cycle :t)` (working title) which encodes "this might be a cyclical data structure," and offer a constructor `make-cycle` with type `(Cycle :t) -> :t`. Now a cyclical list has to be defined like `(define-type (CircleList :elt) CircleNil (CircleCons :elt (Cycle (CircleCons :elt))))`.

eliaslfox:
Alternatively, tag data types that would be new typed in release mode, and don't allow them to be recursively constructed. If recursive definitions are needed, users can opt into de-optimization with `(repr :lisp)`.

gefjon:
> Alternatively, tag data types that would be new typed in release mode, and don't allow them to be recursively constructed. If recursive definitions are needed, users can opt into de-optimization with (repr :lisp).

That version feels bad to me. I'd much prefer opting in to the optimization, i.e. opting out of recursive construction.

eliaslfox:
I'd rather be fast by default. Also I think the newtypes will be used much more frequently than recursive constructor definitions. As long as Coalton generates a descriptive and actionable error message it shouldn't be that inconvenient.

Something like:

```
Unable to recursively define x of type T
  In release mode T will be compiled as a transparent type, and thus cannot be defined recursively
  To disable this behavior add (repr :lisp) to the definition of T
in definition of F
in COALTON-TOPLEVEL
```

gefjon:
> I'd rather be fast by default. Also I think the newtypes will be used much more frequently than recursive constructor definitions.

If you believe that, then I'd rather have the explicit types version. Essentially, if we want `letrec`-style construction to be opt-in, then I think we should go all the way; if we want performance to be opt-out, then I think we should go all the way. It feels silly and error-prone to have `letrec`-style construction to be sometimes-default-but-sometimes-opt-in.

eliaslfox:
> It feels silly and error-prone to have letrec-style construction to be sometimes-default-but-sometimes-opt-in.

I don't understand how the behavior I described is error prone. Binding groups will always be type checked recursively. And how a type is defined will have no impact on scoping rules.

gefjon:
> I don't understand how the behavior I described is error prone.

Making small, seemingly innocuous changes about a type will affect a seemingly unrelated property, whether it's possible to recursively construct the type. Personally, I'd be pretty frustrated if a change like removing a useless variant from a type made cyclic construction break.

In general, I think that when there's a trade-off between performance and ergonomics, we should make ergonomics the default choice, and offer a way for users to opt in to performance instead. `(repr :transparent)` already exists to give users a way to enable this optimization if they care about it. I also think it would be handy to have `define-newtype` syntax anyway, something like `(define-newtype (MyInt IFix))`, which could include a `(repr :transparent)` declaration for the type it defined.

eliaslfox:
I don't think that code should ever be accepted in debug mode that errors in release mode. Anything that could be compiled in debug mode but would error in release mode should signal an error.

I don't really see the trade off as being between performance and ergonomics. I see it as being between what we want to make ergonomic. Either we require explicit annotations in a small number of cases to allow recursive construction, or we require `repr :transparent` on every newtype to not incur a performance overhead for type safety. Of course users who care about performance could always add `repr :transparent`, but it adds visual noise. I also don't like the idea of having a series of "go fast" annotations (see dynamic-extent, slots in python, frozen string literals in ruby) that users of coalton are required to memorize. I want as much as possible, for the compiler to make the best choice by default.

gefjon:
We could just as easily make constructing recursive structures require an explicit `repr :lisp`. I just don't feel comfortable having the default repr do "recursive construction might work, it might not."

eliaslfox:
That's fine with me

stylewarning:
I feel like a REPR shouldn't break Coalton code, only perhaps Coalton-Lisp code (though in a perfect world, that wouldn't break either). If a Coalton library writer decides they don't want REPR, or they change it, then downstream Coalton code  shouldn't change.

I lean toward @gefjon's earlier suggestion: have a dedicated `define-newtype` (but, please, a *much* better name) that doesn't get this letrec benefit, and is always transparent regardless of mode. And thus have `define-type` always cons up a new object (but you can change how this object is concretely represented with REPR).

I think this sort of thing would make *documenting* Coalton APIs easier, and allow us to cleanly divide what features/guarantees that each type definition form allows. If a person changes a type to a newtype or vice versa, it's a clear signal of potential API breakage.

stylewarning:
> On the other hand, if we require explicit `repr` specifications at the `define-type` for any repr-dependent features, then we can say that removing or altering a `repr` specification is a breaking change, but adding one is not.

I guess I'm just not familiar with programming languages where evaluation semantics and valid terms depend on the presence or absence of REPR/PRAGMA/etc. like forms. It seems odd for us to buck that trend.

gefjon:
> I guess I'm just not familiar with programming languages where evaluation semantics and valid terms depend on the presence or absence of REPR/PRAGMA/etc. like forms. It seems odd for us to buck that trend.

Rust is like this to some extent with its `#[repr(C)]` and `#[repr(transparent)]`; there are several things that they leave undefined about the default `#[repr(Rust)]`, though in their case as a systems language it's more about slot layout than `letrec`-ability. But slot layout is definitely a part of their evaluation semantics.

Actually, now that I think about it, Common Lisp doesn't have the language of `repr`, but differences in behavior between structs and classes, specialized and non-specialized arrays, adjustable and non-adjustable arrays, arrays with fill-pointers and those without, varying metaclasses, etc, all feel morally similar to our use of `repr`.

Personally, I'm thinking of `repr` as being a bona-fide part of the Coalton language with defined semantics, and of a `define-type`'s declared `repr` or lack thereof as a part of its definition. I am not thinking of them as PRAGMA-ey might-be-ignored hints to the compiler.

As to the issue at hand, I don't really care whether we preference `letrec` as allowed for the default repr, or we preference `newtype` performance. I see two possibilities, neither of which seems more or less appealing to me:

## The default repr can be recursively constructed by `let`.
  - Defining types that can be constructed recursively is easy, fun, and requires no syntactic overhead.
  - Recursive constructability is an implicit part of a type's interface.
  - `repr :transparent` can no longer be automatically applied by the compiler.
  - Defining no-overhead newtypes requires either syntactic overhead or the introduction of a new defining form.
  - Adding `repr :transparent` to a type of previously unspecified `repr` is a breaking change.

## The default repr cannot be cyclically constructed by `let`, only explicit `repr :lisp` can.
  - Defining types that can be constructed recursively requires syntactic overhead.
  - Recursive constructability is an explicit part of a type's interface.
  - `repr :transparent` can still be automatically applied by the compiler.
  - Defining no-overhead newtimes is easy, fun, and requires no syntactic overhead.
  - Adding `repr :transparent` (or any other `repr`) to a type of previously unspecified `repr` is not a breaking change.

I don't feel like I have any horse in this race; implementing either of these would be pretty easy.

stylewarning:
I am marginally fine with REPR being a part of a definition.

If #2 is the choice then I would want to change the REPR-keyword `:lisp` to `:boxed`, `:opaque`, or something like that.

I do still think it's a design smell to not have the default representation be one that's actually compatible with proposed Coalton syntax (or compatible with Lisp, in a Lisp-embedded language no less). Users will no doubt just guess to check if recursive construction will work, and become frustrated when it doesn't. 

1. Why do we want to be so spartan by default?
2. Why wouldn't we provide a UI that Haskell or OCaml provide? Are we implicitly saying they're wrong?
3. Though it's raising an issue from the dead, why should library authors get to decide what data structures library users get to opt-in to Lisp?

On the last point, I'm imagining somebody writing a beautiful JSON parser, only to not `(repr :lisp)` their stuff, and someone else having to fire up their browser to file an issue saying "please, for the love of all that's mighty, allow me to use your types from Lisp!"

The more I think about it, the more I believe defaulting `(repr :lisp)` is the best choice for the library writer and the library user; follows the principle of least surprise; is the easiest to debug in any mode; and is the most idiomatically Lispiest choice (safe and flexible by default, opt-in options to make rigid and efficient but less featureful/dynamic/debuggable).

gefjon:
> I am marginally fine with REPR being a part of a definition.

> Adding or removing them will always be a breaking change when lisp interop is involved.

Removing or changing a `repr` tag is always going to be a breaking change; there's no way around that. The crossroads we're at is, how defined is the default repr? If it's defined more than "the absence of any repr-specific behavior," then adding a `repr` tag is also a breaking change. I don't think that's a bad thing, but it is a fact that we should be explicit about in discussion.

> If 2 is the choice then I would want to change the REPR-keyword :lisp to :boxed, :opaque, or something like that.

My pitch is to rename `repr :lisp` to `repr :explicit` if we decide on 2, or to `repr :default` if we decide on 1.

> Also there's already a precedent for repr tags changing the allowed operations on a type w/ Addressable.

If we decide in this discussion to define the behavior of the default `repr`, we can always change that.

> Users will no doubt just guess to check if recursive construction will work, and become frustrated when it doesn't.

> On the last point, I'm imagining somebody writing a beautiful JSON parser, only to not (repr :lisp) their stuff, and someone else having to fire up their browser to file an issue saying "please, for the love of all that's mighty, allow me to use your types from Lisp!"

I think these arguments are quite strong, and frankly I think more convincing than Elias' concerns about the performance impact of not optimizing away newtypes. The overhead of explicitly constructing newtypes is small enough that I don't think it will matter a vast majority of the time, and when it does matter, an explicit `repr :transparent` is easy.

eliaslfox:
I ran benchmarks locally, and the performance cost of allocating wrapper structs was much lower than I anticipated. I now think we should go with option 1. I agree with @gefjon that `:default` is a good replacement for `:lisp`.

gefjon:
I've implemented (re-enabled) a check during type-checking to signal an error when attempting to recursively construct forms that aren't recursively constructable, either because they're constructor applications of non-native constructors, or because they're applications of non-constructor functions.

Question: in the `typed-node` ast, are there any compound forms which could include variable references other than `typed-node-abstraction` and `typed-node-application`? If so, additional error checks and test cases should be added.

Also, we (I) should add test cases for:
- Indirect applications.
- Not-fully-saturated constructor applications (I think this may be broken, actually...).

And we (I) should define subclasses of the condition `coalton-impl/typechecker::self-recursive-variable-definition` which describe the different failure modes:
- Direct application of non-constructor.
- Direct application of constructor with non-default repr.
- Not-fully-saturated application of constructor.
- Indirect application.

What am I missing?

eliaslfox:
`validate-bindings-for-codegen` needs to check that sccs with more than one element are either all recursive constructor groups, or all functions. We compile functions out to `cl:labels` blocks, so they can't be mutually recursive with non abstraction values.

Everything else looks really great!

gefjon:
> validate-bindings-for-codegen needs to check that sccs with more than one element are either all recursive constructor groups, or all functions.

I'm not sure how to do this, as the SCCs don't seem to be computed until codegen. I could just compute the SCCs during typechecking and throw them away, which feels a little wasteful, but basically fine. Or I could defer that error until codegen time, which doesn't feel great. I think a really more correct thing to do would be to compute SCCs once during typechecking and edit the various `node` classes to pass the SCCs down through to codegen, but that seems like a lot of work that I don't really want to do.

eliaslfox:
Compute sccs and then throw them away. We previously attached sccs to let nodes, and it was really messy because optimizer passes needed to keep them up to date. Codegen checks this already, but we are setup to generate much better error messages in the type checker.

eliaslfox (approved):
Squash this and it's good to go

PR #622, opened 2022-06-13
Adds an `eql` operator

rvs314:
Adds an `eql` operator, which tests for pointer equality. Akin to the `==` operator in OCaml

### DIFF
--- a/library/builtin.lisp
+++ b/library/builtin.lisp
@@ -61,7 +61,7 @@
 
   (declare eql (:a -> :a -> Boolean))
   (define (eql a b)
-    "Returns True iff A is the same object as B"
+    "Returns True iff A is the same object, number, or character as B"
     (lisp Boolean (a b)
       (to-boolean (cl:eql a b)))))
 
--- a/library/builtin.lisp
+++ b/library/builtin.lisp
@@ -10,7 +10,8 @@
    #:boolean-not
    #:boolean-or
    #:boolean-and
-   #:boolean-xor))
+   #:boolean-xor
+   #:eql))
 
 #+coalton-release
 (cl:declaim #.coalton-impl:*coalton-optimize-library*)
@@ -56,8 +57,13 @@
     "Are X or Y True, but not both?"
     (match x
       ((True) (boolean-not y))
-      ((False) y))))
+      ((False) y)))
 
+  (declare eql (:a -> :a -> Boolean))
+  (define (eql a b)
+    "Returns True iff A is the same object as B"
+    (lisp Boolean (a b)
+      (to-boolean (cl:eql a b)))))
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/BUILTIN")


library/builtin.lisp
```diff
@@ -56,8 +57,13 @@
     "Are X or Y True, but not both?"
     (match x
       ((True) (boolean-not y))
-      ((False) y))))
+      ((False) y)))
 
+  (declare eql (:a -> :a -> Boolean))
+  (define (eql a b)
+    "Returns True iff A is the same object as B"
```
stylewarning, line 64: same object or same number as B.

stylewarning (approved):

eliaslfox (changes requested):
I feel iffy about exposing this.

Values which are converted to new types in release mode will have different behavior in dev/release mode. Also maintaining pointer (in)equality can preclude some optimizations.

eliaslfox:
I think if some types require pointer equality, then they should get type specific equality functions, or just use `cl:eq` in their own `Eq` instances.

rvs314:
> I think if some types require pointer equality, then they should get type specific equality functions, or just use `cl:eq` in their own `Eq` instances.

I can think of cases where both `==` and `cl:eq` are meaningfully but distinctly defined.

> Values which are converted to new types in release mode will have different behavior in dev/release mode.

This seems like the right behavior in this case. The only time when this would occur is when wrapping a value type in a newtype, a case in which `==` should (probably) be the same as `cl:eq`. 

If need be, I can move this to a different area and name it something marking it unsafe, but it's something I think needs to exist.

eliaslfox:
> This seems like the right behavior in this case. The only time when this would occur is when wrapping a value type in a newtype, a case in which == should (probably) be the same as cl:eql.

Here's an example:

```
(coalton-toplevel
  (define-type (Wrapper :a) (Wrapper :a))
   
  (define t (Tuple 1 2))
   
  (define a (Wrapper t))
  
  (define b (Wrapper t))
  
  (define x (eql a b))) ; True in release mode, false in debug mode
```

> it's something I think needs to exist.

I disagree. Pointer equality already has confusing semantics, even without optimizations that would break it further. And it's easy enough call into lisp when necessary. I don't see the need to bless this into the standard library.

stylewarning:
Pointer equality is a practical and useful optimization when the objects you're comparing are drawn from a finite universe imposed by a given problem instance. This type of optimization has the potential to provide polynomial speed-ups on problems without resorting to other solutions which require an obscure or fanciful data structure to achieve the same benefit.

Let me offer an example:

Memoizing a pure function `f` is building a constant-time cache which associates a set of inputs (say `x`) with their computed outputs `f(x)`. You may want to memoize if `f` is recursive and computes lots of identical sub-problems. (That's the whole point of using memoization.)

One might create a hash table which requires the inputs be `Hash` and thus `Eq`. This hash table would be queried before computing the guts of `f`.

For just about every recursive data structure (e.g., list, tree, etc), equality and hashing require at minimum time complexity on the order of the size of said data structure. With only hashing and equality, we cannot achieve constant-time memoization.

On the other hand, with an EQL-like operator (or EQ if you really want to be a pointer purist, which I disagree with), knowing that our universe of inputs will be drawn from some finite universe (say subtrees of a given tree), we can use EQL to memoize and achieve constant-time memoization. 

The only way I know we can avoid making use of EQL is to create a "shadow tree" of unique integers associated with each node:

    (declare label-nodes (Tree :a -> Tree (Tuple :a UFix)))

So, we can just look at the integer labels, and use an array (instead of a hash table) to associate that integer with whatever value we want (say `f(x)`). (Though in this case we won't have identified subtrees that internally share structure, a meager optimization that pops up from time to time.)

But what have we done by using this "shadow tree" approach? We've just re-invented addresses. Every node already has an integer: its address. But with this approach, in the name of referential transparency (or whatever), we've at least tripled the amount of memory (original tree, new tree with double the node sizes).

Using EQL is no less safe, is much more efficient, and is much less verbose than the alternative.

eliaslfox:
I agree that it's a useful optimization. It also can be very easily accessed already with `lisp` forms. 

I would prefer not to expose it in the standard library until we've decided what the semantics of should will be. There are many useful optimizations such as replacing `(map NewTypeConstructor list)` or `(filter (const True) list)` with the original list. Both of those will break pointer equality.

stylewarning:
We can avoid the discussion of pointers (a red herring) and instead think of EQL as something like:

> If `(eql a b)`, then `a` and `b` are observationally equivalent.
>
> Two objects `a` and `b` are *observationally equivalent* if and only if for every (non-`lisp`) Coalton term containing free variable `x`, the evaluation of that term is invariant to the choice of substitution of either `a` or `b` for `x`.

or something like that. Crucially, the implication is one-way; the converse need not be true. This definition means it's technically perfectly valid to

    (define (eql x y) False)

Note that with this definition, assuming a lawful `Eq` (e.g., reflexivity), this means

    (eql a b) ==> (== a b)

(I'd even hazard to say *every* method `==` should first check `eql` as a short-circuit, but that's beside the point.)

gefjon:
> We can avoid the discussion of pointers (a red herring) and instead think of EQL

If we actually want `eql` and not `eq`, then we can ignore discussion of pointers. I'd argue that the polymorphism in `eql` is actually *not* what we want in this case, especially if you want a function that can be a short-circuit to optimize `==` tests. If we use `eql` to do that, then methods for heap-allocated numeric types (complex, double-float, bignums) will duplicate work by first calling `cl:eql`, which will call `cl:=`, and then calling the `==` method, which will also call `cl:=`. This seems undesirable; what you really want to short-circuit `==` methods is "if these are the same object, just return true," not "if these are the same object or numerically equal, just return true."

The issue specifically calls out pointer equality as what they're looking for, which makes me think we want to expose `eq`.

gefjon:
> I would prefer not to expose it in the standard library until we've decided what the semantics of it should will be. There are many useful optimizations such as replacing `(map NewTypeConstructor list)` or `(filter (const True) list)` with the original list. Both of those will break pointer equality.

I think it's perfectly reasonable to define the semantics of `identical?` or `same-object?` or whatever as being just, 

> Returns true if these arguments are the same object (or possibly the same number too, if we want `eql`).
> The compiler is free to remove allocations whenever it deems useful, so some operators may return objects that are `same-object?` to the input. These behaviors may differ between debug and release compilation.

That definition allows you to observe some optimizations, but I wouldn't say it *breaks* them. Essentially, it shifts the burden of verifying whether a no-op transform allocated a new object or not onto the user. I think that's perfectly reasonable, given that `same-object?` is a somewhat unusual operator that I hope will not see widespread usage. I'd also advocate for `same-object?` not being in the prelude.

rvs314:
I believe a different implementation of this based on `Addressable` discussed elsewhere will usurp this PR, so I'm closing this.

Issue #624, opened 2022-06-14
Monomorphizer fails on over-application


aijony:
```
(coalton-toplevel
  (define (f x y)
    (+ x y))

  (declare g (Num :a => :a -> :a))
  (define g
    (f 2))

  (monomorphize)
  (declare h (Integer -> Integer))
  (define (h y)
    (g y)))
```
Errors with:
`The bounding indices 2 and COMMON-LISP:NIL are bad for a sequence of length 1.`

PR #625, opened 2022-06-14
Fix some monorphizer edge cases

aijony:
- Adds some substitution methods
  - `node-bare-abstraction` is hopefully straight foward
  - I can't imagine any substitutions should be taking place in `node-lisp`
- Does _not_ fix a bug where monomorphizing fails on over-application, just changes an instance of over-application in the standard library. https://github.com/coalton-lang/coalton/issues/624
- Fixes a rare case where hoisted variable would be called before it was defined when monomorphizing (`optimizer.lisp`)
  This one I am not so sure about. I don't know if it makes sense to monomorphize _after_ optimizing direct application - perhaps that defeats the purpose - but swapping the order of those does prevent invalid codegen from being generated.

You can reproduce these bugs by monomorphizing `generate-maform-output-with-double` in https://github.com/quil-lang/quilc/pull/819 on SBCL 2.2.5

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -522,14 +522,14 @@
               x xs)))))
 
   (declare maximum (Ord :a => ((List :a) -> (Optional :a))))
-  (define maximum
+  (define (maximum l)
     "Returns a greatest element of a list, or None."
-    (optimumBy >))
+    (optimumBy > l))
 
   (declare minimum (Ord :a => ((List :a) -> (Optional :a))))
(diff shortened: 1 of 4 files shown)


aijony:
> > Fixes a rare case where hoisted variable would be called before it was defined when monomorphizing
> 
> Do you have a minimal example of this I could look at?

Here #626

eliaslfox (approved):
This makes sense now

Issue #619, opened 2022-06-13
Cannot construct recursive alues


rvs314:
In many statically typed languages, values can be constructed recursively, for example in the following ocaml code:

```ocaml
let rec k = 0 :: k
```

The equivalent coalton code doesn't work:

```lisp
(coalton 
  (let ((a (cons 0 a)))
    a))
```

Giving the following error:

```
Variable A-405 cannot be defined recursively.
In (LET ((A (CONS 0 A)))
  A)
In COALTON
   [Condition of type COALTON-IMPL/TYPECHECKER::SELF-RECURSIVE-VARIABLE-DEFINITION]
```

Mutually recursive definitions give a different error:

```lisp
(coalton
  (let ((a (cons 0 b))
        (b (cons 0 a)))
    a))
```

```
Invalid scc binding group. This should have been detected during typechecking.
   [Condition of type COMMON-LISP:SIMPLE-ERROR]
```

karlosz:
Maybe adding `letrec` as a primitive is in order? Unlike scheme, it would have to imply lazy semantics (which I assume OCaml is doing).

rvs314:
> Unlike scheme, it would have to imply lazy semantics (which I assume OCaml is doing).

I don't think this is the case; the following OCaml code fails:

```ocaml
let cons_zero v = 0 :: v
let rec k = cons_zero k
```

I believe they just have a special case for constructors

gefjon:
> it would have to imply lazy semantics (which I assume OCaml is doing).

I don't believe OCaml has laziness for this case; I think they just do what I tried in PR [620]. That strategy is:

- Only allow `letrec` of constructors.
- Expand `letrec x = Ctor a b in body` into `let x = Ctor uninit uninit in (set! (Ctor-a x) a); (set! (Ctor-b x) b); body`

The only thing holding us back from doing that is that, in release mode, Coalton will optimize single-variant single-slot `define-types` to not allocate; i.e. for `(define-type Foo (Foo Bar))`, the constructor `Foo` is the identity function, and the representation of the type `Foo` is the same as `Bar`.

PR #627, opened 2022-06-15
draft interface for `Cycle` functor

gefjon:
Re #619: 

Here's a different way we could support cyclical data structures, inspired by Rust's [Arc::new_cyclic](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.new_cyclic). The short version is, define a functor `Cycle` which encodes circular data, so that a circular list is of type `(Cycle (List :elt))`. This is, IMO, ergonomically much worse than #620, since it requires added type annotations and `map` calls and whatnot. It might also perform worse, since `(Cycle :elt)` has to always allocate. On the other hand, this version doesn't require an optimization regression preventing the implicit `repr :transparent` of newtypes on release mode, and this version doesn't require adding `*print-circle*` support to our `print-object` methods, or introduce the potential to mistakenly try to map a function over a circular list.

Open questions:
- What operations do we need to define to make working with `Cycle` possible?
- Is it possible to define a map operation for `(Cycle (List :elt))` that produces a new circular list rather than looping infinitely?

I still personally prefer #620, but it seemed worth prototyping an alternative for the purposes of discussion.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -138,6 +138,7 @@
                (:file "hashtable")
                (:file "monad/state")
                (:file "iterator")
+               (:file "cycle")
                (:file "system")
                (:file "prelude")))
 
--- a/library/cycle.lisp
+++ b/library/cycle.lisp
@@ -1,0 +1,71 @@
+(coalton-library/utils:defstdlib-package #:coalton-library/cycle
+  (:use
+   #:coalton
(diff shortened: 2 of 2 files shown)


gefjon:
#620 is progressing nicely, so I'll close this draft.

PR #630, opened 2022-06-16
`Addressable` typeclass for object identity comparisons

gefjon:
Per discussion elsewhere, this commit adds a wrapper around `cl:eq` encoded in the `Addressable` type class's `eq?` method.

Except for `repr :native` types, `Addressable` is only implemented by the compiler. This is exciting because we now have a proof-of-concept for how the compiler can automatically define ("derive," in Rust/Haskell lingo) instances for user-defined types.  There are a fair number of internal changes that you can read about in the commit messages, but the user-facing changes are:

- A new type class, `Addressable`, is exported from the package `coalton-library/addressable`.
- `Addressable` has a single method, `eq?`, which is intended to correspond 1:1 with `cl:eq`.
- The compiler will automatically define instances of `Addressable` for `define-type` types which specify either `repr :lisp` or `repr :enum`.
- Types which specify `repr :native` may define instances of `Addressable` for themselves, provided the CL spec defines sensible behavior for `eq` on those types. The sensibility is not enforced, so users are strongly encouraged not to define instances unless they know what they're doing. Defining methods which do something other than call `cl:eq` is punishable by disbarment.
- Types which do not specify a `repr`, or which specify `repr :transparent`, cannot define instances of `Addressable`. This is enforced by a compiler error.
- The library defines `Addressable` instances for `String`, `(List :elt)`, `(Vector :elt)`, `(Cell :content)`, `(Hashtable :key :value)` and `(Slice :elt)`.


src/toplevel-define-type.lisp
```diff
@@ -2,76 +2,137 @@
...
-(defun process-toplevel-type-definitions (deftype-forms repr-table env)
-  (declare (type list deftype-forms)
-           (type environment env)
-           (values type-definition-list
-                   environment))
-
-  ;; Parse type definitions into a list of TYPE-DEFINITION objects
-  (let ((parsed-deftypes (parse-type-definitions deftype-forms repr-table env)))
-    (declare (type type-definition-list parsed-deftypes))
+(defparameter *default-eq?-method*
```
eliaslfox, line 5: Is there a reason to manually construct the `make-instance-definition` here instead of generating code for the instance and processing it w/ the hand written instances?

gefjon: A reason and an after-the-fact-justification:
- This way bypasses the `type-definition-explicit-addressable-p` check and error; I couldn't figure out a way to pass a Coalton source s-expression to `process-toplevel-instance-definitions` and friends that wouldn't trigger the error.
- Generating a source form, parsing it and type-checking it would result in a lot of wasted computation for the compiler.

eliaslfox: I'm not sure about how to best encode the checks. But I would rather keep the compilers internal api surface area small than avoid type checking a one line function. Most of the large structs that are created and eventually fed into a [translation-unit](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/translation-unit.lisp#L3-L10) are only constructed in a single place. This has significantly aided refactoring how data flows through the compiler.

eliaslfox: Also currently no state is stored outside of the global environment, no part of the compiler is called before it's entirely loaded, and intermediate IRs are never stored between compiler invocations. I would like to maintain all of these properties.

gefjon: These are noble goals, and I'd like to live up to them. The only way I can think of right now is introducing a new form `unchecked-internal-define-instance-yes-i-really-mean-it` that parses like a `define-instance`, but bypasses the check for user-defined instances. I don't love that, but it does preserve the data flow invariants, which seem desirable.

eliaslfox: Can you process the two lists of instances separately and add a flag to `process-toplevel-instance-definitions` that skips checks for the compiler generated instances?

```
;; in process-coalton-toplevel
(predeclare-toplevel-instances manual-instances env)
(predeclare-toplevel-instances auto-instances env :auto-instances t)

;; type check code

(process-toplevel-instances manual-instances env)
(process-toplevel-instandes auto-instances env :auto-instances t)
```

We've had to do this in the type checker to handle the differences between local and top level bindings. Seee https://github.com/coalton-lang/coalton/blob/main/src/typechecker/derive-type.lisp#L364-L365.

gefjon: Oh, sure, that makes sense.

src/typechecker/parse-instance-definition.lisp
```diff
@@ -94,7 +105,23 @@
             predicate
             (package-name package)))))
 
-(defun parse-instance-definition (form package env)
+(defun check-for-naughty-explicit-addressable (form predicate class-name env)
```
eliaslfox, line 108: Give this function a more descriptive name, and a doc string

eliaslfox (approved):
This looks great. Make sure to manually squash your commits before merging.

PR #635, opened 2022-06-20
convert list of strings to string

yaktender:


gefjon:
I don't really like this instance. I feel like `(Into :a :b)` is for when there's a single obvious way to convert an `:a` into a `:b`, and we want to give it preference over all other functions `:a -> :b`. The various number to string functions fall into that category, I think, because most of the time you want to print a number, you want to write it in variable-width base ten. I do not think concatenating all the strings in a list falls into that category, though. In the world of `(List String) -> String` functions, I think all of the following are obvious and useful, and none feels more commonplace to the extent that it should be the default:

- Space-separated and wrapped in parens, like the reader wants lists.
- Space-separated but not wrapped in parens.
- Comma-separated.
- Newline-separated.
- This version, smushed together with no separator at all.

So my proposal is, instead of offering an `Into (List String) String` instance, how about we offer a function `str:join` of type `String -> (List String) -> String`, analogous to Python's `string.join`? That is,

```
(define (join separator strs)
  (match strs
    ((Nil) "")
    ((Cons fst rst) (concat (concat fst separator) (join separator rst)))))
```

For performance reasons, we might want to also have `join!` on `String -> (Iterator String) -> String`, and we might want to implement both mutably with a single buffer rather than the naive functional way, which conses and copies a lot of strings

yaktender:
ok, so mean use fold and fold!?

stylewarning:
@yaktender I don't think we will include this change. There are so many ways to convert a list into a string, and not a single one of them is "obviously" the right one, and not a single one of them is even broadly convenient.

Usually that's a pretty strong reason to *not* define a type class instance.

yaktender:
That is fair. I look forward to this https://github.com/coalton-lang/coalton/pull/636.

PR #636, opened 2022-06-20
draft `StringBuilder` interface to encode mutable string operations

gefjon:
It's Java's `StringBuilder`. But in Coalton.

After thinking about #635 , I realized that it would be nice to have a way for the library to implement efficient string-wrangling operations like `join` without allocating and copying a bunch of intermediate strings. We could write all those algorithms in `lisp` blocks, or we could chicken out and offer mutable operations on `String`, but I think I'd rather have a type to encode mutable strings, and pretend `String` is immutable.

Open questions:
- What operations do we want to offer on `StringBuilder`? I wrote just `push-char!` and `push-str!`, but it seems sensible at the very least to offer `capitalize!`, `upcase!` and `downcase!`, since CL makes them super easy for us. There are probably others I'm forgetting, too.
  - I'd prefer not to offer operations for which `(vector character)` has poor performance as an implementation of strings, like left-trim, push-front or insert-at. The point of `StringBuilder` is to be an efficiency hack, so I think it should only offer operations for which it's actually efficient.
- Should this be its own package, or a part of `coalton-library/string`? The latter is appealing because it would make it possible to define string operations (like `join`) in that package in terms of `StringBuilder`, where currently I fear we'd have to have another package downstream of both. Or we could move `StringBuilder` earlier in the library system so that string.lisp depends on it instead of vice-versa.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -142,6 +142,7 @@
                (:file "hashtable")
                (:file "monad/state")
                (:file "iterator")
+               (:file "string-builder")
                (:file "system")
                (:file "prelude")))
 
--- a/library/string-builder.lisp
+++ b/library/string-builder.lisp
@@ -2,11 +2,13 @@
   (:use
    #:coalton
    #:coalton-library/builtin
+   #:coalton-library/functions
    #:coalton-library/classes)
   (:local-nicknames (#:addr #:coalton-library/addressable)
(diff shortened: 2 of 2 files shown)


stylewarning:
Closing due to inactivity. Happy to re-open if desired!

PR #640, opened 2022-06-22
Add type aliases

rvs314:
A simple implementation of type aliases. There are a few desiderata that I haven't included (aliases can't have type variables or constraints, and I'm not 100% sure how naming should work given our current shadowing convention), but it works for the use-case I needed them for.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -88,6 +88,7 @@
                              (:file "program")
                              (:file "package")))
                (:file "toplevel-define-type")
+               (:file "toplevel-define-alias")
                (:file "toplevel-declare")
                (:file "toplevel-define")
                (:file "toplevel-define-instance")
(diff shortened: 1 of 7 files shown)


eliaslfox:
To be more generally useful, type aliases should support parametric polymorphism. 

Instead of being defined before types, their definition should be interspersed with type definition checking based on dependence. See https://github.com/coalton-lang/coalton/blob/main/src/typechecker/parse-type-definition.lisp#L297-L305 for where type definition SCCs are checked. 

Also:
* type aliases should error at definition time, not expansion time if they are invalid
* parametric type aliases should not be poly kinded
* applied type aliases should not need to be kind `*`
* type aliases must always be fully applied

This should work:
```lisp
(define-alias (Ap :f :a) (:f :a))

(define-alias Maybe Optional)

(declare x (Ap Maybe String))
(define x (Some "hello"))
```

This should not work:
```lisp
(define-alias (Ap :f :a) (:f :a))

(define-alias (Maybe :a) (Optional :a))

(declare x (Ap Maybe String))
(define x (Some "hello"))
```

rvs314:
This is going to require some non-trivial changes, so I'm going to close this PR and re-open it when the changes are implemented.

Issue #631, opened 2022-06-17
Optimize curried top-level function applications


stylewarning:
Right now, suppose we have

```
(define (f x1 x2 ... xn) ...)
```

Then the call

```
(f a1 a2 ... ak)
```

for `k < n` will allocate `k` intermediate closures.

We already optimize the case when the call is fully saturated (i.e., `k = n`).

This issue concerns itself with optimizing top-level functions getting called (not local functions getting called).

karlosz:
Let's break down what is actually suboptimal with the current partial application implementation:

If we partially apply a function entry object `f` (representing a first class Coalton function use) of arity `N` with `M` (< N) arguments, the way we represent function entry objects now means that this partial application is done by calling a nested Lisp lambda repeatedly. This means we currently end up having to do `M` function calls and `M` closure environment allocations to unpeel each layer of nesting. The resulting closure, `f_p`, when supplied with the remaining `M - N` arguments, still has to partially apply away the arguments one-by-one. The only time the saturation optimization kicks in at the moment is when the original function entry object `f` is supplied all of its arguments, and is never used for the intermediate partially applied functions.

I think we could avoid the unpeeling all together by modifying the `FUNCTION-ENTRY` structure to contain more entry points and ditching the direct use of Lisp lambdas.

Issue #632, opened 2022-06-17
Optimize curried local function calls


stylewarning:
Similar to #631, but for local functions. This is a more difficult for a variety of reasons.

PR #653, opened 2022-06-29
 Use jump tables to eliminate under- and over-application overhead. 

karlosz:
Before this change, if we partially apply a function entry object
`f` (representing a first class Coalton function use) of arity `N` with
`M` (< `N`) arguments, this partial application is done by calling a
nested Lisp lambda repeatedly. This means we currently end up having
to do `M` function calls and `M` closure environment allocations to
unpeel each layer of nesting. The resulting closure, `f_p`, when
supplied with the remaining `M - N` arguments, still has to partially
apply away the arguments one-by-one. The only time the saturation
optimization kicks in at the moment is when the original function
entry object `f` is supplied all of its arguments, and is never used
for the intermediate partially applied functions. The same applies for
an over-saturated function call: Closures are applied and created one
by one for each argument.

This change avoids the unpeeling altogether by modifying the
`FUNCTION-ENTRY` structure to contain entry points for each possible
argument count (in esssence a jump table), ditching the direct use of
Lisp lambdas. We now no longer allocate any more extra closures than
are ever explicitly created by functional partial applications, and
partially evaluated functions are now also eligible to be
saturated. Unlike the old way, there is no asymmetry with respect to
the optimizations applicable to partially evaluated functions and
functions defined elsehow.

To achieve this, we create a function entry object on every under- or
over- application, instead of specially-handled nested Lisp
closures. Because the jump table used for currying is stored as a
load-time constant with `LOAD-TIME-VALUE`, the overhead for creating a
function entry object consists of consing the structure and setting
the fields, which is a constant time overhead similar to that of
allocating a closure environment. Hence, this is pretty much a strict
improvement over the old way of doing things, while being much more
efficient for a larger number of arguments. Overapplication is also
more efficient because we saturate the function all at once with the
arity information stored in the function entry object before doing the
normal application dispatch via the jump table.

`FUNCTION-ARITY-LIMIT` was lowered a tiny bit to make SBCL not barf on
the large block compilation.

Fixes https://github.com/coalton-lang/coalton/issues/631, https://github.com/coalton-lang/coalton/issues/632.

### DIFF
--- a/src/codegen/codegen-class.lisp
+++ b/src/codegen/codegen-class.lisp
@@ -11,7 +11,6 @@
    #:lisp-type)
   (:import-from
    #:coalton-impl/codegen/function-entry
-   #:apply-function-entry
    #:construct-function-entry)
   (:local-nicknames
    (#:tc #:coalton-impl/typechecker))
--- a/src/codegen/codegen-expression.lisp
+++ b/src/codegen/codegen-expression.lisp
@@ -41,7 +41,7 @@
     (let* ((arity (length (node-application-rands node)))
 
            (function-applicator
-             (gethash arity *function-application-functions*)))
+             (aref *function-application-functions* arity)))
 
       `(,function-applicator
         ,(codegen-expression (node-application-rator node) current-function env)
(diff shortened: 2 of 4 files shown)


eliaslfox (approved):
This is awesome

Issue #658, opened 2022-06-30
arithmetic error COMMON-LISP:FLOATING-POINT-INVALID-OPERATION


Ambrevar:
On 0a95e326e567c09c06be2dcc2073f1aa09736a6c I get this when I run the tests:

```log
[...]
  TEST-SLICE-CHUNKED-ITERATION............................................[ OK ]
  TEST-FLOAT
Unhandled COMMON-LISP:FLOATING-POINT-INVALID-OPERATION in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                                                    {1003378073}>:
  arithmetic error COMMON-LISP:FLOATING-POINT-INVALID-OPERATION signalled

Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {1003378073}>
0: ("bogus stack frame")
1: (BIG-FLOAT::|INSTANCE/INTO SINGLE-FLOAT BIG-FLOAT-COALTON-LIBRARY/CLASSES:INTO| #<SINGLE-FLOAT quiet NaN>)
2: (COALTON-LIBRARY/LIST::|INSTANCE/FUNCTOR LIST-COALTON-LIBRARY/CLASSES:MAP| #<COALTON-IMPL/CODEGEN/FUNCTION-ENTRY::FUNCTION-ENTRY :ARITY 1> (#<SINGLE-FLOAT quiet NaN> 0.0 1.0 3.1415927 2.7182817 -3.1415927 1.4142135 1.8137995 0.5 10.0 100.0))
3: (COALTON-LIBRARY/LIST::|INSTANCE/FUNCTOR LIST-COALTON-LIBRARY/CLASSES:MAP| #<COALTON-IMPL/CODEGEN/FUNCTION-ENTRY::FUNCTION-ENTRY :ARITY 1> (#.SB-EXT:SINGLE-FLOAT-POSITIVE-INFINITY #<SINGLE-FLOAT quiet NaN> 0.0 1.0 3.1415927 2.7182817 -3.1415927 1.4142135 1.8137995 0.5 10.0 100.0))
4: ((COMMON-LISP:LABELS TEST-FLOAT :IN TEST-FLOAT))
5: ((COMMON-LISP:LABELS FIASCO::RUN-TEST-BODY :IN FIASCO::RUN-TEST-BODY-IN-HANDLERS))
[...]

PR #661, opened 2022-07-01
Fix float tests

eliaslfox:
It seems like coercing NaN values to big floats is an error in sbcl
2.2.6.


eliaslfox:
See #658

aijony (approved):
My personal favorite technique for fixing a failing test case

PR #662, opened 2022-07-01
Fix bug in cell implementation

rvs314:
Currently, `cell:decrement` doesn't work: the problem is an issue with the semantics of `(- 1)`, which is equivalent to `(fn (x) (- 1 x))`, not `(fn (x) (- x 1))`, which is what was intended.


eliaslfox (approved):

stylewarning:
FYI: [`1-` exists](https://github.com/coalton-lang/coalton/blob/main/library/math/arith.lisp#L591)

Issue #539, opened 2022-05-22
Add list-analagous constructors for vectors and hash tables


ornamentist:
Since Coalton provides the `make-list` list constructor, could we also have analogous `make-vector` and `make-hash-table` constructors? 

I understand the `make-hash-table` one is probably more subtle than it sounds, needing a convention over order of keys and values. Are there other Lisps conventions we can draw on?

eliaslfox:
I've been thinking about this. I would like to implement a system like Haskell's [overloaded lists](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_lists.html). Then we could add instances like

```
(define-class (FromList :a :b)
  (fromList (List :a -> :b :a))
  (fromListN (UFix -> List :a -> :b :a)))

(define-instance (FromList :a (List :a)))
(define-instance (FromList :a (Vector :a)))
(define-instance (Hash :a => (FromList (Tuple :a :b) (HashMap :a :b))))
```

For syntax I want to use reader macros. Probably `#[1 2 3 4]` so that `#(1 "hello" None)` can be a three element tuple.

ornamentist:
That would be excellent.

ornamentist:
Are there any caveats or issues with creating reader macros for Coalton code?  I can see some other (library-specific) uses for them but won't go down that path if they're too problematic.

stylewarning:
> Are there any caveats or issues with creating reader macros for Coalton code?  I can see some other (library-specific) uses for them but won't go down that path if they're too problematic.

I don't think there's a problem with reader macros on their own, but there's a problem if Coalton users are forced to use them.

gefjon:
Personally, I'm very wary of using reader macros for things like this -- they have a lot of weird subtle interactions with other parts of the ecosystem, including pretty bad editor and debugger support. Some concerns I have:

- Does `#[a b c]` construct a new list/vector/etc at read-time or at runtime? If read-time, you get subtle bugs where types aren't defined yet, and you violate non-lispers' expectations about what happens when you write `(let ((a b)) #[a b])`. If runtime, you violate lispers' expectations about how reader macros work, and I don't see how you implement the construction efficiently (i.e. without consing a list as an intermediate sequence to pass to `make-array` or friends).
[...]

PR #663, opened 2022-07-04
`make-vector` and `make-hash-table` macros

gefjon:
re: #539 

including a little machinery for best-effort macroexpand-time detection of duplicate keys
in `make-hash-table`.

still necessary: docstrings and docs.

open for discussion: we *could* define that, if the same key appears twice in a
`make-hash-table` form, the latter value takes precedence. this seems stupid to me; i
believe statically-detectable duplicate keys in `make-hash-table` are always bugs, but i
mention it here because it's possible there's some use case i'm not thinking of. (macros
generating `make-hash-table` forms?)

note: `make-hash-table` does not do constant folding, it doesn't get a parsed ast, etc,
etc, so it can't detect all cases of duplicate keys, e.g. it misses `(let x =
a) (make-hash-table (x foo) (a bar))`. it's just a sanity check for the obvious cases.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -109,7 +109,8 @@
                       (funcall compile)))
   :depends-on (#:coalton/compiler
                #:coalton/hashtable-shim
-               #:trivial-garbage)
+               #:trivial-garbage
+               #:alexandria)
   :pathname "library/"
   :serial t
   :components ((:file "utils")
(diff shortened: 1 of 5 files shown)


eliaslfox:
Can you rename `make-hash-table` to `make` and `make-vector` to `make`. `hash-table:make-hash-table` seems a little redundant.

eliaslfox (approved):

PR #665, opened 2022-07-06
reduce `function-arity-limit` to control compile-time heap usage

gefjon:
an undocumented "not-user-facing" change in SBCL 2.2.6 increases heap usage for top-level
evaluation. that broke our CI builds, as the old large value of `function-arity-limit`
caused a heap overflow when compiling src/codegen/function-entry.lisp. this commit lowers
`function-arity-limit` to 32, which hopefully will fit in SBCL's default heap limit on the
CI runner.

### DIFF
--- a/src/codegen/function-entry.lisp
+++ b/src/codegen/function-entry.lisp
@@ -1,6 +1,6 @@
 (in-package #:coalton-impl/codegen/function-entry)
 
-(defconstant function-arity-limit 45)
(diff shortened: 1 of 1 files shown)


karlosz:
It may not be about top-level function calls actually. I looked at when that was introduced and it was 2.2.5. Maybe a bisection on sbcl would reveal the real issue.

eliaslfox (approved):

PR #666, opened 2022-07-06
Restructure COALTON-LIBRARY/MATH

eliaslfox:
* move INTO and TRYINTO instances to library/math/conversions.lisp
* move NUM and other instances to library/math/num.lisp
* move FRACTION defintions to library/math/fraction.lisp


library/math/bounded.lisp
```diff
@@ -0,0 +1,57 @@
...
+  (:export
+   #:Bounded #:minBound #:maxBound))
+
+#+coalton-release
+(cl:declaim #.coalton-impl:*coalton-optimize-library*)
+
+(in-package #:coalton-library/math/bounded)
+
+(coalton-toplevel
+  (define-class (Bounded :a)
```
stylewarning, line 20: How about IFIX and UFIX? You can use `(lisp Ifix () cl:most-positive-fixnum))` and the like.

stylewarning: Also document the class.

stylewarning (changes requested):

library/math/conversions.lisp
```diff
@@ -0,0 +1,566 @@
...
+
+      (when (< x 0)
+        (return (Err "value out of range")))
+
+      (Ok (lisp U8 (x) x))))
+
+  (define-instance (TryInto I32 I8)
+    (define (tryInto x)
+      (when (> x (into (the I8 maxBound)))
+        (return (Err "value out of range")))
```
stylewarning, line 279: any reason to not define a global constant

    (define out-of-range (Err ...))

library/math/conversions.lisp
```diff
@@ -0,0 +1,566 @@
...
+
+      (Ok (lisp U8 (x) x))))
+
+  (define-instance (TryInto U64 I8)
+    (define (tryInto x)
+      (when (> (into x) (the Integer (into (the I8 maxBound))))
+        (return (Err "value out of range")))
+
+      (Ok (lisp I8 (x) x))))
```
stylewarning, line 344: good ol `unsafe-reinterpret-cast`

library/math/conversions.lisp
```diff
@@ -0,0 +1,566 @@
...
+(cl:declaim #.coalton-impl:*coalton-optimize-library*)
+
+(in-package #:coalton-library/math/conversions)
+
+(coalton-toplevel
+  ;;
+  ;; Conversions from U8
+  ;;
+
+  (define-instance (TryInto U8 I8)
```
stylewarning, line 26: I think you'd have 4-5x fewer lines with

    (define-upconversion from-small to-big)

and

    (define-downconversion from-big to-small)

macros, or something.

stylewarning (changes requested):

gefjon:
Should we just have `unsafe-reinterpret-cast` in the library, to simplify a common CL interop pattern and clarify intent?

library/math/conversions.lisp
```diff
@@ -0,0 +1,566 @@
...
+  (define-instance (Into IFix Integer)
+    (define (into x)
+      (lisp Integer (x) x)))
+
+  (define-instance (Into UFix Integer)
+    (define (into x)
+      (lisp Integer (x) x))))
+
+(coalton-toplevel
+  (declare integer->double-float (Integer -> Double-Float))
```
aijony, line 554: This can result in an overflow for large integers that can't be represented by floats. Which is fine, but it should probably be documented, as well as perhaps a use-case to justify it. Otherwise, `fromInt` has achieves the same thing but with infs instead of errors. Although, `Into` instances of fixed-length integers would be completely safe.

eliaslfox: Yeah we should just use fron int.

library/math/conversions.lisp
eliaslfox: Yeah probably. It's a little more annoying because all of the signed -> unsigned conversions aren't really upcasting or downcasting. Also I'd rather read the expanded code.

eliaslfox:
> Should we just have unsafe-reinterpret-cast in the library, to simplify a common CL interop pattern and clarify intent?

I don't think so. If we define `unsafe-case` as `:a -> :b` and it's used like:
```
(f (unsafe-cast x))
```
Changing the argument type of `f` to something else will still compile. Using casts with `lisp` requires specifying the intended output type.

eliaslfox:
* moved definitions for {IU..I64} from early types to `library/fixed-sized-numbers.lisp`
* added `Bounded` instances for `IFix` and `UFix`
* removed overflow checks from `IFix` as discussed

stylewarning (approved):

PR #668, opened 2022-07-12
use `cl:length` in `vector:length` rather than explicit `cl:fill-pointer`

gefjon:
for purposes of lisp-coalton interop, it is useful to allow the coalton `Vector` type to
include Common Lisp `simple-vector`, i.e. to not require that coalton `Vector`s have fill
pointers and be adjustable.

the `repr :native` of `Vector` already allowed this; it is `(vector t)`, not `(and (vector
t) (not simple-array))`. but `coalton-library/vector:length`, which is called by many
vector operators including `map`, read a vector's length using the less-generic
`cl:fill-pointer` rather than the more-generic `cl:length`.

`cl:length` does what you want when applied to a vector: for vectors with fill pointers,
it reads the fill pointer, and for vectors without a fill pointer, it reads the
capacity. `cl:fill-pointer`, in the second case, is a type error.

this commit redefines `coalton-library/vector:length` to call `cl:length`, so i can
meaningfully operate on simple vectors.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -260,4 +260,5 @@
                (:file "iterator-tests")
                (:file "addressable-tests")
                (:file "call-coalton-from-lisp")
+               (:file "vector-tests")
                (:file "recursive-let-tests")))
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -65,7 +65,7 @@
   (define (length v)
     "Returns the length of V"
     (lisp Integer (v)
-      (cl:fill-pointer v)))
+      (cl:length v)))
 
   (declare capacity (Vector :a -> Integer))
   (define (capacity v)
--- a/tests/vector-tests.lisp
+++ b/tests/vector-tests.lisp
@@ -1,0 +1,5 @@
+(cl:in-package #:coalton-native-tests)
+
+;; test that it's possible to treat a lisp `simple-vector' as a coalton `Vector'
+(define-test vector-length-on-non-adjustable ()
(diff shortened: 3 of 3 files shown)


eliaslfox:
Seems reasonable. If we add fixed sized arrays, then this will probably get reverted. Vectors were never intended to wrap simple-arrays, the lisp type definition was an oversight on my part.

eliaslfox (approved):

gefjon:
> Seems reasonable. If we add fixed sized arrays, then this will probably get reverted. Vectors were never intended to wrap simple-arrays, the lisp type definition was an oversight on my part.

I've always thought it was silly (read: error-prone and wrong) that there isn't a useful way in CL to differentiate between simple arrays and the different kinds of non-simple arrays; if we can come up with names for them, I think it would be nice to have Coalton give different types to:

- Simple arrays.
- Displaced arrays.
- Arrays with fill pointers.
- Adjustable arrays.
- All combinations of the above.

If we can figure out an interface for that, then I'd like `Vector` to mean "non-displaced, adjustable array with fill-pointer."

For now, though, I just have a case where Lisp calls Coalton with a vector, and sometimes it has a fill pointer but sometimes it doesn't, and I need to `map` over it to convert it into a Coalton `Vector` containing something useful. I'd want to support that use-case too, but I think what we need is "I have a `cl:vector` with some indeterminate properties, and I want you to turn it into a Coalton `Vector`," not "I need a type that can encompass all possible `cl:vector`s without copying."

PR #671, opened 2022-07-20
Remove implicit `Unit` from `when` and `unless`

rvs314:
Currently, our implementations of `coalton:when` and `coalton:unless` do not execute their code in tail position. For example:

```lisp
(coalton
  (progn
    (let c = (cell:new 0))
    (let ((lp (fn (i)
                 (when (> i 0)
                   (cell:increment! c)
                   (lp (- i 1))))))
      (lp 100000)))) 
```

Gives a stack overflow, when other languages don't.

The reason is that we implicitly add a `Unit` form to the end each `when` and `unless` clause, which
breaks the tail positioning. Removing these means that tail recursion works, but it breaks some imperative code 
which relies on that type-checking to pass. I've fixed the relevant places in the stdlib which rely on it.

### DIFF
--- a/library/slice.lisp
+++ b/library/slice.lisp
@@ -158,7 +158,8 @@
             (foreach2
              (fn (e1 e2)
                (unless (== e1 e2)
-                 (cell:write! out False)))
+                 (cell:write! out False)
+                 Unit))
              s1 s2)
             (cell:read out)))))
 
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -247,7 +247,8 @@
             (foreach2
              (fn (e1 e2)
                (unless (== e1 e2)
-                 (cell:write! out False)))
+                 (cell:write! out False)
+                 Unit))
              v1 v2)
             (cell:read out)))))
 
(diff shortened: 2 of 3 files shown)


stylewarning:
That specific example could also be fixed by changing the WHEN to COND, I assume.

rvs314:
> That specific example could also be fixed by changing the WHEN to COND, I assume.

Yes - only `when` and `unless` add the implicit `Unit`

eliaslfox:
I don't like this. Adding `Unit` will add verbosity to standard uses of when and unless. The above code can easily be written with an if expression.

gefjon:
I'd like to take this time to remind everyone that I still think `when` and `unless` should return an `Optional` by wrapping the taken branch in `Some` and returning `None` when not taken.

rvs314:
> I don't like this. Adding `Unit` will add verbosity to standard uses of when and unless. The above code can easily be written with an if expression.

For reference - I needed to change two lines in the entire standard library for this to work - every other usage of `when` and `unless` already has this property.

rvs314:
Also, this code is in tail position in:
- Common Lisp
- Scheme
- OCaml

A silent error here really violates principle of least astonishment.

eliaslfox:
I think @rvs314 is right that clear tail call semantics are more important than the line noise.

eliaslfox (approved):

PR #672, opened 2022-07-22
added json->string function to coalton-json

yaktender:


examples/coalton-json/src/representation.lisp
```diff
@@ -66,4 +77,37 @@
...
+             (cl:format cl:nil "{~{~a~^, ~}}" pairs)))))))
+
+  (declare json->string (Json -> String))
+  (define (json->string jsn)
+    (match jsn
+      ((Json-Null) "nil")
+      ((Json-Boolean b) (if b "true" "false"))
+      ((Json-Number n) 
+       (lisp String (n)
+         (cl:format cl:nil "~$" n)))
```
gefjon, line 106: I'm not sure `~$` is the right float formatter here. In fact, I'm reasonably sure it is not; `~$` (with no additional arguments) prints exactly two digits after the decimal. This means that `(json->string (Json-Number 0.001))` will return `"0.00"`. I believe you want `~f`.

examples/coalton-json/tests/tests.lisp
```diff
@@ -33,7 +33,16 @@
...
+             (coalton-json:parse-json "[1, 1.0, true, false, [], \"x\"]"))))
   (is-type 'coalton-json::json/json-object
            (coalton-json:parse-json "{}"))
   (is-type 'coalton-json::json/json-object
-           (coalton-json:parse-json "{\"x\":0, \"y\":{\"z\":\"z\"}}")))
+           (coalton-json:parse-json "{\"x\":0, \"y\":{\"z\":\"z\"}}"))
+  (is-type 'coalton-json::json/json-object
+           (coalton-json::parse-json 
+            (coalton-json::json->string
+             (coalton-json:parse-json "{\"x\":0, \"y\":{\"z\":\"z\"}}")))))
```
gefjon, line 47: I'd like to see some tests here that compare the output off `json->string` to an expected string, for representatives of at least integers (small positive, large positive, small negative, large negative, zero, perhaps), floats (positive and negative zero, very large positive and negative numbers, very small positive and negative numbers, and numbers with non-zero digits to both the left and right of the decimal point, perhaps), booleans (both of em), and strings (idk, whatever). It would also be nice to demonstrate, for types with `Eq` instances, that `(compose parse-json json->string)` is like `id`.

gefjon (changes requested):

examples/coalton-json/src/representation.lisp
yaktender: updated

examples/coalton-json/tests/tests.lisp
yaktender: added more tests but unfortunately can't do `(compose parse-json json->string)` is like `id` because of the formatting of floats like 1 to 1.0.

gefjon (approved):

PR #667, opened 2022-07-07
Remove the correct NaN this time

eliaslfox:
See #658

PR #675, opened 2022-07-23
Portable `Big-Float`

aijony:
Here is a first attempt at making a portable `Big-Float` implementation.  See the readme for usage details.

I went with a binary splitting algorithm for evaluating power series so I could implement all of the transcendental functions using one algorithm. This can leave a lot to be desired in terms of performance, but hopefully this lays the ground work so we can start implementing faster methods.

For regular arithmetic, I think it actually should be fairly performant.

I also wanted to avoid lookup tables for now just to keep things feeling a bit more lean. Although, I think having pre-computed values for ln2 or pi would make sense at some point in the future.

Couple of slightly tangential notes:
- #666 changed the type of the sign function, making it difficult to extract the sign of an element as an integer. I changed the sign function to a more flexible type signature.
- Moved fibonnaci benchmarks to their own file
- Implements #667

### DIFF
--- a/benchmarks/big-float.lisp
+++ b/benchmarks/big-float.lisp
@@ -1,0 +1,101 @@
+;;;; big-float.lisp
+;;;;
+;;;; Benchmarks for arbitrary precision floats
+
+(cl:in-package #:coalton-benchmarks)
+
+(cl:defvar *big-float-bench-precision*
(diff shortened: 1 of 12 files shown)


eliaslfox (approved):
This looks good, but I'll wait for @stylewarning's review

stylewarning (approved):

PR #676, opened 2022-07-25
test for substring appearance

gefjon:
Define two functions, `string:substring-index` and `string:substring?`, for searching a string for the appearance of a substring.

`substring-index` is `String -> String -> (Optional UFix)`, returning the index of the start of the substring.

`substring?` is `String -> String -> Boolean`, returning true if the substring appears in the string.

Both of these functions take the "needle" substring as their first argument, and the "haystack" superstring as their second argument. I believe this to be more useful for currying; `(substring? "foo")` is a function which, given a string, tests if `"foo"` appears in that string, and `(substring-index "foo")` is a function which, given a string, returns the first index of `"foo"` in that string.

### DIFF
--- a/library/string.lisp
+++ b/library/string.lisp
@@ -12,7 +12,9 @@
    #:strip-suffix
    #:parse-int
    #:ref
-   #:ref-unchecked))
+   #:ref-unchecked
+   #:substring-index
+   #:substring?))
 
 #+coalton-release
 (cl:declaim #.coalton-impl:*coalton-optimize-library*)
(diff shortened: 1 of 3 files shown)


library/string.lisp
```diff
@@ -90,6 +92,22 @@ does not have that suffix."
         (Some (ref-unchecked str idx))
         None))
 
+  (declare substring-index (String -> String -> (Optional UFix)))
+  (define (substring-index small big)
+    "If SMALL appears in BIG, return the index into BIG of the start of SMALL."
```
eliaslfox, line 97: Use first and second argument instead of SMALL and BIG.

eliaslfox (approved):

PR #677, opened 2022-07-25
simplify integer conversions and define fixnum conversions

gefjon:
Replace hand-written `define-instance Into` and `define-instance TryInto` forms with forms
generated by a macro, `define-integer-conversions`.

Define polymorphic conversion functions `unsafe-cast` and `cast-if-inbounds` to encode the
conversion logic, so that the instances generated by `define-integer-conversions` are all
either `define-instance (Into _ _) (define into unsafe-cast)` or `define-instance (TryInto
_ _) (define tryInto cast-if-inbounds)`. `cast-if-inbounds` reads `minBound` and
`maxBound` from the `Bounded` instance for the target type, `unsafe-cast`s them to
`Integer`, and compares them against the value of the source type.

Define conversions to and from IFix and UFix for fixed-width integers, which previously
were only present on `Integer`. Caveat: because `define-integer-conversions` uses
`cl:subtypep` to decide whether two types should have an `Into` or `TryInto` instance,
this will generate different sets of instances on 32-bit platforms as compared to
64-bit ("normal") platforms. Personally, I am unconcerned by this, as I don't believe
32-bit platforms exist in the wild.


eliaslfox:
I'm uncomfortable with defining different conversions on different platforms. If we wan't to guarantee that a U32 can fit in a fixnum, then that should be checked at compile time. And compiling Coalton on a 32 bit platform should error.

Also I think changing this file to be macro generated makes the code much harder to read for a questionable increase in maintainability.

gefjon:
> I'm uncomfortable with defining different conversions on different platforms. If we wan't to guarantee that a U32 can fit in a fixnum, then that should be checked at compile time. And compiling Coalton on a 32 bit platform should error.

Hmm, yeah, this seems like a reasonable concern. I personally would prefer to assert that fixnum is always larger than 32-bit and smaller than 64-bit, but I'm open to other opinions.

> Also I think changing this file to be macro generated makes the code much harder to read for a questionable increase in maintainability.

I'm prepared to chalk this one up to different backgrounds, but personally I find the macro version to be much more readable. It separates concerns so that all of the actual conversion is handled by either `unsafe-cast` or `cast-if-inbounds`, both of which (I think) have easy-to-understand names and clear behavior (although the implementation of `cast-if-inbounds` is sorta messy because we don't have scoped type variables), and it encodes the logic of "what types can be converted" in a concise and precise way. With the old version, I had some trouble because I had to manually search through a large file of `define-instance` forms to find the instances I cared about (or where they would've been if they were defined), and to read each individual `define-instance` to determine how it behaved.

I also think the increase in maintainability is more than "questionable." While authoring this change, I first defined `unsafe-cast` and several different helpers like `cast-if-positive`, `cast-if-less`, etc., and rewriting each individual `define-instance` form to call the appropriate function. After around the third incremental improvement to the logic functions, I found editing each instance to be prohibitively tedious, so I wrote the macro. After I had the macro in place, I found iterating through designs much, much easier, and eventually settled on the `cast-if-inbounds` you see today. If it turns out that `cast-if-inbounds` is incorrect or suboptimal (e.g. we get scoped-type-variables and want to remove `unify`), the macro will save having to update a bunch of instances by hand again. I believe updating instances by hand to be both tedious and error-prone, so I'd prefer to avoid it.

That said, the `define-integer-conversions` macro as I wrote it is underdocumented, uncommented and dense. It could certainly benefit from having parts of its logic unloaded into separate top-level definitions. At the very least, I should write `define-infallible-into-instance` and `define-fallible-tryinto-instance` (or whatever names) as separate top-level macros, and `repr-subtypep?` should be a top-level `eval-when :compile-toplevel` function with a better name.

eliaslfox:
> Hmm, yeah, this seems like a reasonable concern. I personally would prefer to assert that fixnum is always larger than 32-bit and smaller than 64-bit, but I'm open to other opinions.

Yeah I think that's the right way to do it.

> With the old version, I had some trouble because I had to manually search through a large file of define-instance forms to find the instances I cared about (or where they would've been if they were defined), and to read each individual define-instance to determine how it behaved.

But it was at least possible to see which instances are defined.

gefjon:
> But it was at least possible to see which instances are defined.

Thus `sly-macroexpand-1`.

eliaslfox:
Which doesn't work on Github. I've read parts of other languages standard libraries, and most of it has been on the web from doc links. I'd like to make reading the stdlib as easy as possible, including for non contributors, and people not currently in emacs.

gefjon:
I am sympathetic to that desire, but I think the solution to it is getting better HTML doc-gen, not treating our code as its own documentation and thus imposing artificial verbosity.

Also, again, I personally think the macro version is easier to reason about; reading a handful of lines of Lisp tells me all of the instances that are defined, as opposed to searching through multiple screen-lengths of near-identical code to tell whether the instance I want to know about is there or not.

Until we get rigorous HTML doc-gen, would a docstring or comment in the file describing in English the rules about what instances are defined satisfy you? Something like:

> `Into :small :large` is defined for any pair of integer types where :LARGE can represent every element of :SMALL, e.g. `Into U8 I64` is defined.
>
> `TryInto :from :to` is defined for any other pair of integer types, where there are some elements of :FROM which cannot be represented in :TO. These fall into a few categories:
> - `TryInto :signed :unsigned` is implemented rather than `Into :signed :unsigned` where :SIGNED is one of I8, I16, I32, I64, IFix; and :UNSIGNED is one of U8, U16, U32, U64; because negative numbers in the source type cannot be represented in the destination type.
> - `TryInto :larger :smaller` is implemented rather than `Into :smaller :larger` where :LARGER is wider than :SMALLER (e.g. U64 is wider than U32, and I64 is wider than I32) because larger (both more-positive and more-negative) values of the source type cannot be represented in the destination type.

It would also be pretty easy to have the macro generate a human-readable listing of all the instances it generated, but that doesn't solve the problem of formatting and displaying generated documentation to a user.

eliaslfox (approved):

PR #678, opened 2022-07-25
use `UFix` as sequence indices rather than `Integer`

gefjon:
All array types (vectors, strings, slices, etc.) require that their indices be positive
fixnums, never bignums (see the spec on `ARRAY-TOTAL-SIZE-LIMIT`,
http://www.lispworks.com/documentation/HyperSpec/Body/v_ar_tot.htm).

For consistency's sake, it seems reasonable to impose the same restriction on list
indices; I find it unlikely that anyone will be able to cons a non-circular list with
length > most-positive-fixnum, as on both 32-bit and 64-bit machines with 1-bit fixnum
tags, that many cons cells would fill the entire address space.

Lazy iterators are also unlikely to ever have non-fixnum indices, at least within a human
user's lifetime. (This might not be true on a 32-bit machine, but I maintain that 32-bit
machines aren't real.)

In light of that, this commit restricts the library sequences' length- and index-related
operators to have type `UFix` rather than `Integer`.

It also generalizes the type of `list:range` to support any type that is `Num` and `Ord`,
rather than just `Integer`; this came up when editing thih-coalton to use `UFix` to
represent `TGen` generated types (which are indexes into some list of types).

### DIFF
--- a/examples/thih/src/thih.lisp
+++ b/examples/thih/src/thih.lisp
@@ -54,7 +54,7 @@
     (TVar Tyvar)
     (TCon Tycon)
     (TAp Type Type)
-    (TGen Integer))
+    (TGen UFix))
 
   (define-instance (Eq Type)
       (define (== t1 t2)
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -138,14 +138,14 @@ Behavior is undefined if the iterator is advanced after a destructive modificati
     "Yield successive elements of VEC.
 Behavior is undefined if the iterator is advanced after a destructive modification of VEC."
     (map ((flip vector:index-unsafe) vec)
-         (up-to (fromInt (vector:length vec)))))
+         (up-to (vector:length vec))))
 
   (declare string-chars (String -> Iterator Char))
   (define (string-chars str)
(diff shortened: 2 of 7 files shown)


examples/thih/src/thih.lisp
```diff
@@ -54,7 +54,7 @@
     (TVar Tyvar)
     (TCon Tycon)
     (TAp Type Type)
-    (TGen Integer))
+    (TGen UFix))
```
eliaslfox, line 57: Leave this as integer. It matches the paper.

eliaslfox (approved):

stylewarning:
Seems reasonable.

examples/thih/src/thih.lisp
gefjon: I'm not sure I agree that a closer correspondence to the paper is more valuable here than the added conversions I'll have to add to `quantify`. I think using `UFix` is both more correct and cleaner here, and the only reason the paper uses `Integer` is because Haskell didn't have unsigned integer types until quite recently.

eliaslfox: I guess UFix is fine then

PR #681, opened 2022-07-29
Add kind signatures and poly-kinds

aijony:
Introduces a polykinds, that is a kind `:a` that satisfies kinds `*`, `(* -> *)`, etc.
It doesn't do anything too fancy, it's just an escape hatch from the least general kinds to more general kinds.

It is off by default, so in order to facilitate its usage, I have added declarations for kind signatures:
```
(declare-type Unit *)
(define-type Unit Unit)

(declare-type Proxy (:a -> *))
(define-type (Proxy :a) Proxy)
```


eliaslfox:
I think I found an issue:

```
(coalton-toplevel
  (declare-type Proxy2 (:a -> :a -> *))
  (define-type (Proxy2 :a :b) Proxy2)

  (declare p (Proxy2 Integer Optional))
  (define p Proxy2))
```

`Integer` and `Optional` have different kinds and shouldn't unify against `:a -> :a -> *`.

gefjon:
> Integer and Optional have different kinds and shouldn't unify against :a -> :a -> *.

I think the syntax here is unclear. It seems like the version @aijony is proposing has each `:a` in a kind signature referring to a new, unique kind variable, so `:a -> :a -> *` parses as something like `(KindVar 0) -> (KindVar 1) -> *`.

It seems like the version @eliaslfox expects (I also would expect this version) would treat any keyword in a kind signature as referring to a kind variable, unique by `eq`, so `:a -> :a -> *` would parse as something like `(KindVar 0) -> (KindVar 0) -> *`, and `:a -> :b -> *` would parse as `(KindVar 0) -> (KindVar 1) -> *`.

eliaslfox:
It looks like Coalton thinks the two `:a`s are the same kind variable.

````
COALTON-USER> (kind-of 'proxy2)
∀ :A. :A → (:A → *)
```

aijony:
> I think I found an issue:
> 
> ```
> (coalton-toplevel
>   (declare-type Proxy2 (:a -> :a -> *))
>   (define-type (Proxy2 :a :b) Proxy2)
> 
>   (declare p (Proxy2 Integer Optional))
>   (define p Proxy2))
> ```
> 
> `Integer` and `Optional` have different kinds and shouldn't unify against `:a -> :a -> *`.

Great find. I know do substitution of `kpoly`s as well. That fixed the issue and I added a test case.

eliaslfox (approved):
This is really great. 

I think eventually we'll want some form of `declare-class`, but I don't it should block this pr.

eliaslfox:
I think the last change broke something else:

```
(coalton-toplevel
  (declare-type Proxy (:a -> *))
  (define-type (Proxy :a) Proxy)
                
  (declare p1 (Proxy Integer))
  (define p1 Proxy))
```

The above fails with:

```
Failed to unify types PROXY and PROXY
In unification of types (PROXY INTEGER) (type of type of binding) and (PROXY #T595) (type of type of variable)
In definition of P1
In COALTON-TOPLEVEL

```

eliaslfox:
Reverted this because it seems unusable.

PR #684, opened 2022-07-29
Hashtables Improvements

aijony:
This provides:
- Portable hash-tables using [genhash](https://github.com/pnathan/genhash)
- Load-forms for SBCL hash-tables
- Remove SBCL version-specific code (I still need to test this actually works on previous versions of SBCL)

Depends on: #682


eliaslfox:
Closing as this would have required #681.

PR #690, opened 2022-07-31
add doc to typed-node-list-p

Jobhdez:


src/typechecker/typed-node.lisp
```diff
@@ -9,6 +9,7 @@
   (unparsed (required 'unparsed) :type t         :read-only t))
 
 (defun typed-node-list-p (x)
+  "Returns True if every element in the list X is of TYPED-NODE."
```
stylewarning, line 12: "Is every element of the list X a TYPED-NODE?"

stylewarning (changes requested):

eliaslfox:
@stylewarning or @gefjon feel free to weight in. But personally I don't see the point in adding docstrings to these functions. Especially because they only exist for the subsequent deftype.

stylewarning:
> @stylewarning or @gefjon feel free to weight in. But personally I don't see the point in adding docstrings to these functions. Especially because they only exist for the subsequent deftype.

In this case, for the reason you state, I'm ambivalent. You're right that these kinds of functions aren't intended to really be used directly.

eliaslfox:
Closing this for reasons described above. If the type checker used per file packages, then this function wouldn't be exported.

PR #693, opened 2022-08-01
Fix typechecking of toplevel restricted bindings

### LINKED ISSUE #654: Invalid codegen on top level assignment of functions to variables
The following fails with an arity error:
```
(coalton-toplevel
  (define add +))

(coalton (add 1 2))
```

eliaslfox:
Fixes #654


stylewarning (approved):
LGTM

PR #694, opened 2022-08-01
Enable direct application in coalton macro

### LINKED ISSUE #679: Data-Letrec doesn't work in Coalton form
When I use data letrec in a toplevel definition, it works as expected:

```lisp
(coalton-toplevel
  (define foo
    (let ((a (cons 0 a)))
      a)))

(coalton foo) ; => #1=(0 . #1#)
```

However, a seemingly equivalent operation fails:

```lisp
(coalton
  (let ((a (cons 0 a)))
    a))

; caught COMMON-LISP:WARNING:
;   undefined variable: GRAMMAR-COMPRESSION::A-767
```

@gefjon @eliaslfox - is there a reason for this? Let me know if I'm just using it wrong or if this is an actual bug.

eliaslfox:
This also fixes #679, because recursive construction required that the
constructor was a direct application.

### DIFF
--- a/src/codegen/codegen-expression.lisp
+++ b/src/codegen/codegen-expression.lisp
@@ -224,8 +224,10 @@
             ,(codegen-expression (node-bind-body expr) current-function env)))))))
 
 (defun find-constructor (initform env)
-  (and (node-direct-application-p initform)
-       (tc:lookup-constructor env (node-direct-application-rator initform) :no-error t)))
+  (if (or (node-application-p initform) (node-direct-application-p initform))
+      (and (node-rator-name initform)
+           (tc:lookup-constructor env (node-rator-name initform) :no-error t))
+      nil))
 
 (defun data-letrec-able-p (initform env)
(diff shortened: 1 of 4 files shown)


src/codegen/codegen-expression.lisp
```diff
@@ -224,8 +224,9 @@
             ,(codegen-expression (node-bind-body expr) current-function env)))))))
 
 (defun find-constructor (initform env)
-  (and (node-direct-application-p initform)
-       (tc:lookup-constructor env (node-direct-application-rator initform) :no-error t)))
+  (if (or (node-application-p initform) (node-direct-application-p initform))
+      (and (node-rator-name initform)
+           (tc:lookup-constructor env (node-rator-name initform) :no-error t))))
```
stylewarning, line 229: NIL as else branch? This code looks like it might be a mistake, but probably isn't.

stylewarning (approved):
a nit but nbd

PR #695, opened 2022-08-02
Document specializations

eliaslfox:
Also add `print-specializations`.


stylewarning (approved):
Always love doc improvements!

PR #702, opened 2022-08-10
define `eq-hash`, a thin wrapper around `cl:sxhash`.

gefjon:


library/addressable.lisp
```diff
@@ -26,3 +26,13 @@
 
 (define-instance-addressable (List :elt))
 (define-instance-addressable String)
+
+(coalton-toplevel
+  (declare eq-hash (Addressable :obj => :obj -> UFix))
+  (define (eq-hash obj)
+    "Compute a hash for OBJ based on its `eq?' identity.
```
stylewarning, line 33: this doesn't really make a difference but SXHASH does it based off of EQ so this will collide more often than not in ordinary circumstances

stylewarning (approved):

eliaslfox (approved):
I think the docstring should mention that this just calls `cl:sxhash`, so the results are implementation dependent.

stylewarning:
LGTM. Can you squash?

PR #705, opened 2022-08-12
Make map tail-recursive

rvs314:
Currently, for some reason, `list:map` is not a tail recursive operation, though it definitely seems like it should be. The proposed change does allocate more than the original does, but there's no way to prevent this unless we were to add something like `nreverse`.

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -634,9 +634,7 @@ This function is equivalent to all size-N elements of `(COMBS L)`."
 
   (define-instance (Functor List)
     (define (map f l)
-      (match l
-        ((Cons x xs) (Cons (f x) (map f xs)))
-        ((Nil) Nil))))
+      (reverse (fold (fn (a x) (Cons (f x) a)) Nil l))))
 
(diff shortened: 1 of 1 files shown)


gefjon:
This seems fine, but personally I would have no problem with defining a `reverse!` (or something) that calls `cl:nreverse`, and using that to implement `map`.

rvs314:
The problem with writing a `reverse!` would be that we would expose the mutability of cons cells, which seems like a can of worms that doesn't need to be opened yet. I'd be fine using it locally as an optimization, but exposing it as part of the stdlib seems problematic - not to say mutability is necessarily an issue, it just seems jarring to only expose it in one high-level function.

gefjon:
Sure, defining it internally and not exporting it (and adding a comment to the definition that it is intentionally not exported) seems totally reasonable.

rvs314:
Looking through the `list` package, there's lots of code here that should be written tail-recursively that isn't. I'm going to close this and open a new PR fixing all of them.

PR #706, opened 2022-08-13
Rewrite list module to be tail recursive

rvs314:
I rewrote some parts of the list module to use tail recursion where reasonable. I also added some simple tests for the functions.

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -157,7 +157,7 @@
   ;; This is only for internal usage
   (declare %reverse! (List :a -> List :a))
   (define (%reverse! xs)
-    "A mutating reverse operation"
+    "A mutating reverse operation. After (%reverse! LST), LST must not be referenced; the original list, and all its sublists, should be treated as consumed by this operation. Callers should use the return value and only the return value."
     (lisp (List :a) (xs)
       (cl:nreverse xs)))
 
@@ -264,15 +264,15 @@
           (%reverse! (inner start end Nil))
           (inner end start Nil))))
 
-  (define (%append list result)
+  (define (append-rev list result)
     (match list
       ((Nil) result)
-      ((Cons x xs) (%append xs (Cons x result)))))
+      ((Cons x xs) (append-rev xs (Cons x result)))))
 
   (declare append (List :a -> List :a -> List :a))
   (define (append xs ys)
     "Appends two lists together and returns a new list."
-    (%reverse! (%append ys (%append xs Nil))))
+    (%reverse! (append-rev ys (append-rev xs Nil))))
 
   (declare concat (List (List :a) -> List :a))
   (define (concat xs)
@@ -282,7 +282,7 @@
   (declare concatMap ((:a -> (List :b)) -> List :a -> List :b))
   (define (concatMap f xs)
     "Apply F to each element in XS and concatenate the results."
-    (%reverse! (fold (fn (a b) (%append (f b) a)) Nil xs)))
+    (%reverse! (fold (fn (a b) (append-rev (f b) a)) Nil xs)))
 
   (declare member (Eq :a => (:a -> (List :a) -> Boolean)))
   (define (member e xs)
@@ -306,7 +306,7 @@
                          (member x ys))
                      (rec xs acc)
                      (rec xs (Cons x acc))))))))
-      (%reverse! (%remove-duplicates ys (rec xs Nil)))))
+      (%reverse! (remove-duplicates-rev ys (rec xs Nil)))))
 
   (declare intersection (Eq :a => ((List :a) -> (List :a) -> (List :a))))
   (define (intersection xs ys)
@@ -318,7 +318,7 @@
                           (inner xs ys (Cons x acc))
                           (inner xs ys acc)))
                      ((Nil) acc)))))
-      (%reverse! (inner (%remove-duplicates xs Nil) (%remove-duplicates ys Nil) Nil))))
+      (%reverse! (inner (remove-duplicates-rev xs Nil) (remove-duplicates-rev ys Nil) Nil))))
 
   (declare lookup (Eq :a => (:a -> (List (Tuple :a :b)) -> (Optional :b))))
   (define (lookup e xs)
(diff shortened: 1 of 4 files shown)


library/list.lisp
```diff
@@ -350,44 +368,46 @@
...
-      ((Tuple (Cons x xs)
-              (Cons y ys))
-       (Cons (f x y) (zipWith f xs ys)))
-      (_ Nil)))
+    (let ((rec
+            (fn (xs ys acc)
+               (match (Tuple xs ys)
+                 ((Tuple (Cons x xs) (Cons y ys))
+                  (rec xs ys (Cons (f x y) acc)))
+                 (_ (%reverse! acc))))))
```
stylewarning, line 376: lift out reverse?

library/list.lisp
```diff
@@ -350,44 +368,46 @@
...
-               (Cons y ys)
-               (Cons z zs))
-       (Cons (f x y z) (zipWith3 f xs ys zs)))
-      (_ Nil)))
+    (let ((rec
+            (fn (xs ys zs acc)
+               (match (Tuple3 xs ys zs)
+                 ((Tuple3 (Cons x xs) (Cons y ys) (Cons z zs))
+                  (rec xs ys zs (Cons (f x y z) acc)))
+                 (_ (%reverse! acc))))))
```
stylewarning, line 387: lift out reverse?

library/list.lisp
```diff
@@ -412,15 +432,15 @@
...
-         ((GT)
-          (Cons y (insertBy cmp x ys_)))
-         (_
-          (Cons x ys))))))
+    (let ((rec 
+            (fn (ys acc)
+              (match ys
+                ((Nil) (Cons x acc))
+                ((Cons y yss)
+                 (match (cmp x y)
```
stylewarning, line 440: why not just IF here?

stylewarning (approved):
lgtm. any thoughts sbout lifting reverse! calls? it doesn't really matter either way i suppose.

library/list.lisp
```diff
@@ -597,9 +615,9 @@ This function is equivalent to all size-N elements of `(COMBS L)`."
           (True (match l
                   ((Nil) Nil)
                   ((Cons x xs) (append
-                              (map (Cons x) (combsOf (- n 1) xs)) ; combs with X
-                              (combsOf n xs)                      ; and without x
-                              ))))))
+                                (map (Cons x) (combsOf (- n 1) xs)) ; combs with X
+                                (combsOf n xs)))))))                      ; and without x
```
gefjon, line 619: i have no strong feelings about putting the parens on the line or after it, but could you keep the comments lined up?

library/list.lisp
gefjon: is `(if (== GT (cmp x y)) ... ...)` really better than a two-armed `match`? i'm ambivalent.

actually, this makes me think we should have a `(matches? PATTERN EXPR)` macro that returns `True` if the result of `EXPR` matches `PATTERN`, and `False` if it doesn't.

library/list.lisp
```diff
@@ -151,6 +154,13 @@
     ;; like (fold (flip Cons) Nil xs)
     (%reverse xs Nil))
 
+  ;; This is only for internal usage
+  (declare %reverse! (List :a -> List :a))
+  (define (%reverse! xs)
+    "A mutating reverse operation"
```
gefjon, line 160: even though this is only for internal usage, i'd like to add to the docstring the spec's caveat that `nreverse` is not required to (and won't) mutate the list in such a way that `(eq list (nreverse list))`, so you have to use the result and not re-use the original.

i think i'd add another line:

```
After (%reverse! LST), LST must not be referenced; the original list, and all its sublists, should be treated as consumed by this operation. Callers should use the return value and only the return value.
```

library/list.lisp
```diff
@@ -283,31 +292,33 @@
...
   (define (union xs ys)
     "Returns a new list with the elements from both XS and YS and without duplicates."
-    (match xs
-      ((Cons x xs)
-       (if (or (member x ys)
-               (member x xs))
-           (union xs ys)
-           (Cons x (union xs ys))))
-      ((Nil) (remove-duplicates ys))))
+    (let ((rec
```
gefjon, line 300: anyone know why our `union` doesn't just call `cl:union`?

library/list.lisp
```diff
@@ -321,26 +332,33 @@
               (lookup e xs)))))
       ((Nil) None)))
 
-  (declare remove-duplicates (Eq :a => ((List :a) -> (List :a))))
-  (define (remove-duplicates xs)
-    "Returns a new list without duplicate elements."
+  (declare %remove-duplicates (Eq :a => ((List :a) -> (List :a) -> (List :a))))
```
gefjon, line 335: i'd prefer a more descriptive name that encodes how this function is different from the exported `remove-duplicates`, like `remove-duplicates-reversing` or something.

library/list.lisp
```diff
@@ -321,26 +332,33 @@
...
+        
+  (declare remove-duplicates (Eq :a => ((List :a) -> (List :a))))
+  (define (remove-duplicates xs)
+    "Returns a new list without duplicate elements."
+    (%reverse! (%remove-duplicates xs Nil)))
 
-  (declare delete (Eq :a => (:a -> (List :a) -> (List :a))))
-  (define (delete x ys)
-    "Return a new list with the first element equal to X removed."
+  (declare %delete (Eq :a => (:a -> (List :a) -> (List :a) -> (List :a))))
```
gefjon, line 349: same comment re: names as with `%remove-duplicates`.

gefjon (changes requested):
loving the tests, and this all is a huge improvement!

library/list.lisp
stylewarning: Am I missing something? Why not `(if (> x y) ...)`?

library/list.lisp
stylewarning: is it a pain to call the type class method `==` in the `:test`?

library/list.lisp
gefjon: oh right. duh.

library/list.lisp
gefjon: a little bit; you need to use a helper function that returns a tuple of `(:a, (:a -> :a -> Boolean))` to build a closure over the instance. `%get-hash-test-funcs` does that in hashtable.lisp, and it's a little nasty. i think scoped-type-variables would eliminate the need for that.

rvs314: Another caveat with this is that SBCL's `union` uses a hash-table under the hood, and we can't guarantee that `Hash` will ever be implemented.

library/list.lisp
rvs314: The reason is that we don't actually have an `Ord` instance on `:a`, we only have a function which returns an `Ord` value. I don't really care whether we use an `if` with `==` or `match` in the body of the thing.

library/list.lisp
rvs314: I've added the `-rev` prefix to the functions that implicitly reverse their first argument. It may not be a bad idea to export something like SRFI 1's `append-reverse` for more efficient list functions.

library/list.lisp
gefjon: looking at the source for SBCL's `union`, it looks like they only use a hash-table when the test is one of the standard hash-table tests (`eq`, `eql`, `equal`, `equalp`), and otherwise do a relatively naïve `remove-duplicates` that's essentially the same as ours. our `==` will never be one of the standard hash-table tests, so at this point i don't see any reason to call `cl:union`.

that said, i think it would be reasonable to require `Hash :a` and remove duplicates by hash table; that's what i wrote for the iterator version.

eh.

gefjon (approved):

tests/package.lisp
```diff
@@ -24,14 +24,16 @@
   (:documentation "Tests for the COALTON system, written in Coalton.")
   (:use #:coalton-testing)
   (:local-nicknames
+   (#:fun  #:coalton-library/functions)
```
eliaslfox, line 27: `coalton-library/functions` should be used already

gefjon: `coalton-testing` use-reexports `coalton-prelude` which use-reexports `coalton-library/functions`.

eliaslfox (approved):

library/list.lisp
stylewarning: ah, that makes sense. MATCH is fine then

PR #713, opened 2022-08-22
Refactor define-type parsing

eliaslfox:
The new dataflow is easier to follow.

### DIFF
--- a/src/typechecker/parse-type-definition.lisp
+++ b/src/typechecker/parse-type-definition.lisp
@@ -4,7 +4,6 @@
 ;;; Parsing type defintions
 ;;;
 
-
 (defstruct type-definition
   (name              (required 'name)              :type symbol                 :read-only t)
   (type              (required 'type)              :type ty                     :read-only t)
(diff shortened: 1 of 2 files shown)


gefjon (approved):
`parse-type-definitions` is still long as hell; any chance I can convince you to unload some parts of its logic into separate named functions?

aijony:
> `parse-type-definitions` is still long as hell; any chance I can convince you to unload some parts of its logic into separate named functions?

I think a major point of this refactor was to do the exact opposite of this (as far as I can tell). Both methods have their merits, but in this case I think the sentimentality is more important than modularity.

src/typechecker/parse-type-definition.lisp
```diff
@@ -289,128 +290,204 @@ Returns TYPE-DEFINITIONS"
...
+                            :runtime-type repr-arg
+                            :explicit-repr (list repr repr-arg)
+                            :enum-repr nil
+                            :newtype nil
+                            :constructors parsed-ctors
+                            :constructor-types ctor-types
+                            :docstring docstring))
+
+                          ((and newtype (eql repr :transparent))
+                           (let (;; The runtime type of a newtype is the runtime type of it's only constructor's only argument
```
aijony, line 370: Probably should be "its", but eh.

aijony (approved):
I like this also.

eliaslfox:
I think the data flow would be more difficult to track across multiple functions.

PR #714, opened 2022-08-23
Remove all renamed constructors

eliaslfox:
This increases consistency and makes printing readably easier.

### DIFF
--- a/src/ast/node.lisp
+++ b/src/ast/node.lisp
@@ -17,41 +17,27 @@
 (deftype binding-list ()
   `(satisfies binding-list-p))
 
-;;;;;;;;;;;;;;;;;;;;;;;;; The types of nodes ;;;;;;;;;;;;;;;;;;;;;;;;;
-
-(defstruct
-    (node-literal
-     (:include node)
-     (:constructor node-literal (unparsed value)))
+(defstruct (node-literal (:include node))
   "A literal value. These include things like integers and strings."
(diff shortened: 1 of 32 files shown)


stylewarning (approved):
I don't fully understand the reason for removing BOA constructors. Keyword-argument constructors are a lot slower, and syntactically noisier, since keyword arguments need to be parsed in every call, whereas BOA ("based on order of args") constructors are fast and cheap. (Seems desirable for a compiler that makes a lot of constructor calls.)

gefjon:
It's important that types we want to print readably have default constructors rather than boa constructors; I had to remove all of them in the translate-coalton-file branch to get dumping compiled Coalton code to Lisp files as an intermediate build artifact. Basically, any struct that goes inside of an environment, or otherwise appears in Coalton compiler output, needs to have a default constructor.

(Alternatively, we could define `print-object` methods that use `#.` instead of `#s`, like we do for Coalton `define-types`, but... eugh. I don't want to write and maintain those.)

For structs that won't ever be in compiled output, we could use boa constructors, but I don't think I agree that they're more desirable syntactically; I personally find `(make-node-the :unparsed foo :type bar :subnode baz))` easier to read than just `(make-node-the foo bar baz)`.

If argument list parsing is a real concern, we should be ensuring that calls to constructors are inlined, which will bypass not only kwarg parsing but also argument count and type checks.

PR #718, opened 2022-08-25
fix doc generation re: removal of `tcon` and `tvar` wrapper structs

gefjon:


eliaslfox (approved):
Thanks for fixing this

PR #719, opened 2022-08-25
emit function param names in documentation

gefjon:
in documentation, functions defined in a way that produces parameter
names (i.e. either (define (foo bar) baz) or (define foo (fn (bar) baz))) will now have
those parameter names shown in generated documentation.

where previously the documentation would show

```
FOO [FUNCTION]
```

it now shows

```
(FOO BAR) [FUNCTION]
```

this requires subclassing `name-entry` and `documentation-value-entry` with
`function-name-entry` and `documentation-function-entry` respectively to store the
original source form's `param-names`.

this allows us to reference parameter names in docstrings rather than refering to
parameters by index, like in the documentation for `<`:

```
"Is X less than Y?"
```

### DIFF
--- a/src/doc/generate-documentation.lisp
+++ b/src/doc/generate-documentation.lisp
@@ -46,6 +46,10 @@
 (deftype documentation-value-entry-list ()
   '(satisfies documentation-value-entry-list-p))
 
+(defstruct (documentation-function-entry
+            (:include documentation-value-entry))
+  (param-names (required 'param-names) :type list :read-only t))
+
 (defstruct documentation-file-entry
   (filename      (required 'filename)      :type t                              :read-only t)
   (package       (required 'package)       :type symbol                         :read-only t)
@@ -246,6 +250,22 @@
     file-entries))
 
 
+(defun name-entry->documentation-value-entry (env name name-entry)
+  (let ((documentation (coalton-impl/typechecker::name-entry-docstring name-entry))
+        (location (coalton-impl/typechecker::name-entry-location name-entry))
+        (type (lookup-value-type env name)))
+    (alexandria:if-let ((source-param-names
+                         (coalton-impl/typechecker:lookup-function-source-parameter-names env name)))
+      (make-documentation-function-entry :name name
+                                         :type type
+                                         :documentation documentation
+                                         :location location
+                                         :param-names source-param-names)
+      (make-documentation-value-entry :name name
+                                      :type type
+                                      :documentation documentation
+                                      :location location))))
+
 
 (defun get-doc-value-info (env package)
   (declare (type environment env)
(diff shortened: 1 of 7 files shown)


eliaslfox:
I like this.

Don't subclass `name-entry`. The argument names should be attached to the functions [`function-env-entry`](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/environment.lisp#L368) instead.

gefjon:
> Don't subclass name-entry. The argument names should be attached to the functions [function-env-entry](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/environment.lisp#L368) instead.

I don't believe this is feasible (or at least it would require a lot more work); the function env is constructed based on `codegen/ast`, which I think is after source variable names have been discarded and typeclass instance parameters have been made explicit. I could theoretically thread the source arglist through that additional layer of AST, but doing so would require changing several different sites that construct or copy `codegen/ast:node-abstraction`s, and overall seems sorta silly.

Personally, it feels right to me that parameter names go in the name entry.

eliaslfox:
Parameter names should go on the function env entry because functions have them, and non-function values don't. The environment is a database in 3NF. Each type of environment is a table, and each entry is a row. 

You shouldn't need to store original param names in the codegen ast. Just read them off of the typed-nodes, and store them in a hash table in [this loop](https://github.com/coalton-lang/coalton/blob/main/src/codegen/program.lisp#L50).

gefjon:
> Just read them off of the typed-nodes, and store them in a hash table in [this loop](https://github.com/coalton-lang/coalton/blob/main/src/codegen/program.lisp#L50).

Should I move all of the machinery for creating `function-env-entry`s and inserting them into the environment into that loop? Because that is not where they're constructed currently; right now it's done in [transformations.lisp](https://github.com/coalton-lang/coalton/blob/main/src/codegen/transformations.lisp#L197).

eliaslfox:
I think function envs are constructed in the correct location. The optimizer can rewrite bindings, and add new ones. So we want to calculate the function environment as late as possible.

gefjon:
If the `function-env` is intended to store post-optimization information about the functions we're emitting, then I think it should probably not be used to also store source-level information about functions' user-provided parameter names. For one thing, compiler-generated functions will not have an arglist to store, so their `param-names` slot will be `nil`; for another thing, having `(length param-names)` not equal to `arity` seems confusing.

I could add another `env` slot / table mapping toplevel user-defined functions to their source arglists, but frankly I think that seems unnecessary, and uglier than subclassing `name-entry`. The subclassing version appeals to me because consumers that don't care about function param names can ignore them transparently, and consumers that do care can use `typep`/`typecase` dispatch to determine whether it's available, rather than something like `(alexandria:if-let (function-param-names (function-param-names-lookup env name)) ... ...)`. Still, if you strongly prefer adding a new "table" for this purpose, I can do that.

eliaslfox (approved):

PR #732, opened 2022-09-05
implement Integer -> {Single,Double}-Float as TryInto

stylewarning:
too convenient to not have. open to other suggestions on how to write this conversion


stylewarning:
we could also return `infinity` or `nan`

eliaslfox (approved):
LGTM

PR #734, opened 2022-09-06
Fix codegen bug with memberless classes

eliaslfox:
* move generation of singleton values to struct-or-class.lisp
* remove duplicate function-env constructor generation from
  codegen-type-definition.lisp
* remove unnecessary constructor functions for 0 arity structs/classes

### DIFF
--- a/src/codegen/struct-or-class.lisp
+++ b/src/codegen/struct-or-class.lisp
@@ -43,43 +43,52 @@
         (list
          `(defstruct (,classname
                       (:copier nil)
+                      (:predicate nil)
                       ,@(when superclass
                           (list `(:include ,superclass)))
-                      (:constructor ,constructor ,field-names)
-                      (:predicate nil))
+                      (:constructor ,constructor ,field-names)) 
             ,@(loop :for field :in fields
                     :for name := (struct-or-class-field-name field)
                     :collect `(,name (error ""))))))
 
        (:class
-        (list 
-         `(defclass ,classname
-              ,(if superclass
-                   (list superclass)
-                   (list))
-            ,(loop :for field :in fields
-                   :for name := (struct-or-class-field-name field)
-                   :for package := (symbol-package classname)
-                   :for accessor
-                     := (alexandria:format-symbol package "~A-~A" classname name)
-                   :collect `(,name
-                              :initform (error "")
-                              :initarg ,name
-                              :accessor ,accessor)))
+        (append 
+         (list
+          `(defclass ,classname
+               ,(if superclass
+                    (list superclass)
+                    (list))
+             ,(loop :for field :in fields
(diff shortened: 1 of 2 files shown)


src/codegen/struct-or-class.lisp
```diff
@@ -45,41 +45,50 @@
...
-       (cons
-        `(global-lexical:define-global-lexical ,constructor
-             ,(construct-function-entry `#',constructor (length fields)))
-        (loop :for field :in fields
-              :for package := (symbol-package classname)
-              :for field-name := (alexandria:format-symbol package "~A-~A" classname (struct-or-class-field-name field))
-              :collect `(global-lexical:define-global-lexical ,field-name
-                          ,(construct-function-entry `#',field-name 1))))))))
+         (list
+          `(global-lexical:define-global-lexical ,constructor
```
stylewarning, line 91: `make-instance` is not defined for `structure-class` metaclasses: http://www.lispworks.com/documentation/lw70/CLHS/Body/f_mk_ins.htm

stylewarning (changes requested):

src/codegen/struct-or-class.lisp
eliaslfox: thanks sbcl :/

src/codegen/struct-or-class.lisp
```diff
@@ -43,43 +43,48 @@
...
+               ,(if superclass
+                    (list superclass)
+                    (list))
+             ,(loop :for field :in fields
+                    :for name := (struct-or-class-field-name field)
+                    :for package := (symbol-package classname)
+                    :for accessor
+                      := (alexandria:format-symbol package "~A-~A" classname name)
+                    :collect `(,name
+                               :initform (error "")
```
stylewarning, line 67: I would prefer `:default-initarg` over `:initform`, also with an actual error string (or maybe some sort of `(unreachable)` call)

src/codegen/struct-or-class.lisp
```diff
@@ -43,43 +43,48 @@
...
+                               :initarg ,name
+                               :accessor ,accessor))))
+
+         (list
+          `(defun ,constructor ,field-names
```
stylewarning, line 72: you might want to inline this

stylewarning (approved):
nits, but otherwise fine

src/codegen/struct-or-class.lisp
gefjon: why do you prefer `:default-initarg` over `:initform`?

stylewarning: habit from [this tip](https://lisptips.com/post/11728375873/initform-and-default-initargs/amp)

PR #740, opened 2022-09-21
Add type annotations to toplevel forms

### LINKED ISSUE #731: Add type annotations for top level definitions

colescott:
Resolves #731

Additionally, adds error when recursive toplevel forms are written since these cannot be represented correctly with global lexicals.

### DIFF
--- a/src/global-lexical.lisp
+++ b/src/global-lexical.lisp
@@ -45,10 +45,9 @@
 (defun (setf top-level-binding-value) (new-value binding)
   (setf (car binding) new-value))
 
-(defmacro define-global-lexical (var val)
+(defmacro define-global-lexical (var type)
   `(progn
      (define-symbol-macro ,var
-         (top-level-binding-value
-          (load-time-value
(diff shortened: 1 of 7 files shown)


eliaslfox (changes requested):
add more comments. just like explain whats going on

eliaslfox (approved):

src/global-lexical.lisp
```diff
@@ -45,10 +45,9 @@
 (defun (setf top-level-binding-value) (new-value binding)
   (setf (car binding) new-value))
 
-(defmacro define-global-lexical (var val)
+(defmacro define-global-lexical (var type)
```
stylewarning, line 48: Can we rename this to DEFINE-GLOBAL-COATON-VALUE or something along those lines?

I also think it needs documentation to establish what needs to be true about VAR when this macro is expanded. (How does VAR get a value? When will it get a value?)

The top level definitions are sufficiently tricky, and have evolved a lot since Coalton 0.000000001, so I think some docs would be good.

PR #742, opened 2022-09-22
Add specialized vectors

colescott:
* `VECTOR:NEW` and `VECTOR:WITH-CAPACITY` will now automatically specialize constructed vectors to the lisp type of the elements.
* `SLICE:NEW` will now automatically specialize constructed vector slices
* Add check for superclass context in instance definitions
* Fix type on node generated for codegen on instances with context


eliaslfox (changes requested):
Looks good. Add tests to check that created vectors and slices are actually specialized.

eliaslfox (approved):

PR #754, opened 2022-09-30
Iterator classes `IntoIterator` and `FromIterator`

gefjon:
note FIXME comments in tests/iterator-tests.lisp: type inference fails using the generic versions in some places i would expect to work.

### DIFF
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -117,20 +117,11 @@ STATE, using INIT as the first STATE."
       (%Iterator (fn () (map func (next! iter))))))
 
 ;;; constructors
-  ;; once coalton gets functional dependencies, associated types or type families, much of this will be
-  ;; abstracted into a class `(IntoIterator :collection :item)', with instances like `(IntoIterator (List :elt)
-  ;; :elt)' and `(IntoIterator String Char)'. It's currently not possible for Coalton to do useful type
-  ;; inference on these classes, so we're stuck with monomorphic constructors like `list-iter`.
   (declare empty (Iterator :any))
   (define empty
     "Yields nothing; stops immediately"
     (%Iterator (fn () None)))
 
-  (declare last! ((Iterator :a) -> (Optional :a)))
-  (define (last! iter)
-    "Yields the last element of ITER, completely consuming it."
-    (fold! (fn (s e) (Some e)) None iter))
-  
   (declare list-iter ((List :elt) -> (Iterator :elt)))
   (define (list-iter lst)
     "Yield successive elements of LST.
@@ -252,7 +243,9 @@ Equivalent to reversing `range-increasing`"
   (declare interleave! (Iterator :a -> Iterator :a -> Iterator :a))
   (define (interleave! left right)
     "Return an interator of interleaved elements from LEFT and RIGHT which terminates as soon as both LEFT and RIGHT do.
-In the case one iterator terminates before the other, the other is exhausted before terminating."
+
+If one iterator terminates before the other, elements from the longer iterator will be yielded without
+interleaving. (interleave empty ITER) is equivalent to (id ITER)."
     (let flag = (cell:new False))
     (%Iterator
      (fn ()
@@ -313,6 +306,7 @@ In the case one iterator terminates before the other, the other is exhausted bef
          (zip! iter
                (up-to count))))
 
+  ;; this appears to be a synonym for `concat!'. i would like to remove it - gefjon
   (declare chain! (Iterator :elt -> Iterator :elt -> Iterator :elt))
   (define (chain! iter1 iter2)
     (%iterator
@@ -361,6 +355,12 @@ In the case one iterator terminates before the other, the other is exhausted bef
          keys))
 
 ;;; consumers
+  (declare last! ((Iterator :a) -> (Optional :a)))
+  (define (last! iter)
+    "Yields the last element of ITER, completely consuming it."
+    (fold! (fn (s e) (Some e)) None iter))
+
+  ;; this and `or!' appear to be inefficent (fully forcing) synonyms for `every! id' and `any! id' respectively
   (declare and! (Iterator Boolean -> Boolean))
   (define and! (fold! boolean-and True))
 
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -28,6 +28,8 @@
    #:fold!
    #:empty
    #:last!
+   #:IntoIterator
+   #:into-iter
    #:list-iter
    #:vector-iter
    #:string-chars
@@ -68,6 +70,8 @@
    #:elementwise-match!
    #:elementwise==!
    #:elementwise-hash!
+   #:FromIterator
+   #:collect!
    #:collect-list!
    #:collect-vector-size-hint!
    #:collect-vector!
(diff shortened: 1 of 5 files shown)


library/iterator.lisp
```diff
@@ -496,6 +520,9 @@ The vector will be resized if ITER contains more than SIZE elements."
     "Construct a `Vector` containing all the elements from ITER in order."
     (collect-vector-size-hint! 0 iter))
 
+  (define-instance (types:RuntimeRepr :elt => FromIterator (Vector :elt) :elt)
+    (define collect! collect-vector!))
```
stylewarning, line 524: this looks like a bug

tests/iterator-tests.lisp
```diff
@@ -2,6 +2,8 @@
 
 (define-test iter-list-iter-collect-list-id ()
   (is (== (the (List Integer) (make-list 0 1 2 3))
+          ;; FIXME: changing this to `iter:collect! (iter:into-iter (make-list 0 1 2 3))' breaks type
```
stylewarning, line 5: ISTM that we should sort this out before merging

tests/iterator-tests.lisp
```diff
@@ -13,6 +15,7 @@
           (iter:next! iter:empty))))
 
 (define-test iter-string-chars ()
+  ;; FIXME: changing `string-chars' to `into-iter' here breaks type inference. i don't know why.
```
stylewarning, line 18: same here

stylewarning (changes requested):
there's one question and one thing that appears to be a typo

As for the tests' FIXMEs, I'm not sure if the root cause should be addressed before or after this PR lands.

library/iterator.lisp
```diff
@@ -476,12 +495,17 @@ The empty iterator will hash as 0."
         ((None) 0)))
 
 ;;; collecting
-  ;; as with `IntoIterator`, these will one day be abstracted into a class `(FromIterator :collection :item)'.
+  (define-class (FromIterator :container :elt (:container -> :elt))
```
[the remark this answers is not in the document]
gefjon: i tend to prefer the functional dependency be left-to-right, but it’s not a strong opinion.

library/iterator.lisp
gefjon: what looks like a bug here?

tests/iterator-tests.lisp
gefjon: agreed

library/iterator.lisp
stylewarning: Oh, oops, I misread the definition. No bug.

eliaslfox (approved):
This looks good. The type inference issues should be fixed by #757.

gefjon:
@eliaslfox  #757 seems to have fixed most, but not all, of the type inference issues. Specifically, `iter-string-chars` and `iter-enumerate`, which use `string-chars` to iterate over a string and then manually `next!` off of it and assert the returned character with `==`, both fail to infer that the iterator returned by `into-iter` is over `Char`, and give an error like:

```
iterator-tests.lisp:15:1:
  warning: 
    The expression (PROGN
                    (LET ((ITER (INTO-ITER abcdef)))
                      (PROGN (IS (== (SOME a) (NEXT! ITER)))
                             (IS (== (SOME b) (NEXT! ITER)))
                             (IS (== (SOME c) (NEXT! ITER)))
                             (IS (== (SOME d) (NEXT! ITER)))
                             (IS (== (SOME e) (NEXT! ITER)))
                             (IS (== (SOME f) (NEXT! ITER)))
                             (IS (== NONE (NEXT! ITER))))))
        of type ∀ :A. (EQ (OPTIONAL :A)) => UNIT
        has unresolved constraint (EQ (OPTIONAL :A))
        add a type assertion with THE to resolve it
    --> PROGN DEFUN PROGN SB-IMPL::%DEFUN SB-IMPL::%DEFUN 
    --> SB-INT:NAMED-LAMBDA FUNCTION BLOCK LET* LABELS 
    --> COALTON-NATIVE-TESTS::ITER-STRING-CHARS BLOCK 
    ==>
      (COALTON:COALTON
       (COALTON:PROGN
        (COALTON:LET ((COALTON-NATIVE-TESTS::ITER #))
          (COALTON:PROGN (COALTON-TESTING:IS #) (COALTON-TESTING:IS #)
                         (COALTON-TESTING:IS #) (COALTON-TESTING:IS #)
                         (COALTON-TESTING:IS #) (COALTON-TESTING:IS #)
                         (COALTON-TESTING:IS #)))))
```

eliaslfox:
@gefjon should be fixed in #761

gefjon:
@eliaslfox that fixed all the issues in iterator-tests.lisp, but then I remembered that red-black-tests.lisp also does some iterator-ey stuff, and sadly some of those are still broken. Sigh. Anyway, many thanks for how responsive you've been fixing these.

eliaslfox:
I looked at iterator tests and I don't think the failures are type system bugs. They all appear to be how the type system is "intended" to work. Although many (all?) of the cases are confusing and generate unhelpful error messages. Starting with the main issue: why does the following not compile?

```
(coalton-toplevel
  (declare tree (coalton-library/ord-tree:Tree Integer))
  (define tree (iter:collect! (iter:up-to 4))))
```

The current compiler error is `Ambigious constraint NUM #T36363` which looks like it should be defaulted. However the complete set of predicates is `NUM #T36363` `ORD #T36363` `FROMITERATOR (TREE INTEGER) #T36363`. Coalton won't default ambiguous variables in the predicate of a multi-parameter type class. This is a relaxation of Haskells more restricted defaulting rule which only defaults predicates for classes defined in the standard library.

The following fails to compile in Haskell because `Zero :a` makes the set `Zero :a` `Num :a` non default-able:
```haskell
main = pure ()

class Zero a where
  zero :: a

instance Zero Integer where
  zero = 0

x = 1 + zero
```

The equivalent does compile in Coalton:
```
 (coalton-toplevel
  (define-class (Zero :a)
    (zero :a))
                
  (define-instance (Zero Integer)
    (define zero 0))
                
  (define x (+ 1 zero)))
```

But with a multi-parameter class it does not:
```
(coalton-toplevel
  (define-class (Add :a :b)
    (add (:a -> :b -> :a)))
                
  (define-instance (Add Integer Integer)
    (define add +))
                
  (define x (+ (add 1 2) 3)))
```

Coalton's defaulting rules are somewhat arbitrary and multi-parameter type class defaulting could be added. Personally I haven't run into issues with Haskell's even more restrictive defaulting rules; but I don't do a lot of numerical programming. The original code can be made to compile by type asserting the input number:

````
(coalton-toplevel
  (declare tree (coalton-library/ord-tree:Tree Integer))
  (define tree (iter:collect! (iter:up-to (the Integer 4)))))
```

The other type inference issue is in the test `tree-always-increasing-order`. Replacing the last `red-black/tree:collect!` with an `iter:collect!` makes the collected type ambiguous. The iterators could be collected into vectors or lists another type and the code would still be "correct." The other two functions can use `iter:collect!` because their return values are compared with the return value of the first. If one function returns a red black tree then all three do.

I opened an issue to improve error messages about ambiguous types in #762. I think adding more context should make the error a lot easier to understand. I don't think there are any more type system bugs; so this pr should be good to merge.

gefjon:
> The other type inference issue is in the test tree-always-increasing-order. Replacing the last red-black/tree:collect! with an iter:collect! makes the collected type ambiguous. The iterators could be collected into vectors or lists another type and the code would still be "correct." The other two functions can use iter:collect! because their return values are compared with the return value of the first. If one function returns a red black tree then all three do.

Makes sense. Will fix with an explicit `the` declaration.

> Coalton won't default ambiguous variables in the predicate of a multi-parameter type class. This is a relaxation of Haskells more restricted defaulting rule which only defaults predicates for classes defined in the standard library.

In addition to improving error messages, could you document this behavior? I find it surprising that wrapping an integer literal in `the Integer` would ever change the way the compiler treats a program, and while I don't necessarily disagree with it, I'd like to have a good English-language description of why it happens, in what circumstances, and the rationale behind that decision that I can refer back to and point others toward.

eliaslfox:
> In addition to improving error messages, could you document this behavior?

yes definitely

> I find it surprising that wrapping an integer literal in the Integer would ever change the way the compiler treats a program

What about `(+ 0.5 1)` vs `(+ 0.5 (the Integer 1))`?

gefjon:
> What about (+ 0.5 1) vs (+ 0.5 (the Integer 1))?

Well sure, I mean, if there are other constraints on the literal that make `Integer` not a valid unification clearly wrapping it in `the Integer` will cause a type error. But it surprises me that there are cases where an unconstrained integer literal causes an error but adding a constraint results in a valid program.

stylewarning (approved):

PR #757, opened 2022-10-05
Fix several type inference related issues

eliaslfox:
* Fundep improvements were not always being applied in `coalton` forms
* Fundep impvovements when matching against more general instances were not always improving types

Example:
`(define-class (C :a :b (:a -> :b))`
`(define-instance (C (List :a) :a)`
The predicate `(C (List String) :c)` should become `(C (List String) String)` but instead it was becomming `(C (List String) :a)`

* Is assertions in tests were lifting values into let expressions which effected type inference. Is assertions will now use shorthand let forms to preserve the intended type checking behavior.


colescott (approved):
lgtm

PR #760, opened 2022-10-07
Add basic Char functions

### LINKED ISSUE #752: Add char functions
- ascii_letter
- ascii_digit
- uppsercase
- lowercase

colescott:
Resolves #752

### DIFF
--- a/library/char.lisp
+++ b/library/char.lisp
@@ -42,7 +42,77 @@
           EQ
           (if (lisp Boolean (x y) (to-boolean (cl:char> x y)))
               GT
-              LT)))))
+              LT))))
+
+  (declare alpha? (Char -> Boolean))
+  (define (alpha? c)
+    "Is C an alphabetic character?"
+    (lisp Boolean (c)
+      (cl:alpha-char-p c)))
+
+  (declare ascii-alpha? (Char -> Boolean))
+  (define (ascii-alpha? c)
+    "Is C an ASCII alphabetic character?"
+    (lisp Boolean (c)
+      (cl:or
+       (cl:<= 65 (cl:char-code c) 90)
+       (cl:<= 97 (cl:char-code c) 122))))
+
+  (declare digit? (Char -> Boolean))
+  (define (digit? c)
+    "Is C a digit character?"
+    (lisp Boolean (c)
+      (to-boolean (cl:digit-char-p c))))
+
+  (declare ascii-digit? (Char -> Boolean))
+  (define (ascii-digit? c)
+    "Is C an ASCII digit character?"
+    (lisp Boolean (c)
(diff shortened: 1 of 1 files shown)


eliaslfox (approved):

library/char.lisp
```diff
@@ -42,7 +42,77 @@
...
+    "Is C an alphabetic character?"
+    (lisp Boolean (c)
+      (cl:alpha-char-p c)))
+
+  (declare ascii-alpha? (Char -> Boolean))
+  (define (ascii-alpha? c)
+    "Is C an ASCII alphabetic character?"
+    (lisp Boolean (c)
+      (cl:or
+       (cl:<= 65 (cl:char-code c) 90)
```
stylewarning, line 58: Note that CHAR-CODE or CHAR-INT has no portable guarantees around encoding.

eliaslfox: We could make these SBCL only? The standard doesn't seem to say anything useful about the representation of chars

gefjon: I have a hard time believing any implementation today is not encoding base-chars as ascii. Making these functions SBCL-only seems silly, but perhaps documentation could note that it relies on the implementation's `code-char` and `char-code` using ascii encoding for base-chars.

stylewarning: It also wouldn't be hard to just make a $2^7$-sized bitmap.

gefjon: I don't think that actually helps; how do you get an index into the bitmap without making assumptions about `char-code`/`code-char`?

gefjon: Anyway, this function could be written portably as `(and standard-char-p alpha-char-p)`.

stylewarning: Here's a schematic of how I was imagining bitmaps. With some more work it could be extended to cover all 96 `standard-char`s. 

```
(deftype char-code ()
  `(integer 0 (,char-code-limit)))

(defconstant min-char-code (min (char-code #\0) (char-code #\a) (char-code #\A)))

(defconstant max-char-code (min (char-code #\9) (char-code #\z) (char-code #\Z)))

(defconstant alphanumeric-code-range (- max-char-code min-char-code))

(defconstant* ascii->code (make-array 128 :element-type 'char-code :initial-contents ...))

(defconstant* code->ascii (make-array alphanumeric-code-range :element-type '(integer 0 (128)) :initial-contents ...))

(defconstant* upper-bitmap (make-array alphanumeric-code-range :element-type 'bit :initial-contents ...)
```

The `...` are an exercise to the reader. 

I agree that the `(and standard-char-p etc)` is better for this implementation, but I just wanted to share what is certainly portable (though not guaranteed efficient).

PR #765, opened 2022-10-14
Char I/O, take 2

gefjon:
Per discussion on Discord, this is a (currently early draft) proposal for a Coalton character streams interface, with a bit of hindsight since #556 .

The core of this interface is three typeclasses:
- `Stream`, with `open?` and `close!` methods.
- `Input`, with `read-char!` and `read-line!` methods.
- `Output`, with `write-char!`, `write-line!` and `flush-output!` methods.

These can be converted to a form that interoperates with Common Lisp using [trivial-gray-streams](https://github.com/trivial-gray-streams/trivial-gray-streams). There's some room to optimize that conversion; in particular, it should be possible to determine that Coalton streams which are already Lisp streams do not need to be wrapped up in gray-streams CLOS objects before passing them back to Lisp.

Not yet included:
- Meaningful error handling. 
  - I defined Coalton analogues for all of SBCL's stream-error classes, but they're all empty because I didn't want to figure out their slots.
- Interfaces for opening and interacting files.
- Stream operators beyond the fundamental ones. 
  - I did define `write-line!` as a proof-of-concept, but CL has a lot more stream functions that I didn't bother with.
  - We also need iterators over stream-chars and stream-lines for input streams, and iterator consumers for output-streams.
- Any notion of printing Coalton objects aside from Chars and Strings.
  - Do we want some sort of `ToString`/`PrettyPrint`/`Display`/`Debug` typeclasses?
- Tests.
- Documentation.


Jason94:
This looks really promising to me. 

Here is some code that would integrate the IO monad. It also defines a `char-stream-pure` package that uses a macro to wrap all of the public functions `char-stream` exposes in an IO computation.

https://github.com/gefjon/coalton/compare/user-streams...Jason94:coalton:user-streams-io

1. I added functions to capture `*standard-input*`/`*standard-output*` as defined in #556.
2. I also added two helper functions `read-line-std!` and `write-line-std!` that don't return a result. This was just for convenience, we don't need them. What would probably be ideal is to use monad transformers to print out an error message and return if you get a left result, otherwise keep going if you get a right result. [Something like this](https://www.fpcomplete.com/haskell/tutorial/monad-transformers/). That can come later, though.
3. To convert the unsafe functions names to a safe function name, I'm just removing the `!` off the end. So `write-line! (:stream -> String -> Result StreamError Unit)` becomes `write-line (:stream -> String -> IO (Result StreamError Unit)`. This works very well for every function _except_ `open?`, which does not end in a `!`. We definitely need a monadic version of `open?` to work with, so I used `pure-open?`. It would be nice if the symbols didn't conflict between the packages, but not necessary, so it could just be `open? (:stream -> IO Boolean)` instead.

gefjon:
> It would be nice if the symbols didn't conflict between the packages, but not necessary, so it could just be open? (:stream -> IO Boolean) instead.

I think shadowing the same name is probably better, assuming the monadic operations are going in a separate package like you've defined them.

> We definitely need a monadic version of open? to work with

This is not obvious to me, as in my mind `open?` is a pure function already. It can observe state changes made by `close!` and friends, but cannot itself cause any observable state changes. Perhaps someone with more experience designing and working with monadic APIs (@eliaslfox ? @colescott ?) could weigh in.

I'll also note that I object morally to your code's use of "safe" (e.g. in `define-safe`) to describe monadic operations. Non-monadic IO is unsafe in Haskell and other non-strict languages, but because Coalton is strict it's completely safe and will behave as expected. I also think "pure" is incorrect, as you're defining a monad that encodes impure operations, but I find that less objectionable. I think I would feel most comfortable with "typed" or "monadic" (or just "monad") as a name differentiator, but I'm sure I could be easily convinced some other way.

Jason94:
> 
> > We definitely need a monadic version of open? to work with
> 
> This is not obvious to me, as in my mind `open?` is a pure function already. It can observe state changes made by `close!` and friends, but cannot itself cause any observable state changes. Perhaps someone with more experience designing and working with monadic APIs (@eliaslfox ? @colescott ?) could weigh in.

That's true, `open?` is pure in the sense that it doesn't cause any side effects. It can still cause problems when combined with an IO monad, though. Here is an example to demonstrate. First, here is an imperative function that opens a stream, prints its open state, closes the stream, and then prints the open state:

```Common Lisp
(coalton-toplevel
  (define (test-open-unbound)
    (match (open-file "numbers.txt" Optionally)
      ((None) Unit)
      ((Some stream)
       (progn
         (trace (show (cs:open? stream))) ;; Prints T
         (cs:close! stream) ;; Stream is now closed
         (trace (show (cs:open? stream))) ;; Prints nil
         Unit))))
```

Here is a function that uses just the IO-bound API. What's actually happening in the `do` block is a series of anonymous functions threading the IO state is wrapped up, and then that anonymous function is executed by the `run!` command. Each line of code in the `do` block doesn't actually perform any read/writes/checks-for-openness, it just wraps those things in anonymous function calls.
```Common Lisp
  (define (test-open-monadic-working)
    (run!
      (match (open-file "numbers.txt" Optionally)
        ((None) (cs-iom:write-line (standard-output) "Failed to open file"))
        ((Some stream)
         (do
           (is-open-a <- (cs-iom:open? stream))
           (cs-iom:write-line (standard-output) (show is-open-a)) ;; Prints T
           (cs-iom:close stream) ;; Will be closed from here when the IO is run
           (is-open-b <- (cs-iom:open? stream))
           (cs-iom:write-line (standard-output) (show is-open-b)))))))) ;; Prints nil
```

This means that by the time `is-open-b` is calculated, `close` has been executed. But that happens up in the `run!` function, not when `cs-iom:close` runs the first time. So if you mix this with non-IO bound functions that check the state of the stream, it can go wrong.

This code does almost the same thing, but it uses the non-IO bound `cs:open?` function:
```Common Lisp
  (define (test-open-monadic-broken)
    (run!
      (match (open-file "numbers.txt" Optionally)
        ((None) (cs-iom:write-line (standard-output) "Failed to open file"))
        ((Some stream)
         (do
           (is-open-a <- (cs-iom:open? stream))
           (cs-iom:write-line (standard-output) (show is-open-a)) ;; Prints T
           (cs-iom:close stream) ;; Will be closed from here when the IO is run
           (let is-open-b = (cs:open? stream))
           (cs-iom:write-line (standard-output) (show is-open-b))))))) ;; Prints T!
```

Since `cs:open?` executes when the line `(let is-open-b = (cs:open? stream))` executes, `is-open-b` is calculated _before_ the stream is closed, even though it looks like the stream was already closed on the previous line.

The takeaway from this is that you really shouldn't mix monadic and non-monadic IO ever. For that reason I think it is a good idea to split `char-stream` and `char-stream-monadic` into three files: one that defines the stream types and typeclasses, one with unbound read/write/open/close/etc functions, and one with the IO-bound functions. Then there's very clear guidance: "Import the file that defines the stream types (if you need them). Then import the file that does the type of IO that you want." Not "Import some stuff from the first file. If you want monadic IO then make sure you don't import some of the stuff from the first file, just the right stuff."

> I'll also note that I object morally to your code's use of "safe" (e.g. in `define-safe`) to describe monadic operations. Non-monadic IO is unsafe in Haskell and other non-strict languages, but because Coalton is strict it's completely safe and will behave as expected. I also think "pure" is incorrect, as you're defining a monad that encodes impure operations, but I find that less objectionable. I think I would feel most comfortable with "typed" or "monadic" (or just "monad") as a name differentiator, but I'm sure I could be easily convinced some other way.

I agree. I renamed it to `char-stream-monadic` and `define-monadic` in my branch. I'm open to other names.

library/char-stream.lisp
```diff
@@ -0,0 +1,427 @@
...
+  (declare %abstract-stream-open? (%AbstractStream -> Boolean))
+  (define (%abstract-stream-open? stream)
+    (lisp Boolean (stream)
+      (cl:open-stream-p stream)))
+
+  (declare %abstract-stream-close! (%AbstractStream -> Result StreamError Unit))
+  (define (%abstract-stream-close! stream)
+    (if (%abstract-stream-open? stream)
+        (Ok (lisp Unit (stream)
+              (cl:close stream)))
```
Jason94, line 118: This throws an error at runtime because `(cl:close stream)` returns a boolean for success, not a unit.

library/char-stream.lisp
```diff
@@ -0,0 +1,427 @@
...
+         (lisp ,to-type (,instance)
+           ,instance)))))
+
+(coalton-toplevel
+  ;; users should interact with streams via the classes `Input' and `Output', rather than explicitly
+  ;; referencing these types
+  (repr :native cl:stream)
+  (define-type %AbstractStream)
+  
+  (repr :native (cl:and cl:stream (cl:satisfies cl:input-stream-p)))
```
Jason94, line 65: Representing these types these way is really clever. Unfortunately `input-stream-p` starts returning `nil` for file streams once they've been closed, and that was causing me runtime errors when I was experimenting with it. In my branch I had to change these three types to just `(repr :native cl:stream)` to get around this.

It's not as elegant, and there might be a better solution. If we do it that way, we'll have to rely on exposing an API that is very careful to only return these types if they're actually that kind of stream.

Here's a CL sample to demonstrate:
```Common Lisp
(in-package :cl-user)

(defun main ()
  (let ((stream (open "numbers.txt")))
    (print (input-stream-p stream)) ;; Prints T
    (close stream)
    (print (input-stream-p stream)))) ;; Prints nil
```

gefjon: > input-stream-p starts returning nil for file streams once they've been closed

It wHAT? That's absurd. I suspect it's SBCL weirdness and not a portable behavior; I can't find anything about it in the spec. Oh well, there go the cool types. `%InputStream` should definitely contain closed streams, so I guess all of those types will just be `cl:stream`. Sigh.

library/char-stream.lisp
gefjon: Oops.

gefjon:
> Then there's very clear guidance: "Import the file that defines the stream types (if you need them). Then import the file that does the type of IO that you want." Not "Import some stuff from the first file. If you want monadic IO then make sure you don't import some of the stuff from the first file, just the right stuff."

Yeah, this seems sensible.

> one that defines the stream types and typeclasses, one with unbound read/write/open/close/etc functions, and one with the IO-bound functions

The defs file seems superfluous. If we're gonna define the monadic operations in terms of the strict operations, we should just have two files, and have the monadic package re-export typeclasses and stuff if it wants to.

gefjon:
> Since cs:open? executes when the line (let is-open-b = (cs:open? stream)) executes, is-open-b is calculated before the stream is closed, even though it looks like the stream was already closed on the previous line.

This seems like a bug in our implementation of `do` to me, though the Haskellers might disagree. It seems to me that
```
(do
  some-monad
  (let foo = some-pure-computation)
  another-monad)
```
should expand to something like
```
(>>= some-monad (fn (_) (let foo = some-pure-computation) another-monad))
```
whereas at present it expands into
```
(>>= some-monad
     (coalton::const (coalton::bind foo some-pure-computation another-monad)))
```
which eagerly evaluates the `bind` form (imo incorrectly).

Jason94:
> > one that defines the stream types and typeclasses, one with unbound read/write/open/close/etc functions, and one with the IO-bound functions
> 
> The defs file seems superfluous. If we're gonna define the monadic operations in terms of the strict operations, we should just have two files, and have the monadic package re-export typeclasses and stuff if it wants to.

Yeah, that's a good idea. I pushed that up to my branch.

> > Since cs:open? executes when the line (let is-open-b = (cs:open? stream)) executes, is-open-b is calculated before the stream is closed, even though it looks like the stream was already closed on the previous line.
> 
> This seems like a bug in our implementation of `do` to me, though the Haskellers might disagree. It seems to me that
> 
> ```
> (do
>   some-monad
>   (let foo = some-pure-computation)
>   another-monad)
> ```
> 
> should expand to something like
> 
> ```
> (>>= some-monad (fn (_) (let foo = some-pure-computation) another-monad))
> ```
> 
> whereas at present it expands into
> 
> ```
> (>>= some-monad
>      (coalton::const (coalton::bind foo some-pure-computation another-monad)))
> ```
> 
> which eagerly evaluates the `bind` form (imo incorrectly).

If by pure you mean "no side effects", then `open?` is certainly pure. (That's a fine definition of 'pure' to me, though some Haskellers might disagree, as you said.) But it's not referentially transparent, since it can return different results with the same stream object. Referential transparency is one of the design goals of the IO monad, so for that reason I think it makes sense to also wrap `open?`. And from a practical standpoint, even if the above is true of `do` notation, having a bound `open?` makes it possible to use with the normal `map`, `lifta2`, and `>>=` functions outside of a `do` block.

Regardless, what do you think is the best way forward with the whole project? It sounds like this will stand separately from the monad package, so do you want to work on the unbound char-stream first? I'd be happy to follow up with a separate PR for the monadic IO once this PR gets merged in. Alternatively you could merge my branch into this PR and do them together. The former sounds more straightforward to me.

library/char-stream.lisp
stylewarning: You might interpret this conclusion from a strict reading of the spec:

Glossary:

> input adj. (of a stream) supporting input operations (i.e., being a ``data source''). 

Function INPUT-STREAM-P:

> input-stream-p returns true if stream is an input stream

Function CLOSE:

> close closes stream. Closing a stream means that it may no longer be used in input or output operations.

stylewarning: (still bogus tho)

library/char-stream.lisp
```diff
@@ -0,0 +1,818 @@
...
+  (define-type %AbstractStream
+    "A Common Lisp `cl:stream' object, with no claims made as to its direction.
+
+Users should interact with streams via the `Stream' class, rather than explicitly referencing this type.")
+
+  ;; it would be nice if this type could be `(and stream (satisfies input-stream-p))', but on SBCL
+  ;; `input-stream-p' and `output-stream-p' return nil for closed input- and output-streams, so that breaks
+  ;; after you close a stream.
+  (repr :native cl:stream)
+  (define-type %InputStream
```
eliaslfox, line 137: Export `%InputStream` `%OutputStream` and `%TwoWayStream`. Un-exported types won't show up on the doc generator.

eliaslfox:
This looks really good. I'm excited to get file io in the standard library.

I'd like to support `cl:file-position`. My use case has been rereading files to find the line associated with a given character position. I'm not sure what the best way to model streams that do and don't have file positions. We could have `FileInputStream` instead of `%InputStream` and a `FileStream` type class?

> Do we want some sort of ToString/PrettyPrint/Display/Debug type classes?

Definitely eventually. I don't think it's necessary for this pr.

> This seems like a bug in our implementation of do to me

Filed an issue #771

eliaslfox:
> There's some room to optimize that conversion; in particular, it should be possible to determine that Coalton streams which are already Lisp streams do not need to be wrapped up in gray-streams CLOS objects before passing them back to Lisp.

This would change which set of conditions get signalled on error.

library/char-stream.lisp
```diff
@@ -0,0 +1,818 @@
...
+    StreamErrorEndOfFile
+    StreamErrorDecoding
+    StreamErrorEncoding
+    StreamErrorTimeout
+    StreamErrorReader)
+
+  (define-class (Stream :stream)
+    "A character stream which, while open, is potentially readable and/or writeable.
+
+(open? STREAM) should return true if STREAM has not yet been `close!'d.
```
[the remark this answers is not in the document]
gefjon: We should formalize a language for doc comments that includes a correct way to include identifiers and code snippets.

library/char-stream.lisp
gefjon: I don't really feel good about exporting types that should not be referred to by user code... would it be acceptable if instead I added a note to the docstring of the `open-*-file` functions that they return a concrete type which should not be named by user code, and stuck in an example for how user code can interact with streams using type variables and class constraints?

gefjon:
>  I'm not sure what the best way to model streams that do and don't have file positions. We could have FileInputStream instead of %InputStream and a FileStream type class?

Yes, this seems good, except I'd like to keep the concrete stream types not user visible, so I'd call it `%FileInputStream`. I'm not sure if there's any way in portable Common Lisp to determine if a given stream supports `file-position`, though. Could we kick the can on that one and have it in a later PR?

> This would change which set of conditions get signalled on error.

Hmm, yeah. I wonder if, once we get actual slots into `StreamError`, we can remove the `coalton-stream-error` class and just throw actual `cl:stream-error`s instead.

eliaslfox:
> Yes, this seems good, except I'd like to keep the concrete stream types not user visible, so I'd call it %FileInputStream.

I don't think this is a good idea. The types are still user visible. They'll still show up in error messages and function signatures. I think returning unexpected types is more likely to be confusing and frustrating than helpful. FWIW Rust forbids having un-exported types in the signatures of exports.

>  I'm not sure if there's any way in portable Common Lisp to determine if a given stream supports file-position

Other than calling file-position I don't think there is. However any stream that comes from a file Coalton opened is safe.

> Could we kick the can on that one and have it in a later PR?

I'd prefer to have it land in this one with all of the other file infrastructure.

Jason94:
> > Yes, this seems good, except I'd like to keep the concrete stream types not user visible, so I'd call it %FileInputStream.
> 
> I don't think this is a good idea. The types are still user visible. They'll still show up in error messages and function signatures. I think returning unexpected types is more likely to be confusing and frustrating than helpful. FWIW Rust forbids having un-exported types in the signatures of exports.

Is there a way to export the type but not export the data constructor? I did that a lot at a previous job using [opaque types in Flow](https://flow.org/en/docs/types/opaque-types/), and it was a great pattern.

That would allow someone to annotate their stream if they were getting it from Common Lisp interop, but would encourage safe creation of streams from within Coalton.

eliaslfox:
Yes, you just have to give the type and constructor different names. See [this](https://github.com/coalton-lang/coalton/blob/main/library/iterator.lisp#L87-L90) for an example. These stream types are just lisp objects so they don't have constructors.

PR #775, opened 2022-11-01
Add initial coalton file parser

eliaslfox:
This is the first pr for supporting Coalton code in external files. It defines a new parser subsystem and AST which will eventually replace the existing ones. The new parser uses [Concrete-Syntax-Tree](https://github.com/s-expressionists/Concrete-Syntax-Tree) and [Eclector](https://github.com/s-expressionists/Eclector) to read s-expressions with source information. 

@colescott developed a new error system based on Rust's which can generate much better errors than were possible before. A few examples below:

When parsing a class with an invalid variable:
```
(define-class (C a))
```

New error message:

```
error: Invalid type variable
  --> /Users/ejlawsonfox/lisp/coalton/test.coalton:3:24
   |
 3 | (define-class (C a))
   |                   ^ expected keyword symbol
help: add `:` to symbol
 3 | (define-class (C :a))
   |                  --
```
Previous error message:

```
Failed to parse (C A)
invalid type class predicate
In DEFINE-CLASS for class C
```

When parsing an if without an else clause:
```
(define x (if y 5))
```
New error message:
```
error: Malformed if expression
  --> /Users/ejlawsonfox/lisp/coalton/test.coalton:3:24
   |
 3 | (define x (if y 5))
   |                   ^ expected a form
```

Previous error message falls through to sbcl:

```
Error while parsing arguments to DEFMACRO IF:
  too few elements in
    (Y 5)
  to satisfy lambda list
    (EXPR THEN ELSE):
  exactly 3 expected, but got 2
```

The new parser should parse all existing syntax, however there is still more work to be done. Macro expansion is not yet implemented. The main challenge is that macros can generate forms without source information. The current error system relies entirely on source information to generate error messages and will require further extension.

Additional work will also be needed to allow running the new parser in the coalton and coalton-toplevel macros. Macro based invocations will require either generating errors without source information or somehow getting source information from SBCL.

Test coverage for the parser is pretty good. Every branch is hit and every error message is signalled. Coverage can be generated locally with `make parser-coverage`. 

Screenshot_2022-11-01 Screenshot

Also note this this pr is into the branch coalton-file which will be used to track work on coalton files.


gefjon:
Very cool!

> Macro expansion is not yet implemented. The main challenge is that macros can generate forms without source information. The current error system relies entirely on source information to generate error messages and will require further extension.

I see two ways forward for this. The version that makes writing macros easy at the cost of error messages is, when Coalton calls `macroexpand` on a form A to produce a form B, it copies A's source information into any parts of B that don't have source information already. The version that generates better error messages has been explored by Racket's syntax objects; essentially, macros are responsible for adding source information tags to their output.

> Additional work will also be needed to allow running the new parser in the coalton and coalton-toplevel macros. Macro based invocations will require either generating errors without source information or somehow getting source information from SBCL.

I am somewhat loathe to suggest this, but this seems like a classic use-case for reader macros. If we replace `(coalton ...)` and `(coalton-toplevel ...)` with, say, `#{coalton ...}` or `#{coalton-toplevel ...}`, we can parse the `...` the way we want. It's not ideal ergonomically, as it will require any Common Lisp code that wants to invoke Coalton that way to use a named-readtable, but I think it's probably better than maintaining two separate parsers which generate different error messages.

eliaslfox:
> The version that makes writing macros easy at the cost of error messages is, when Coalton calls macroexpand on a form A to produce a form B, it copies A's source information into any parts of B that don't have source information already.

This seems quite implementable. 

> I am somewhat loathe to suggest this, but this seems like a classic use-case for reader macros

That's the current plan.

stylewarning (changes requested):
some initial review

src/parser/base.lisp
```diff
@@ -0,0 +1,344 @@
...
+  (declare (type sb-sys::fd-stream file)
+           (type integer index)
+           (values string &optional))
+  (file-position file 0)
+  (loop :for i :from 1 :to index
+        :for line := (read-line file)
+        :when (= i index)
+          :do (return-from get-nth-line line))
+
+  (util:unreachable))
```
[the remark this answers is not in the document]
eliaslfox: good catch

src/parser/base.lisp
```diff
@@ -0,0 +1,344 @@
...
+Returns (VALUES LINE-NUM LINE-START-INDEX)"
+  (declare (type sb-sys::fd-stream file)
+           (type integer index)
+           (values integer integer))
+  (file-position file 0)
+  (let ((line 1)
+        (line-start-index 0))
+    (loop :for i :to index
+          :for char := (read-char file)
+          :when (char= char #\Newline)
```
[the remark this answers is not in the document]
eliaslfox: yes :(

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+           ;; Read unspecified floats as double floats
+           (*read-default-float-format* 'double-float)
+
+           ;; Read the (package) form
+           (package-form (eclector.concrete-syntax-tree:read file nil 'eof)))
+
+      (when (eq package-form 'eof)
+        (error "file starts with eof"))
+
+      (setf *package* (parse-package package-form file))
```
stylewarning, line 321: is there a reason to not bind `*package*` instead of setting it for the context of this function?

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+      (let* ((program (make-program :package *package* :file file))
+
+             (attributes (make-array 0 :adjustable t :fill-pointer t)))
+
+        (block parse-loop
+          (loop :with elem := nil
```
stylewarning, line 328: you can use

```
(loop :named parse-loop
```
 
instead of the `block`

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+                :do (setf elem (eclector.concrete-syntax-tree:read file nil 'eof))
+
+                :when (eq elem 'eof)
+                  :do (return-from parse-loop)
```
stylewarning, line 332: actually, if all you're doing is returning, I think you can just do `(return)`

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+                :do (when (and (parse-toplevel-form elem program attributes file)
+                               (plusp (length attributes)))
+                      (util:coalton-bug "parse-toplevel-form indicated that a form was parsed but did not
+consume all attributes"))))
+
+        (unless (zerop (length attributes))
```
stylewarning, line 339: zerop length => endp

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+        program))))
+
+(defun parse-package (form file)
+  "Parses a coalton package decleration in the form of (package {name})"
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values package))
+
+  ;; Package declarations must start with "PACKAGE"
+  (unless (equalp (symbol-name (cst:raw (cst:first form))) "PACKAGE")
```
stylewarning, line 360: `string=`

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+  ;; Toplevel forms must begin with an atom
+  (when (cst:consp (cst:first form))
+    (error 'parse-error
+           :err (coalton-error
+                 (cst:first form) file
+                 :message "Malformed toplevel form"
+                 :primary-note "unexpected list")))
+
+  (case (cst:raw (cst:first form))
+    (coalton:monomorphize
```
stylewarning, line 416: again I might recommend `((coalton:monomorphize) ...` for cases

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+          (setf (fill-pointer attributes) 0)
+          (setf (toplevel-define-monomorphize define) monomorphize)
+          (push define (program-defines program))
+          (return-from parse-toplevel-form t)))
+
+    (coalton:declare
+     (let ((declare (parse-declare form file))
+
+           (monomorphize)
```
stylewarning, line 483: no parens

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+                    (setf monomorphize attribute)
+                    (setf monomorphize-form attribute-form))))
+
+       (setf (fill-pointer attributes) 0)
+       (setf (toplevel-declare-monomorphize declare) monomorphize)
+       (push declare (program-declares program))
+       (return-from parse-toplevel-form t)))
+
+    (coalton:define-type
+        (let* ((type (parse-define-type form file))
```
stylewarning, line 528: just `let`, and uninitialized vars parens

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+          (setf (fill-pointer attributes) 0)
+          (setf (toplevel-define-type-repr type) repr)
+          (push type (program-types program))
+          (return-from parse-toplevel-form t)))
+
+    (coalton:define-class
+        (let ((class (parse-define-class form file)))
+
+          (unless (zerop (length attributes))
```
stylewarning, line 577: endp

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+                           :span (toplevel-define-class-head-src class)
+                           :message "while parsing define-class")))))
+
+          (push class (program-classes program))
+          (return-from parse-toplevel-form t)))
+
+    (coalton:define-instance
+        (let ((instance (parse-define-instance form file)))
+
+          (unless (zerop (length attributes))
```
stylewarning, line 596: endp

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+
+(defun parse-define (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values toplevel-define))
+
+  (assert (cst:consp form))
+
+  (let ((name)
```
stylewarning, line 628: push the let down and possibly use multiple-value-bind

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+                  :message "Invalid toplevel form"
+                  :primary-note "unknown toplevel form")))))
+
+
+(defun parse-define (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values toplevel-define))
+
+  (assert (cst:consp form))
```
stylewarning, line 626: does this need to change?

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+     :body body
+     :monomorphize nil
+     :source (cst:source form))))
+
+(defun parse-declare (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values toplevel-declare))
+
+  (assert (cst:consp form))
```
stylewarning, line 665: does this need to change?

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+    ;; (define-type)
+    (unless (cst:consp (cst:rest form))
+      (error 'parse-error
+             :err (coalton-error
+                   form file
+                   :message "Malformed type definition"
+                   :primary-note "expected body")))
+
+    (if (cst:atom (cst:second form))
```
stylewarning, line 729: prefer cond over if for stateful actions

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+     :source (cst:source form))))
+
+(defun parse-define-class (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values toplevel-define-class))
+
+  (assert (cst:consp form))
+
+  (let ((unparsed-name)
```
stylewarning, line 807: push lower

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+      ;; (... =>)
+      (when (and left right (null (cdr right)))
+        (error 'parse-error
+               :err (coalton-error
+                     (cst:second form) file
+                     :message "Malformed class definition"
+                     :primary-note "missing class name")))
+
+      (if (null right)
```
stylewarning, line 892: cond

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+        :head-src (cst:source (cst:second form))))))
+
+(defun parse-define-instance (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values toplevel-define-instance))
+
+  (assert (cst:consp form))
+
+  (let ((unparsed-context)
```
stylewarning, line 991: push let down

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+           (values (or null string) node-body))
+
+  (let ((docstring)
+        (unparsed-body))
+
+    ;; (define y 2)
+    (when (cst:atom (cst:rest form))
+      (return-from parse-definition-body (values nil (parse-body form enclosing-form file))))
+
+    (if (and (cst:atom (cst:first form))
```
stylewarning, line 1298: cond

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+
+    (values docstring (parse-body unparsed-body enclosing-form file))))
+
+(defun parse-instance-method-definition (form parent-form file)
+  (declare (type cst:cst form)
+           (type cst:cst parent-form)
+           (type sb-sys::fd-stream file)
+           (values instance-method-definition))
+
+  (let* ((name)
```
stylewarning, line 1314: let* -> let; extra parens

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+    (unless (cst:consp (cst:rest form))
+      (error 'parse-error
+             :err (coalton-error
+                   form file
+                   :message "Malformed method definition"
+                   :primary-note "expected definition name"
+                   :notes (list context-note))))
+
+
+    (setf (values name arguments) (parse-argument-list (cst:second form) file))
```
stylewarning, line 1356: i'd M-V-B

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+           :err (coalton-error
+                 form file
+                 :message "Malformed functional dependency"
+                 :primary-note "unexpected dotted list")))
+
+  (multiple-value-bind (left right)
+      (util:take-until
+       (lambda (cst)
+         (and (cst:atom cst)
+              (eq (cst:raw cst) 'coalton:->)))
```
stylewarning, line 1387: prefer unparenthesized constant first

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+       (cst:listify form))
+
+    (unless left
+      (error 'parse-error
+             :err (coalton-error
+                   form file
+                   :message "Malformed functional dependency"
+                   :primary-note "expected one or more type variables")))
+
+    (unless (cdr right)
```
stylewarning, line 1397: cdr -> rest?

src/parser/parser.lisp
```diff
@@ -0,0 +1,1491 @@
...
+  (unless (cst:consp (cst:rest form))
+    (error 'parse-error
+           :err (coalton-error
+                 form file
+                 :highlight :end
+                 :message "Malformed repr attribute"
+                 :primary-note "expected keyword symbol")))
+
+  (let ((type (parse-type-variable (cst:second form) file)))
+    (if (eq (keyword-src-name type) :native)
```
stylewarning, line 1446: cond

src/parser/pattern.lisp
```diff
@@ -0,0 +1,120 @@
...
+           (type sb-sys::fd-stream file))
+
+  (when (cst:atom form)
+    (when (typep (cst:raw form) 'util:literal-value)
+      (return-from parse-pattern
+        (make-pattern-literal
+         :value (cst:raw form)
+         :source (cst:source form))))
+
+    (when (eq (cst:raw form) 'coalton:_)
```
stylewarning, line 75: constant first

src/parser/pattern.lisp
```diff
@@ -0,0 +1,120 @@
...
+   #:pattern-constructor                ; STRUCT
+   #:make-pattern-constructor           ; CONSTRUCTOR
+   #:pattern-constructor-name           ; ACCESSOR
+   #:pattern-constructor-patterns       ; ACCESSOR
+   #:parse-pattern                      ; FUNCTION
+   ))
+
+(in-package #:coalton-impl/parser/pattern)
+
+(defstruct (pattern
```
stylewarning, line 32: after reading a good number of these files, and now onto this one, I have a general feeling that we have a lot of code that's clearly following some sort of "informal specification" that's not written down. If I didn't know Coalton codebase somewhat (and Coalton itself), I would have a hard time navigating everything.

It doesn't help that documentation is incredibly light. There is some helpful comments in long case analyses (e.g., about errors), but there's nothing as it pertains to what anything means, what invariants need to be satisfied, etc. etc.

I don't want to block this PR over it, but we need to spec something out somewhere, whether it's in the headers of these files, or somewhere else.

src/parser/pattern.lisp
```diff
@@ -0,0 +1,120 @@
...
+            (:include pattern)
+            (:copier nil)))
+
+(defstruct (pattern-constructor
+            (:include pattern)
+            (:copier nil))
+  (name     (util:required 'name)     :type identifier   :read-only t)
+  (patterns (util:required 'patterns) :type pattern-list :read-only t))
+
+(defun parse-pattern (form file)
```
stylewarning, line 64: In reading this code, maybe which is just my personal history as a lisper, I'm getting a little mentally exhausted at the "early return"/"nested when"/"declare now set later" style of coding. I think this style of coding is a welcome sign to complex bugs, since it invites extension simply by splicing in new code in the middle of existing code. `PROGN` is the root of all evil, etc.

While I sometimes think sticking to a strict "expression-only" style of coding can lead to obtuse code (especially when you need to do a complex initialization), in this code, it's getting hard to follow what invariants are true when, since what an expression means now depends on following a sequence of statements, as opposed to reading a nested structure.

src/parser/pattern.lisp
```diff
@@ -0,0 +1,120 @@
...
+                 :primary-note "unknown pattern literal")))
+
+  (unless (cst:proper-list-p form)
+    (error 'parse-error
+             :err (coalton-error
+                   form file
+                   :message "Invalid match branch"
+                   :primary-note "unexpected dotted list")))
+
+  (let ((name)
```
stylewarning, line 99: push let down

stylewarning (changes requested):
a more comprehensive review

src/parser/base.lisp
```diff
@@ -0,0 +1,342 @@
...
+  '(and symbol (not boolean) (not keyword)))
+
+(defun identifier-p (x)
+  (typep x 'identifier))
+
+(defun get-line-from-index (file index)
+  "Get the line number corresponding to the character offset INDEX.
+
+Returns (VALUES LINE-NUM LINE-START-INDEX)"
+  (declare (type sb-sys::fd-stream file)
```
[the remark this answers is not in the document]
eliaslfox: I can change this. Regular streams don't always support `file-position`, and the clhs doesn't define a type for "streams that do support file-position."

I think the best way to avoid SBCL specific functionality would be to run CI against another lisp implementation. I'm not sure what that lisp implementation would be though.

src/parser/expression.lisp
```diff
@@ -0,0 +1,1209 @@
...
+   #:node-do                            ; STRUCT
+   #:make-node-do                       ; CONSTRUCTOR
+   #:node-do-nodes                      ; ACCESSOR
+   #:node-do-last-node                  ; ACCESSOR
+   #:parse-expression                   ; FUNCTION
+   #:parse-body                         ; FUNCTION
+   #:parse-variable                     ; FUNCTION
+   ))
+
+;; TODO: declare in let expressions
```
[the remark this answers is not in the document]
eliaslfox: This is done. I forgot the TODO.

src/parser/expression.lisp
```diff
@@ -0,0 +1,1209 @@
...
+                 :message "Malformed expression"
+                 :primary-note "unexpected dotted list")))
+
+  ;;
+  ;; Keywords
+  ;;
+
+  (when (cst:atom (cst:first form))
+    (case (cst:raw (cst:first form))
+      (coalton:fn
```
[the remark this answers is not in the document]
eliaslfox: Thanks. I didn't realize that was an option.

src/parser/base.lisp
```diff
@@ -0,0 +1,344 @@
...
+
+                        (when (/= end-line start-line)
+                          (setf contains-multiline-note t))
+
+                        (list note start-line start-offset end-line end-offset))))))
+              notes-with-line-info))
+
+           ;; Get the last line mentioned
+           (line-number-width
+             (1+ (floor (log (fourth (car (last notes-with-line-info))) 10)))))
```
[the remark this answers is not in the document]
gefjon: this feels like it should be a named function

src/parser/base.lisp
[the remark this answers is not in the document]
gefjon: Pretty much the only option IMO is CCL unless you can convince Franz to get your hands on an Allegro CI setup.

src/parser/expression.lisp
```diff
@@ -0,0 +1,1209 @@
...
+                   :message "when parsing do expression")))))
+
+  (parse-expression form file))
+
+(defun parse-let-declare (form file)
+  (declare (type cst:cst form)
+           (type sb-sys::fd-stream file)
+           (values node-let-declare))
+
+  (assert (cst:consp form))
```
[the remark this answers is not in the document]
eliaslfox: This is correct. The assert is to indicate that the invariant was previously checked an a user friendly error would have been signaled already.

src/parser/parser.lisp
gefjon: Or `string-equalp`, depending on if you want case-sensitivity.

src/parser/parser.lisp
gefjon: or just `when attributes`

src/parser/parser.lisp
eliaslfox: no

src/parser/parser.lisp
eliaslfox: no

src/parser/parser.lisp
eliaslfox: endp doesn't work on vectors

gefjon:
> As for coding style, the decisions made in the parser are fairly intentional and I would prefer not to change them. Writing code in a more expression based style with cond and multiple-value-bind would make functions denser and push code further to the right. Also most predeclared variables are later assigned to struct slots with declared types which ensures their eventual correct definition.

> For the smaller stuff, let* is preferred over let because I have never wanted let when defining multiple variables. (let ((x)) ...) is preferred over (let (x) ...) because it keeps variable definitions consistent. Both of these match the style used throughout the rest of the codebase.

I agree with some of these, and the ones I disagree with I'm willing to accept as well-thought-out preference, but I think @stylewarning is right that it would be good to have a LISP-STYLE.md or something somewhere in the project that lays out the rationale and the fact that these are intentional decisions. But that should probably be a separate PR, I suppose. I'll open an issue.

stylewarning:
> > As for coding style, the decisions made in the parser are fairly intentional and I would prefer not to change them. Writing code in a more expression based style with cond and multiple-value-bind would make functions denser and push code further to the right. Also most predeclared variables are later assigned to struct slots with declared types which ensures their eventual correct definition.
> 
> 
> 
> > For the smaller stuff, let* is preferred over let because I have never wanted let when defining multiple variables. (let ((x)) ...) is preferred over (let (x) ...) because it keeps variable definitions consistent. Both of these match the style used throughout the rest of the codebase.
> 
> 
> 
> I agree with some of these, and the ones I disagree with I'm willing to accept as well-thought-out preference, but I think @stylewarning is right that it would be good to have a LISP-STYLE.md or something somewhere in the project that lays out the rationale and the fact that these are intentional decisions. But that should probably be a separate PR, I suppose. I'll open an issue.

I will not be agreeable to large deviations from the style of Lisp written for the past many years from dozens of people within the Quil, Coalton, and other adjacent projects, which also deviate from every textbook written in Lisp, which also deviate from many open-source Lisp libraries. How Coalton compiler code has been written over the past year or so is a valid precedent to consider, but it wasn't a result of any principled discussion among developers (though it may have been principled in the heads of individual developers who wrote it).

Especially when a project is getting off the ground, I'm sympathetic to not letting stylistic concerns retard development velocity, as long as architectural concerns are, from a global point-of-view, in decent standing. Despite this, much to my chagrin, even my small one-character requests to keep coding style consistent have been insouciantly sidelined with little to offer in return as justification besides "I don't like it" or "I don't see the point".

Things are getting more serious, not the least of which because Coalton is depended upon in critical projects, and so I take a bit more of a serious stance in coming to an agreement for how we write code.

Pushing code to the right, or code density (at least in this context), are not without merit, but haven't hitherto been accompanied by anything objective or definitive about why it's better or worse than the idiomatic alternative. I offered a concrete reason: using expression-oriented programming as a matter of style, and explicitly delineating non-expression-oriented code through a series of stylistic rules, means that a programmer can more readily rely on assumptions about dependence, control flow, context, etc. when both understanding the code and making refactoring decisions. More specifically, nesting offers a syntactic notion of dependence, whereas concatenation merely implies it.

While there may be said merit in other styles of programming, there's value in following idioms, especially in Lisp code that is ultimately driven by an evolving team of people over years or even decades, *especially* in a language like Common Lisp—a language so frequently (and in my opinion, improperly) regarded as a language with too much flexibility invariably leading to everybody adopting their own incompatible pet styles.

If a `LISP-STYLE.md` is what's needed to rectify this and ensure agreed-upon consistency, then so be it; that has value of being unambiguous. With that said, I'm not convinced relatively routine code of this PR's nature warrants special treatment or a re-philosophization of Lisp style.

This note shouldn't be construed as an unquestionable demand to change something (especially if we now consider such change to be blocked by an agreed style guide), but it is a request for us to broaden our view of who stakeholders are/will be and consequently how decisions affect those people.

eliaslfox:
Closed in favor of #810.

PR #779, opened 2022-11-15
Add custom hash table

eliaslfox:
Add a fallback hash table for use on lisp implementations that don't support creating standard hash tables with custom hash functions.


src/hashtable-shim/hash-table.lisp
```diff
@@ -0,0 +1,321 @@
...
+(defun hash-table-insert (table key element)
+  "Insert or update the values stored at KEY in TABLE."
+  (declare (type hash-table table)
+           (type t key)
+           (type t element)
+           (values null)
+           #.*hash-table-optimize*)
+
+  (when (>= (the fixnum (* (hash-table-size table) 10))
+            (the fixnum (* (the fixnum (hash-table-capacity table)) 7)))
```
macrologist, line 180: Are you comfortable with these magic numbers?

src/hashtable-shim/hash-table.lisp
```diff
@@ -0,0 +1,321 @@
...
+  (let* ((values (hash-table-values table))
+
+         (keys (hash-table-keys table))
+
+         (metadata (hash-table-metadata table))
+
+         (old-capacity (the fixnum (hash-table-capacity table)))
+
+         ;; Resize to double the previous capacity or 7, whichever is larger.
+         (new-capacity (* 2 old-capacity)))
```
macrologist, line 210: Stale comment?

macrologist (commented):
Let me know if you were looking for more specific kinds of commentary.

macrologist (approved):
Other than not "really" being swiss tables (no simd utilization), the implementation looks great! 

One place where, perhaps in the future, you might find room to optimize would be to trigger resize on deletion.  If, for example, this table were backing active client sessions on some web service, and the service has one big day where a million people join and use the service, but then they don't use it again basically ever, your table will be slow for life.

A sane scheme might be to keep a slot that tracks number of deletions since last resize. When deletions hit a threshold (as a fraction of size), trigger a resize. Of course, this would only be of interest in circumstances with that above-described peculiar usage: where large numbers of deletions occur within a short time frame (say all on the same day after people try the web service) but where such periods of "mass deletion" occur fairly infrequently (maybe you only get one "big day" every year or so). 

So I doubt much that such a change is predictably worth it. But in situations in which it was worth it, it would save time during lookup. 

Just a thought.

src/hashtable-shim/hash-table.lisp
eliaslfox: Added a comment

colescott (approved):

src/hashtable-shim/hash-table.lisp
```diff
@@ -0,0 +1,321 @@
...
+             key
+             short-hash
+             i
+             (hash-table-eq-function table))
+        ;; Mark slot as deleted
+        (setf (aref (hash-table-metadata table) i) +slot-deleted+)
+
+        ;; Clear key and element slots to avoid pinning objects
+        (setf (aref (hash-table-keys table) i) nil)
+        (setf (aref (hash-table-values table) i) nil)
```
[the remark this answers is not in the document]
eliaslfox: I do find it easier to read. Other programming languages I've used don't have an equivalent multi-assignment form, so one assignment per line feels a little more familiar.

eliaslfox:
@macrologist thanks for the review. I'm sorry it took me so long to get back to you.

The resizing on deletion is an interesting point. Generally when using resizing data structures I would expect them to never shrink because resizing and copying elements tends to be the most expensive operation. Some other languages have functions for explicitly resizing to a smaller collection, but I don't think there's an equivalent for lisp hash tables. I'll have to think about how Coalton can support that use case. In the mean time, I think bursty workloads will need to allocate a hash table per request to allow reclaiming memory.

PR #786, opened 2022-12-16
Add more docstrings and other minor fixes

eliaslfox:


eliaslfox:
I know `cl:char-code` isn't guaranted by the standard to use ASCII. If we need to support a lisp implementation that doesn't use ASCII we can vendor an ascii table.

library/vector.lisp
```diff
@@ -197,7 +197,7 @@
 
   (declare foreach2 ((:a -> :b -> :c) -> Vector :a -> Vector :b -> Unit))
   (define (foreach2 f v1 v2)
-    "Like vector-foreach but twice as good"
+    "Iterate in parallel over V1 and V2 calling F once for each pair of elements. Iteration stops when the first vector ends"
```
colescott, line 200: Should this be "Iteration stops when the shorter of the two vectors runs out of elements" or something to that effect?

library/char.lisp
```diff
@@ -16,16 +27,19 @@
 (coalton-toplevel
   (declare char-code (Char -> UFix))
   (define (char-code char)
+    "Convert a character to it's ASCII representation."
```
colescott, line 30: `it's` -> `its`

Same below.

colescott (approved):

PR #795, opened 2023-01-18
Add pattern exhaustiveness and usefulness warnings

### LINKED ISSUE #8: Case Exhaustiveness Checking

colescott:
This adds pattern exhaustiveness and usefulness warnings to Coalton `match` forms.

Implementation based on http://moscova.inria.fr/~maranget/papers/warn/index.html

Resolves #8

### DIFF
--- a/library/ord-map.lisp
+++ b/library/ord-map.lisp
@@ -174,7 +174,9 @@ If ITER contains duplicate keys, later values will overwrite earlier values."
                                                       (helper less)))
                    ((Eq) (Some (tree::Branch clr less (MapPair pivot (func val)) more)))
                    ((GT) (coalton-library/classes:map (tree::Branch clr less (MapPair pivot val))
-                                                      (helper more)))))))))
+                                                      (helper more)))))
+                ((tree::Branch _ _ (JustKey _) _) (error "Map contains `JustKey` rather than `MapPair`"))
+                ((tree::DoubleBlackEmpty) (error "Encountered double-black node in ordered map outside of removal operation"))))))
       (coalton-library/classes:map %Map
                                    (helper tre))))
 
@@ -203,7 +205,9 @@ operation, and therefore Map cannot implement Monoid."
                          (tree::Branch clr
                                        (helper left)
                                        (MapPair key (func value))
-                                       (helper right)))))))
+                                       (helper right)))
+                        ((tree::Branch _ _ (JustKey _) _) (error "Map contains `JustKey` rather than `MapPair`"))
+                        ((tree::DoubleBlackEmpty) (error "Encountered double-black node in ordered map outside of removal operation"))))))
         (%Map (helper tre)))))
 
   ;; As with `tree:Tree', `Map' should probably implement `Traversable'.
--- a/library/ord-tree.lisp
+++ b/library/ord-tree.lisp
@@ -125,7 +125,8 @@
        (match (<=> needle elt)
          ((LT) (lookup left needle))
          ((EQ) (Some elt))
-         ((GT) (lookup right needle))))))
+         ((GT) (lookup right needle))))
+      ((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))))
 
   ;;; inserting into and replacing elements of trees
 
@@ -278,7 +279,8 @@
                                  (ins left)))
                       ((EQ) None)
                       ((GT) (map (fn (new-right) (balance clr left pivot new-right))
-                                 (ins right)))))))))
+                                 (ins right)))))
+                   ((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))))))
       (ins tre)))
 
   (declare replace ((Ord :elt) => ((Tree :elt) -> :elt -> (Optional (Tuple (Tree :elt) :elt)))))
@@ -298,7 +300,8 @@ If TRE did not have an element `==' to ELT, return None."
                       ((EQ) (Some (Tuple (Branch clr left elt right)
                                          pivot)))
                       ((GT) (map (uncurry (fn (new-right removed-elt) (Tuple (balance clr left pivot new-right) removed-elt)))
-                                 (ins right)))))))))
+                                 (ins right)))))
+                   ((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))))))
       (ins tre)))
 
   (declare replace-or-insert ((Ord :elt) => ((Tree :elt) -> :elt -> (Tuple (Tree :elt) (Optional :elt)))))
@@ -322,7 +325,8 @@ alongside the new tree."
                       ((GT)
                        (let (Tuple new-right removed-elt) = (ins right))
                        (Tuple (balance clr left pivot new-right)
-                              removed-elt))))))))
+                              removed-elt))))
+                   ((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))))))
       (let (Tuple new-tree removed-elt) = (ins tre))
       (Tuple (as-black new-tree) removed-elt)))
 
@@ -357,7 +361,8 @@ Like `replace-or-insert', but prioritizing insertion as a use case."
          ((EQ) (Some (remove-leaving-double-black tre)))
          ((GT)
           (map (fn (new-right) (bubble-double-black clr left pivot new-right))
-               (remove-without-as-black right elt)))))))
+               (remove-without-as-black right elt)))))
+      ((DoubleBlackEmpty) (error "Encountered double-black node early in `remove` while searching for the node to remove."))))
 
   ;; matt might calls this operation `remove'
   (declare remove-leaving-double-black (((Tree :elt) -> (Tree :elt))))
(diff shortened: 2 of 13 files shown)


gefjon:
I wonder if we should offer `ematch` (name pending) to encode the common pattern of `(match foo ... (_ (error "Unexpected shape")))`. I notice this diff adds quite a few patterns like that.

colescott:
> I wonder if we should offer `ematch` (name pending) to encode the common pattern of `(match foo ... (_ (error "Unexpected shape")))`. I notice this diff adds quite a few patterns like that.

My personal feeling is that any catch-all patterns should have an application-specific error message. I didn't know enough about the ord-tree and ord-map to make useful messages, and I didn't super care for THIH but those can definitely be fixed in the future.

library/ord-tree.lisp
```diff
@@ -125,7 +125,9 @@
        (match (<=> needle elt)
          ((LT) (lookup left needle))
          ((EQ) (Some elt))
-         ((GT) (lookup right needle))))))
+         ((GT) (lookup right needle))))
+      (_
```
gefjon, line 129: I would replace this catch-all clause with `((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))`. One of the invariants of the data structure is that double-black nodes only exist within the removal algorithm, and are cleaned up before a `remove` (or similar) returns.

library/ord-tree.lisp
```diff
@@ -278,7 +280,9 @@
                                  (ins left)))
                       ((EQ) None)
                       ((GT) (map (fn (new-right) (balance clr left pivot new-right))
-                                 (ins right)))))))))
+                                 (ins right)))))
+                   (_
```
gefjon, line 284: Same as previous; change to `((DoubleBlackEmpty) (error "Found double-black node outside of removal process"))`, modulo wording on the error message.

library/ord-tree.lisp
```diff
@@ -298,7 +302,9 @@ If TRE did not have an element `==' to ELT, return None."
                       ((EQ) (Some (Tuple (Branch clr left elt right)
                                          pivot)))
                       ((GT) (map (uncurry (fn (new-right removed-elt) (Tuple (balance clr left pivot new-right) removed-elt)))
-                                 (ins right)))))))))
+                                 (ins right)))))
+                   (_
```
gefjon, line 306: Ditto.

library/ord-tree.lisp
```diff
@@ -322,7 +328,9 @@ alongside the new tree."
                       ((GT)
                        (let (Tuple new-right removed-elt) = (ins right))
                        (Tuple (balance clr left pivot new-right)
-                              removed-elt))))))))
+                              removed-elt))))
+                   (_
```
gefjon, line 332: Another `DoubleBlackEmpty` unreachable.

library/ord-tree.lisp
```diff
@@ -357,7 +365,9 @@ Like `replace-or-insert', but prioritizing insertion as a use case."
          ((EQ) (Some (remove-leaving-double-black tre)))
          ((GT)
           (map (fn (new-right) (bubble-double-black clr left pivot new-right))
-               (remove-without-as-black right elt)))))))
+               (remove-without-as-black right elt)))))
+      (_
```
gefjon, line 369: Another place where `DoubleBlackEmpty` should be impossible, but this time we are technically inside the `remove` process, we just haven't reached a point where we would insert a double-black node yet. Possibly `(error "Encountered double-black node early in `remove` while searching for the node to remove.")`.

library/ord-tree.lisp
```diff
@@ -390,7 +400,9 @@ The result tree may be in an intermediate state with a double-black node."
       ;; nodes with two subtrees
       ((Branch clr left _ right)
        (let (Tuple new-left new-pivot) = (remove-max left))
-       (bubble-double-black clr new-left new-pivot right))))
+       (bubble-double-black clr new-left new-pivot right))
+      (_
```
gefjon, line 404: Same as previous; `DoubleBlackEmpty` is not yet possible because we haven't inserted a double-black node yet.

library/ord-tree.lisp
```diff
@@ -407,7 +419,9 @@ The result tree may be in an intermediate state with a double-black node."
       ((Branch clr left pivot right)
        (let (Tuple new-right removed-max) = (remove-max right))
        (Tuple (bubble-double-black clr left pivot new-right)
-              removed-max))))
+              removed-max))
+      (_
```
gefjon, line 423: Same as previous two.

library/ord-tree.lisp
```diff
@@ -436,7 +450,9 @@ The result tree may be in an intermediate state with a double-black node."
        (cell:push! stack (Expand more))
        (cell:push! stack (Yield elt))
        (cell:push! stack (Expand less))
-       Unit)))
+       Unit)
+      (_
```
gefjon, line 454: Back to the first one; this is outside of `remove`, so `DoubleBlackEmpty` should never be present.

library/ord-tree.lisp
```diff
@@ -451,7 +467,9 @@ The result tree may be in an intermediate state with a double-black node."
        (cell:push! stack (Expand less))
        (cell:push! stack (Yield elt))
        (cell:push! stack (Expand more))
-       Unit)))
+       Unit)
+      (_
```
gefjon, line 471: Ditto.

library/ord-map.lisp
```diff
@@ -174,7 +174,9 @@ If ITER contains duplicate keys, later values will overwrite earlier values."
                                                       (helper less)))
                    ((Eq) (Some (tree::Branch clr less (MapPair pivot (func val)) more)))
                    ((GT) (coalton-library/classes:map (tree::Branch clr less (MapPair pivot val))
-                                                      (helper more)))))))))
+                                                      (helper more)))))
+                (_
```
gefjon, line 178: I would change to `((tree::DoubleBlackEmpty) (error "Encountered double-black node in ordered map outside of removal operation"))`.

library/ord-map.lisp
```diff
@@ -203,7 +205,9 @@ operation, and therefore Map cannot implement Monoid."
                          (tree::Branch clr
                                        (helper left)
                                        (MapPair key (func value))
-                                       (helper right)))))))
+                                       (helper right)))
+                        (_
```
gefjon, line 209: Ditto.

tests/red-black-tests.lisp
```diff
@@ -36,9 +36,11 @@
           (if (== left-ct right-ct)
               (Ok (+ left-ct (match c
                                ((red-black/tree::Black) 1)
-                               ((red-black/tree::Red) 0))))
+                               ((red-black/tree::Red) 0)
+                               (_ (error "unreachable")))))
               (Err (DifferentCountToBlack left-ct left
-                                          right-ct right)))))))))
+                                          right-ct right))))))
+      (_ (error "unreachable")))))
```
gefjon, line 43: These two should probably have a dedicated variant in `InvariantError`. Something like:

```
(define-type (InvariantError :elt)
  ;;  other variants
  (IllegalColor (red-black/tree:Tree :elt)))
```

And then have the two branches, instead of calling `error`, be:

```
(if (== (left-ct right-ct))
  (match c
    ((red-black/tree::Black) (Ok (1+ left-ct)))
    ((red-black/tree::Red) (Ok left-ct))
    (_ (Err (IllegalColor tre))))
  (Err (DifferentCountToBlack left-ct left right-ct right)))
```

```
((DoubleBlackEmpty) (Err (IllegalColor tre)))
```

library/ord-tree.lisp
colescott: It looks like this one is also missing `Empty`

library/ord-tree.lisp
colescott: This one is also missing `Empty`

library/ord-tree.lisp
gefjon: Ah, yes, that's unreachable because `remove-leaving-double-black` is called by `remove-without-as-black` only in the case of a non-empty tree. In that case, this should probably be a catch-all pattern like
```
(_ (error "Attempt to `remove-leaving-double-black` on an empty tree"))
```

library/ord-tree.lisp
gefjon: This function currently is not exported and is only called by `remove-leaving-double-black` on non-empty trees, so it could have `(_ (error "Attempt to `remove-max` from an empty tree"))` like previous, but we might want to export this operation, in which case it should return an `Optional (Tuple (Tree :elt) :elt)`, return `None` in the `(Empty)` case, and error in the `(DoubleBlackEmpty)` case. But that change seems outside the scope of this PR, so I suggest the catch-all error.

gefjon:
> ;   Match patterns are non-exhaustive. Missing pattern (BRANCH _ _ (JUSTKEY _) _)
> ;   In match on SUBTREE

These are unreachable because the `JustKey` variant of `MapPair` is never inserted into a map. From the source:
```
;; hack: when searching for an entry, we need a thing that holds just a key, but is of type `MapPair :key
;; :any', to pass to `tree:lookup' or `tree:remove'. A `Map' will never hold a `JustKey'.
```

For these, I would use the match branch
```
((Branch _ _ (JustKey _) _) (error "Map contains `JustKey` rather than `MapPair`"))
```

eliaslfox (approved):

PR #801, opened 2023-01-19
Add `IntoIterator` instances for Optional and Result

fosskers:
While nearly trivial, these are still quite reasonable to have, and are added in the spirit of "if a lawful instance is possible, then write it".

Let me know if you'd like tests written.

### DIFF
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -147,6 +147,28 @@ Behavior is undefined if the iterator is advanced after a destructive modificati
   (define-instance (IntoIterator (Iterator :elt) :elt)
     (define into-iter id))
 
+  (define-instance (IntoIterator (Optional :elt) :elt)
+    (define (into-iter opt)
+      "Yield the single value of this OPT, or nothing."
+      (match opt
+        ((Some a) (let ((cell (cell:new True)))
+                    (%Iterator (fn () (if (cell:read cell)
+                                          (progn (cell:write! cell False)
+                                                 (Some a))
+                                          None)))))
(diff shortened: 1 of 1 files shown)


library/iterator.lisp
```diff
@@ -147,6 +147,22 @@ Behavior is undefined if the iterator is advanced after a destructive modificati
   (define-instance (IntoIterator (Iterator :elt) :elt)
     (define into-iter id))
 
+  (define-instance (IntoIterator (Optional :elt) :elt)
+    (define (into-iter opt)
+      "Yield the single value of this OPT, or nothing."
+      (match opt
+        ((Some a) (let ((cell (cell:new (make-list a))))
+                    (%Iterator (fn () (cell:pop! cell)))))
```
fosskers, line 155: I borrowed the `cell` technique from the `List` instance of this typeclass.

eliaslfox:
I don't think this needs to allocate a list. It should be possible to store a boolean flag in a cell instead.

fosskers:
Let me try.

gefjon:
It should probably look like
```
(define-instance (IntoIterator (Optional :elt) :elt)
  (define (into-iter opt)
    (match opt
      ((None) empty)
      ((Some elt)
       (let done? = (cell:new False))
       (new (fn ()
              (if (cell:read done?)
                  None
                  (Some elt))))))))
```

eliaslfox (approved):
LGTM

fosskers:
Updated. It's what @gefjon suggested just in a different shape.

PR #806, opened 2023-01-24
Add `Bifunctor` and two major instances

fosskers:
This PR implements the [Bifunctor](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Bifunctor.html#t:Bifunctor), nearly as-is from Haskell.

Note that for the moment, since Coalton requires that all typeclass functions be manually implemented, only `bimap` is demanded here. `first` and `second` have default implementations based on `bimap` that can't be overridden.


library/result.lisp
```diff
@@ -102,6 +102,12 @@
         ((Ok x) (f x))
         ((Err e) (Err e)))))
 
+  (define-instance (Bifunctor Result)
+    (define (bimap f g res)
+      (match res
+        ((Ok x) (Ok (g x)))
+        ((Err e) (Err (f e))))))
```
fosskers, line 109: Took me a second to realise that while the `Result` name is the same as Rust, the type param order is the same as `Haskell` and that `:b` is the happy value :laughing:

gefjon:
Why require `bimap` and define `first` and `second`, instead of the other way around? I don't object to the choice, particularly, I just wonder if there's a reason to prefer that construction.

library/tuple.lisp
```diff
@@ -69,6 +69,12 @@
           (hash a)
           (hash b))))))
 
+  (define-instance (Bifunctor Tuple)
+    (define (bimap f g item)
+      (match item
+        ((Tuple a b)
+         (Tuple (f a) (g b))))))
```
fosskers, line 76: Haskell also provides instances for all the bigger `Tuple*` types. Should I do that here as well?

fosskers:
> Why require bimap and define first and second, instead of the other way around?

A complete whim. I'm happy to do it the other way around.

eliaslfox:
Defining bimap directly should be more efficient than "deriving" it from calling first and then second on large structures.

library/tuple.lisp
eliaslfox: Might as well just for completeness.

gefjon:
> Defining bimap directly should be more efficient than "deriving" it from calling first and then second on large structures.

Ah yes, that makes sense.

library/classes.lisp
```diff
@@ -267,6 +268,20 @@ specify `repr :lisp`."
...
   (define sequence (traverse (fn (x) x)))
 
+  (define-class (Bifunctor :f)
+    "Types which take two type arguments and are functors on both."
+    (bimap ((:a -> :b) -> (:c -> :d) -> (:f :a :c) -> (:f :b :d))))
+
+  (declare first ((Bifunctor :f) => (:a -> :b) -> (:f :a :c) -> (:f :b :c)))
+  (define (first f b)
+    "Map over the first argument of a Bifunctor."
+    (bimap f (fn (x) x) b))
```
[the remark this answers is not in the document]
eliaslfox: That's fine. I don't think it's importable without introducing a circular dependency.

library/tuple.lisp
gefjon: I don't love that, because e.g. for `Tuple3`, the instance will be `Bifunctor (Tuple3 :a)`, right? And `first` will operate on the second element of the tuple, and `second` on the third? Feels a little bit wrong.

eliaslfox:
I think functions should also have a bifunctor instance. In Coalton unapplied function types can be written like:

```
(define-instance (Bifunctor Arrow)
  ...)
```

fosskers:
@eliaslfox Oh, that's quite (conceptually) interesting. What's the use-case? Funky function composition?
Also, would that map (logically) to any of the instances listed here? https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Bifunctor.html#t:Bifunctor

gefjon:
> I think functions should also have a bifunctor instance.

How would that work? Wouldn't you need one end to be contravariant?

eliaslfox:
It looks that actually can't work. Haskell defines a similar instance on [Protofunctor](https://www.stackage.org/haddock/lts-14.20/profunctors-5.3/Data-Profunctor.html#t:Profunctor) where one side is covariant and the other is contravariant.

fosskers:
Another topic from Discord: are `map-first` and `map-second` better names?

gefjon:
> fosskers: In that case, since fst and snd already exist for Tuples, do you want map-fst and map-snd?

I find this aesthetically pleasing.

fosskers:
Let me know if you'd like unit tests.

gefjon:
I never turn down a unit test.

library/result.lisp
fosskers: Yeah re: `Functor (Result :a)`

library/tuple.lisp
fosskers: I'd tend to agree with @gefjon . And at that point, shouldn't the user be making a custom type anyway? Why are they fiddling around with length-N tuples? I've never personally used larger than a size 3 or 4, and even then I never needed to map across it.

stylewarning (approved):
lgtm

Issue #778, opened 2022-11-11
Support other lisp implementations


chrisdone-artificial:
> While we try to be ANSI-conforming, Coalton may only work on SBCL 2.1.x.

If I were to try to use Coalton on LispWorks or ECL, for example, how much trouble would I encounter? Alternatively, how much work would it be to make it just depend on the CL standard?

My other car is a cdr @chrisdone

stylewarning:
Right now there are some SBCL-specific things we rely on. There's not much (e.g., custom hash tables); most of the code is ANSI enough.

We are actually currently making a push to identify all of those areas and rectify them.

Our ultimate goal is to have CI testing against at least one other Lisp besides SBCL.

chrisdone:
Neat, thanks!

eliaslfox:
As of #780 Coalton now builds on Allegro. @colescott and I attempted to get Coalton running on other lisp implementations and we ran into the following issues:

* ECL: Attempting to compile [library/types.lisp](https://github.com/coalton-lang/coalton/blob/main/library/types.lisp) caused a segfault.

* Clasp: The compiler errors at runtime, the issue appears to be a miscompile of a `(declare (values ...))` annotation. Values declarations are not standard lisp and Clasp seems to ignore them by default. However, in at least one case it is compiled to `(typep x 'values)` which errors at runtime.

* Clisp doesn't support package local nicknames. There is a pr, but it's unmerged https://gitlab.com/gnu-clisp/clisp/-/merge_requests/3
 
* CCL appears to use fpu traps internally which breaks math on NaN values.

* ABCL compiles [library/types.lisp](https://github.com/coalton-lang/coalton/blob/main/library/types.lisp), but then it appears to generate an invalid fasl file which it fails to load.

I haven't had a chance to try LispWorks yet.

gmpalter:
You can use `(ccl:set-fpu-mode :overflow nil :underflow nil :division-by-zero nil :invalid nil :inexact nil)` to disable traps

eliaslfox:
@gmpalter masking float traps works, thank you.

eliaslfox:
Closing as CCL and Allegro are now running in CI.

PR #817, opened 2023-02-10
Add CCL support

eliaslfox:
The tests run locally in ccl. CI support isn't added yet. Masking float traps is required as mentioned in https://github.com/coalton-lang/coalton/issues/778#issuecomment-1422675106.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,13 +15,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        LISP_IMPL: ["sbcl", "allegro"]
+        LISP_IMPL: ["sbcl", "allegro", "ccl"]
         COALTON_ENV: ["release", "development"]
         SAFETY: ["0", "3"]
         exclude:
-          # Since we don't have a way of switching safety on allegro, we can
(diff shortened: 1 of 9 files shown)


gmpalter:
Just thought I'd mention that, in CCL, the form `(pathname-type *.fasl-pathname*)` will give you the correct file type for whatever hardware platform CCL is running on.

eliaslfox:
I ran into two issues with NaN values in CCL.
* https://github.com/Clozure/ccl/issues/440
* https://github.com/Clozure/ccl/issues/441

Neither is technically blocking as Coalton's tests run on CCL currently. The first one we could correct within Coalton by explicitly checking for NaN values in Coalton's comparison functions. The second one can be avoided by disabling float traps.

eliaslfox:
CI is failing because Github Actions can't pull a docker container. Pulling locally works, so I think the runner cached a permission failure.

colescott:
The Github actions permission issue seems to have solved itself. Now it looks like we just need to disable floating point traps either by default in Coalton or add that to the tests.

Additionally, it looks like CCL is not returning a non-zero on failure. I don't see anything in the docs about this but I suspect using `--batch` might get the behavior we want.

colescott (approved):

PR #829, opened 2023-03-13
Fix support for progn in coalton toplevel

colescott:


gefjon:
Seems like we should add a regression test for this. Even something trivial, like
```
(coalton-toplevel
  (progn
    (declare foo String)
    (define foo "foo")))
```
Not having looked at the test infrastructure for a while, I'm not sure if there's a good place to put this. Could it be a case in `tests/parser`?

PR #837, opened 2023-03-16
Fix parsing let expressions without bindings

### LINKED ISSUE #835: `let cl:nil ...` should be allowed
I have a macro which, in some invocations, returns a `let` form with no bindings. This seems to get treated as being like `(let cl:nil ...)`, and rejected by the `cst:consp` test on line 573 of src/parser/expression.lisp.

eliaslfox:
Fixes #835

### DIFF
--- a/src/parser/expression.lisp
+++ b/src/parser/expression.lisp
@@ -569,8 +569,7 @@
                     :message "Malformed let"
                     :primary-note "expected let body")))
 
-     ;; (let x ...)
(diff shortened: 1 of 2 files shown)


gefjon:
Does removing the `bad.coalton` test for
```
(define f (let ()))
```
mean that form is now allowed?

eliaslfox:
Oops lol.

Tests are fixed now. That form isn't allowed.

gefjon (approved):

PR #838, opened 2023-03-17
Macro expansion limit

### LINKED ISSUE #257: Possible infinite macro-expansion
Function `parse-application` in `src/ast/parse-form.lisp` calls `macro-function` on a symbol that is the head of a Coalton function application and applies the macroexpansion. I spotted this case out of habit of writing `(lambda ...)` instead of `(fn ...)`, where the form expands as `(function (lambda ...))`, which results in infinite recursion.

This is not really a bug since `lambda` is not part of the Coalton language, but on the other hand there should ideally be a better error  message than reaching Control Stack Exhausted.

There could be a general mechanism to guard against infinite recursions (e.g. limit depth (simple), or check if there is a repetitive pattern in the expansion (harder)) 

Also, macroexpansion is done for arbitrary symbols, maybe symbols from the CL package should not be expanded (they tend to produce non-Coalton code anyway). Or only symbols marked as being Coalton macros should be expanded.

I am opening this issue to discuss possible options before suggestion a pull request.

eliaslfox:
Error after 500 recursive macro expansions. This generates a nice error than "stack exausted."

Fixes #257

### DIFF
--- a/src/parser/expression.lisp
+++ b/src/parser/expression.lisp
@@ -135,6 +135,10 @@
 
 (in-package #:coalton-impl/parser/expression)
 
+(defvar *macro-expansion-count* 0)
+
+(defconstant +macro-expansion-max+ 500)
+
 ;;;; # Expression Parsing
 ;;;;
 ;;;; Note that "expression" in the EBNF corresponds to the struct "node" in the lisp code.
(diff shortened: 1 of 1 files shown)


colescott (approved):

src/parser/expression.lisp
```diff
@@ -864,12 +868,23 @@
...
-     (let ((*coalton-error-context*
-             (adjoin (make-coalton-error-context
-                      :message "Error occurs within macro context. Source locations may be imprecise")
-                     *coalton-error-context*
-                     :test #'equalp)))
-       (parse-expression (expand-macro form) file)))
+
+     (let ((*macro-expansion-count* (+ 1 *macro-expansion-count*)))
+
+       (when (= *macro-expansion-count* +macro-expansion-max+)
```
stylewarning, line 874: it would have been cool to make this a continuable error with CERROR

PR #839, opened 2023-03-17
Remove usage of with-float-traps-masked

### LINKED ISSUE #816: Investigate `float-features:with-float-traps-masked` usage
This macro is used a lot in [library/math/elementary.lisp](https://github.com/coalton-lang/coalton/blob/main/library/math/elementary.lisp). Investigate if this is required.

eliaslfox:
Fixes #816

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -133,7 +133,7 @@
                #:alexandria)
   :pathname "library/"
   :serial t
-  :components ((:file "set-float-traps" :if-feature :ccl)
+  :components ((:file "set-float-traps")
                (:file "utils")
                (:file "types")
                (:file "fixed-size-numbers")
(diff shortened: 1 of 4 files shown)


colescott (approved):

stylewarning:
*the price is right sad sound*

What's the status of tests locally?

eliaslfox:
The tests pass with float traps disabled. 

By default SBCL enabled some float traps. I tried to see if there was a commit explaining why, but it seems that the default float trap settings were inherited from CMUCL (see [src/code/cold-init.lisp](https://github.com/sbcl/sbcl/blob/a530bbe337109d898d5b4a001fc8f1afa3b5dc39/src/code/cold-init.lisp#L216-L219)).

SBCL's default float trap settings are in contrast to other environments like C which runs with float traps disabled, and Java which doesn't even support float traps.

We already had to disable float traps on CCL, this pr also disables them on SBCL.

stylewarning:
@eliaslfox what about the CCL unit test?

eliaslfox:
GitHub actions is broken. I'll restart it.

PR #840, opened 2023-03-20
[WIP] Basic loops

simendsjo:
Not sure how we should implement basic loops in an efficient way. This implementation requires tail-recursion, which was only available when optimizing for speed on my machine. And I guess the lambdas `condition` and `body` won't be inlined which causes unnecessary function call overhead.

This PR is only to trigger discussion as it's easily buried a bit in Discord. Ref https://discord.com/channels/888196168067199046/888196168067199049/1084182802104340590

### DIFF
--- a/library/iterator.lisp
+++ b/library/iterator.lisp
@@ -23,6 +23,9 @@
    #:coalton-library/hashtable
    #:Hashtable)
   (:export
+   #:while
+   #:until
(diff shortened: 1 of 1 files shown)


eliaslfox:
Moved to #847. I'm in favor of something like this, but I want to have a more holistic design before settling on an implementation.

PR #844, opened 2023-03-22
Print readably when *print-readably* is set

### LINKED ISSUE #707: Printed representation of Coalton types is not readable
The printed representation of Coalton types uses the `#.` macro in an attempt to be readable, but does not quote subobjects. This works fine for subobjects which are also Coalton types, or self-evaluating CL types, but breaks with lists (and would probably break with `repr :native symbol` objects).

E.g.

```
(coalton-toplevel
  (define-type ContainsList
    (ContainsList (List UFix))))

(cl:prin1-to-string (coalton (ContainsList (make-list 0 1 2))))
; => "#.(CONTAINSLIST (0 1 2))"

(cl:read-from-string (cl:prin1-to-string (coalton (ContainsList (make-list 0 1 2)))))
; in: CONTAINSLIST (0 1 2)
;     (0 1 2)
; 
; caught COMMON-LISP:ERROR:
;   illegal function call
; 
; compilation unit finished
;   caught 1 ERROR condition
```

Easy solution: stick a quote in front of each argument, so that `ContainsList` prints as `#.(CONTAINSLIST '(0 1 2))`.

eliaslfox:
Fixes #707

### DIFF
--- a/src/codegen/codegen-type-definition.lisp
+++ b/src/codegen/codegen-type-definition.lisp
@@ -100,9 +100,18 @@
                            (declare (type stream stream)
                                     (type ,classname self)
                                     (values ,classname))
-                           (format stream "#.(~s~{ ~s~})"
-                                   ',(tc:constructor-entry-name constructor)
-                                   (list ,@(mapcar (lambda (slot) `(slot-value self ',slot)) field-names)))
+                             (format stream "#.(~s" ',(tc:constructor-entry-name constructor))
+                             ,@(loop :for slot :in field-names
+                                     :collect `(if (and *print-readably*
+                                                        (or (listp (slot-value self ',slot))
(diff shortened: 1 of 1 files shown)


colescott (approved):

stylewarning:
If the variable is off, maybe we should do away with #. entirely

eliaslfox:
Yeah we could do that. I kinda like how it looks, but it does add noise.

gefjon:
I believe `print-object` methods which rely on sharp-dot to `*print-readably*` are also supposed to check the value of `*read-eval*` and signal an error if it's false, and in the case of `(not (or *print-readably* *read-eval*))` one should not print a sharp-dot form.

eliaslfox:
Read eval is fixed in #846

PR #854, opened 2023-03-28
Callbacks in foreach functions now return Unit

### LINKED ISSUE #755: `foreach` functions should require their callback to return `Unit`

eliaslfox:
Previously callbacks in forach functions could return an arbitrary value. This made code slightly denser, but adding another parameter to a callback function could cause it to be partially applied instead of evaluating. This would be difficult to debug without a type error.

Fixes #755


stylewarning (approved):
I am a fan of this change.

PR #855, opened 2023-03-28
Document defaulting rules

### LINKED ISSUE #763: Document defaulting rules

eliaslfox:
Fixes #763

### DIFF
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -689,7 +689,10 @@ The following functions all take an optional package parameter.
 
 ## Instance Defaulting
 
-Coalton has a similar [type defaulting system](https://www.haskell.org/onlinereport/decls.html#sect4.3.4) as Haskell. Type defaulting is invoked on implicitly typed definitions and code compiled with the `coalton` macro.
+Coalton has a similar [type defaulting system](https://www.haskell.org/onlinereport/decls.html#sect4.3.4) as Haskell. Type defaulting is invoked on implicitly typed definitions and code compiled with the `coalton` macro. Defaulting is applied to a set of ambigious predicates, with the goal to resolve an ambigious type variable to a valid type. Coalton will only default if one or more of the predicates is a numeric type class (Num, Quantizable, Reciprocable, Complex, Remainder, Integral). Coalton will default an ambgious variable to either Integer, Double-Float, or Single-Float; taking the first type that is valid for all predicates referencing that type variable. Coalton will not default when one or more of the predicates containing an ambigous variable is a multi-parameter type class.
+
+
+Differences from Haskell 98. Haskell would consider `Num (List :a)` to be ambigious, Coalton would default it to `Num Integer`. Haskell would consider (`Num :a` `CustomTypeClass :a`) to be amgbigious, Coalton would default to (`Num Integer` `CustomTypeClass Integer`) assuming `CustomTypeClass Integer` was a valid instance.
 
 ## Functional Dependencies
 
--- a/src/parser/toplevel.lisp
+++ b/src/parser/toplevel.lisp
@@ -384,10 +384,7 @@
            (eclector.readtable:copy-readtable eclector.readtable:*readtable*))
 
          ;; Initial package to read (package) forms into
-         (*package* *package*)
-
-         ;; Read unspecified floats as double floats
-         (*read-default-float-format* 'double-float))
+         (*package* *package*))
 
     ;; Parse package form
     (when (eq :file mode)
--- a/tests/type-inference-tests.lisp
+++ b/tests/type-inference-tests.lisp
@@ -432,7 +432,7 @@
       (let ((y 1))
         (+ 0.5 y)))"
 
-   '("f" . "(:a -> Double-Float)"))
+   '("f" . "(:a -> Single-Float)"))
 
   ;; Check that the monomorphism restriction still applies to defaulted bindings
   (signals tc:tc-error


docs/intro-to-coalton.md
```diff
@@ -190,10 +190,7 @@ Coalton supports a few numeric types. The main ones are `Integer`, `Single-Float
   (define num-df  5.0d0))
 ```
 
-One can leave off the suffix and just write `5.0`, which will be resolved
-depending on the read-time value of `cl:*read-default-float-format*`. That
-variable is set to `cl:single-float` by default, meaning unadorned floats will
-be single-precision.
+Unlike in Common Lisp, float literals without a suffix will always be treated as double floats.
```
stylewarning, line 193: How is this the case? Does the reader ensure this? What if we're not in the readtable?

docs/intro-to-coalton.md
```diff
@@ -689,7 +686,10 @@ The following functions all take an optional package parameter.
...
+Coalton has a similar [type defaulting system](https://www.haskell.org/onlinereport/decls.html#sect4.3.4) as Haskell. Type defaulting is invoked on implicitly typed definitions and code compiled with the `coalton` macro. Defaulting is applied to a set of ambigious predicates, with the goal to resolve an ambigious type variable to a valid type. Coalton will only default if one or more of the predicates is a numeric type class (Num, Quantizable, Reciprocable, Complex, Remainder, Integral). Coalton will default an ambgious variable to either Integer, Double-Float, or Single-Float; taking the first type that is valid for all predicates referencing that type variable. Coalton will not default when one or more of the predicates containing an ambigous variable is a multi-parameter type class.
+
+
+Differences from Haskell 98. Haskell would consider `Num (List :a)` to be ambigious, Coalton would default it to `Num Integer`. Haskell would consider (`Num :a` `CustomTypeClass :a`) to be amgbigious, Coalton would default to (`Num Integer` `CustomTypeClass Integer`) assuming `CustomTypeClass Integer` was a valid instance.
```
stylewarning, line 692: Can you bold the "Differences from Haskell 98" or make it a subsub* section?

stylewarning, line 692: Spelling errors

stylewarning (changes requested):

docs/intro-to-coalton.md
eliaslfox: The read table does ensure this. If we're not in the readtable, then the sexpression is stringified and then reread with the Coalton read table.

stylewarning: This would cause some very non-obvious corner-case errors, where the float is written with enough digits for double precision, it gets read by the lisp reader as single precision (thus losing digits), then gets re-read with double precision (but having only single-precision precision).

eliaslfox: I didn't think about that. 

We could set `*read-default-float-format*` to `double` before stringifying the s-expression, then single floats would be printed with an exponent marker and re-read as such.. Or we could go back to not modifying `*read-default-float-format*` in the reader.

I prefer the first, I think (almost) always reading unadorned floats as doubles is a much nicer behavior for the language. Previously we've had code that failed to type check when the default float format was changed in a sbclrc file.

stylewarning: Maybe the "best" solution is to make the use of the Coalton reader pervasive. Maybe we need to have a

```
#lang coalton
```

at the top of the file, or a

```
[coalton ...]
```

(or similar) reader macro that gets installed into the current readtable without creating a new named readtable. That way we can have total control over the reader, and worry less about the impedance mismatch between Coalton and Lisp's character syntax.

Maybe this issue isn't the right place to discuss this, but it seems that there's a problem worth solving here.

gefjon: > reader macro that gets installed into the current readtable

I strongly advocate against mutating the standard readtable.

eliaslfox: I reverted back to the original behavior of deferring to the host systems `*read-default-float-format*`. I don't think it's possible to read floats as doubles only within Coalton code without creating a lot of edge cases.

> That way we can have total control over the reader, and worry less about the impedance mismatch between Coalton and Lisp's character syntax.

When users use named readtables we do have full control over the reader. But if we want to keep allowing Coalton to be compiled in the standard readtable, then we need to continue to support the stringification hack.

I'm not opposed on principle to requiring the coalton readtable for Coalton code, but we also fallback to stringification for weird cases like:

```lisp
(cl:defmacro dont-do-this ()
  (cl:let ((x 'coalton-toplevel))
    `(,x (define y 0.25))))
```

We could require every `coalton-toplevel` to be caught by the reader, but that would break most of our tests :/

stylewarning: > I strongly advocate against mutating the standard readtable.

Why?

gefjon: Basically, because the standard readtable is a global mutable resource, any other system which wants to do anything readtable-ey is likely to break if you mutate the standard readtable, and debugging problems caused by a readtable having something unexpected in it is miserable. Note that there's no way (afaik) either inside or outside the spec to say "give me a readtable like the standard readtable before anyone modified it," there's just a single mutable global that everyone observes.

stylewarning: I'm wondering why you might think, for Lisp as written today, this would be particularly problematic for Coalton users. Is it more of an in-theory or in-practice issue?

gefjon: I'm opposed to globally disabling float traps too, though I feel less strongly about it because I've never had to debug an issue with it. I have worked on projects that wanted to have multiple distinct readtables which inherited from the standard readtable that would have broken if some dependent package was mutating it. In my mind, part of making Coalton seamlessly interoperate with Common Lisp is that loading Coalton should not cause the miscompilation of any conformant Common Lisp code that doesn't irreparably mutate the global environment, and that means that Coalton itself should not irreparably mutate the global environment either.

stylewarning: Anyway I don't have a definite opinion either way. I don't think most people are touching the readtable, and the ones that are probably aren't using Coalton. 

It feels dirty to mess with the readtable, but these shenanigans also don't feel nice.

Regardless of mutating the global table, maybe we need more deliberate reader syntax that's required in all circumstances, that way nobody can mess up and forget IN-READTABLE, or get bad errors/float reader issues/etc.

colescott (approved):

PR #858, opened 2023-03-30
Run tests in CI with specializations disabled

### LINKED ISSUE #607: Turn-off specialization for non-release mode tests
A specialized procedure should have the same input and output as its non-specialized counter-part, so it would make sense to have a flag to run tests without specialization. Perhaps even do so on the development 0 CI for `coalton-library`.

eliaslfox:
Fixes #607


eliaslfox:
It's just for our own testing.

PR #859, opened 2023-03-31
Fix monomorphization and add tests

### LINKED ISSUE #624: Monomorphizer fails on over-application

eliaslfox:
Fixes #624


stylewarning:
this front end refactor is paying dividends

src/utilities.lisp
```diff
@@ -156,6 +157,18 @@
          (nreverse out)
          result)))))
 
+(defun take (n list)
```
stylewarning, line 160: these functions work for any sequence

colescott (approved):

src/codegen/monomorphize.lisp
```diff
@@ -199,103 +199,104 @@ recompilation, and also maintains a stack of uncompiled candidates."
...
+
+                      (remaing-names (loop :for arg :in remaining-args
+                                           :collect (gentemp)))
+
+                      (remaining-types (loop :for arg :in remaining-args
+                                             :for ty :in (tc:function-type-arguments new-type)
+                                             :collect ty)))
+
+                 ;; All remaining arguments must be unpropagated
+                 (unless (every (lambda (x) (eq x '@@unpropagated)) remaining-args)
```
stylewarning, line 260: is it ok that EVERY is true for an empty list?

stylewarning (approved):

src/codegen/monomorphize.lisp
eliaslfox: Because of the check in the cond branch the list will never be empty.

gefjon: This seems like a great thing to note in a quick comment.

src/utilities.lisp
gefjon: Restricting their type with a declaration in a local function allows better type inference and checking, plus potentially compiling away some dispatch.

stylewarning: what you say is true yet i don't see many calls to LIST-LENGTH in our code

stylewarning: 😬

Beware, it returns NIL on circular lists.

PR #860, opened 2023-03-31
add package definition ex. to intro

stylewarning:


docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+```
+
+This way, we can now just type `str:strip-prefix`. If we are using several specific packages, we would create several nicknames:
+
+```lisp
+(defpackage #:my-package
+  (:use #:coalton)
+  (:local-nicknames (#:str    #:coalton-library/string)
+                    (#:vec    #:coalton-library/vector)
+                    (#:c      #:coalton-library/math/complex)))
```
eliaslfox, line 35: `coalton-library/math` re-exports the `coalton-library/math/*` packages. This example should demo importing math instead of a sub-package.

eliaslfox:
I think the section about PLNs should mention some of the PLNs setup coalton-user already https://github.com/coalton-lang/coalton/blob/main/library/prelude.lisp#L211-L228

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
 
 This document is aimed toward people who are already familiar with functional programming languages.
 
+## Systems and Packages
+
+Coalton uses ordinary Common Lisp packages (and ASDF systems) to organize code. If you're starting a new project, add `#:coalton` to your ASD's `:depends-on` list.
+
+Unlike Common Lisp, Coalton's standard library is organized as a large collection of packages. For example, string-related functions are in the `#:coalton-library/string` package. Refer to the [Coalton Reference](https://coalton-lang.github.io/reference) for a complete list of standard library packages.
+
+When making a new project, you will want to define a new package and establish your standard library imports. *Do not `:use` the `#:cl` or `#:common-lisp` packages!* Instead, you'll want to, at minimum, `:use #:coalton`:
```
gefjon, line 13: This should probably mention `coalton-prelude` as well. Something like:

> [...] you'll want to, at minimum, `:use #:coalton`, and almost always `#:coalton-prelude` too. [... example DEFPACKAGE ...]

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+
+```lisp
+(defpackage #:my-package
```
gefjon, line 16: I'd prefer the examples use `uiop:define-package` rather than `cl:defpackage`, but I know that the outdated ASDF/UIOP versions bundled by SBCL and CCL make this impractical. Still, I think it would be nice to add an aside that `uiop:define-package` is a better abstraction, but that the version exposed out-of-the-box by SBCL and CCL doesn't support `:local-nicknames`, and therefore that users should install a newer ASDF version if able.

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+  (:use #:coalton))
+```
+
+In idiomatic usage, this minimal package definition would be far too limiting to be useful. For instance, if we wanted to use the standard library function `strip-prefix` (which strips a prefix off of a string), we would have to type `coalton-library/string:strip-prefix` in our program. Instead, we can create a local nickname for this package:
+
+```lisp
+(defpackage #:my-package
+  (:use #:coalton)
+  (:local-nicknames (#:str    #:coalton-library/string)))
```
gefjon, line 25: Multiple spaces looks kinda weird here when there aren't multiple clauses to align. :shrug: Probably not worth getting hung up over, but it stuck out to me.

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+                    (#:c      #:coalton-library/math/complex)))
+```
+
+**We do not recommend `:use`ing more specific packages!** This is because different standard library packages may actually have symbols with the same name. For example, both the string package and the vector package have a symbol named `length`, i.e., `str:length` and `vec:length` both exist. If we `:use` either of those packages, the code we read and write in that package may be confusing if we intend to operate on both data structures.
+
+
 ## Program Structure
 
-To start, we recommend changing your package to the `COALTON-USER` package like so:
+To start, we recommend either creating your own package, or changing your package to the `COALTON-USER` package like so:
```
gefjon, line 55: I think this sentence should offer a more precise recommendation. I'd say:

> If you're playing around with Coalton in the REPL, we recommend changing your package to `COALTON-USER` like so: [... example `in-package` form ...] If you're writing code that's saved to a file, you should define your own package, like so: [... example `defpackage` form which uses `COALTON` and `COALTON-PRELUDE`, and has a `:local-nicknames` clause with an ellipsis in it ...] Whenever you need to reference a new definition from the Coalton standard library, add it to the `:local-nicknames` clause in your `defpackage` form.

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+```lisp
+(defpackage #:my-package
+  (:use #:coalton)
+  (:local-nicknames (#:str    #:coalton-library/string)
+                    (#:vec    #:coalton-library/vector)
+                    (#:c      #:coalton-library/math/complex)))
+```
+
+
+For packages that offer pervasive, common, and general functionality, such as the `#:coalton-library/classes` package, it is OK to wholesale `:use` them, like so
```
gefjon, line 39: I'm not convinced this recommendation is useful, or at least, it seems unhelpful unless presented alongside a mention of `coalton-prelude`. As I understand it, `coalton-prelude` reexports all of the symbols that fall into the category of "pervasive, common and general functionality," so we should just advise people to `:use #:coalton-prelude`, and to `:local-nickname` anything else.

docs/intro-to-coalton.md
```diff
@@ -4,9 +4,55 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
 
 This document is aimed toward people who are already familiar with functional programming languages.
 
+## Systems and Packages
+
+Coalton uses ordinary Common Lisp packages (and ASDF systems) to organize code. If you're starting a new project, add `#:coalton` to your ASD's `:depends-on` list.
```
gefjon, line 9: I've been on record on numerous occasions advocating to use `:keywords` for package names in e.g. `defpackage` forms, and `#:uninterned-symbols` for non-package names like those in `:export` or `:import-from`. I understand this is somewhat controversial (and I also don't care enough to make a stink), but I'd like to put on record (here, and everywhere else I can) that the only meaningful difference between keywords and uninterned symbols in a `defpackage`-ey context is how SL[Y|IME] autocomplete treats them, and I believe that it's generally desirable for autocomplete to suggest package names.

docs/intro-to-coalton.md
eliaslfox: I agree with this. We should strongly discourage using other library packages. I have no interest in ending up in a situation like Alexandria where adding new exports is considered a breaking change.

gefjon: I think we're pretty far away from that point, but I agree. Currently Coalton is at a place where adding exports to the prelude is acceptable, but I think it's desirable to have it always be true that adding exports to any stdlib package (other than the prelude) is a non-breaking change.

docs/intro-to-coalton.md
eliaslfox: I understand the desire, but I don't think it makes sense to put in the intro document. We want experience of trying out Coalton to be as smooth as possible. Telling people to update asdf first would detract from that.

gefjon:
I wonder if there's a way to get the docs to automatically include a listing of all the local-nicknames included by `coalton-prelude`. I don't know enough about the PLN APIs offered by Lisp impls to know if this is feasible, but I think it's worth investigating.

docs/intro-to-coalton.md
gefjon: I think you're probably right, but it bums me out.

eliaslfox:
SBCL seems to implement the same PLN APIs as Lisp Works as documented [here](http://www.lispworks.com/documentation/lw80/lw/lw-hcl.htm).

docs/intro-to-coalton.md
stylewarning: It's interesting that I basically have the opposite opinion: I only like keyword completion for keyword arguments, and at no other time. The fact that every (obviously interned) keyword autocompletes everywhere vexes me.

If it weren't for Allegro/modern-mode/etc., I'd normally advocate for `"ALL-CAPS"` strings for everything.

stylewarning:
@eliaslfox @gefjon I think I incorporated all corrections and opinions (except the keyword packages, mostly for consistency elsewhere).

docs/intro-to-coalton.md
```diff
@@ -4,15 +4,63 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+
+After creating your package `#:my-package`, you must switch to it with:
+
+```lisp
+(named-readtables:in-readtable coalton:coalton)
+
+(in-package #:my-package)
+```
+
+The `named-readtables` liine is optional, but (assuming you've included it as a dependency to your system definitition) you will have vastly better error messages. Using Coalton's reader allow for compiler errors to accurately refer to source code and provide correct line numbers.
```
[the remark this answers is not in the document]
gefjon: Also, I think it'd be nice to say how the error messages are better, rather than just that they're "vastly better." The removed paragraph from later on about `named-readtables` integration does a good job of this. I think I would replace this paragraph with:

> The `named-readtables:in-readtable` form is optional but encouraged. Using Coalton's reader allows  compiler errors to accurately refer to source code and provide correct line numbers.

docs/intro-to-coalton.md
```diff
@@ -4,15 +4,63 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+
+**We do not recommend `:use`ing any other package besides `#:coalton` and `#:coalton-prelude`.** Historically, in Common Lisp, this has caused backwards-compatibility issues when the packages themselves introduce new exported symbols. Moreover, unlike Common Lisp, Coalton follows a pattern of using the same symbol name for similarly defined functions of different packages. For example, both the string package and the vector package have a symbol named `length`, i.e., `str:length` and `vec:length` both exist.
+
+
+## Program Structure
+
+After creating your package `#:my-package`, you must switch to it with:
+
+```lisp
+(named-readtables:in-readtable coalton:coalton)
```
gefjon, line 56: It seems like the library tends to put `in-package` before `in-readtable`. It doesn't matter, but for consistency's sake we might as well use the same order in the example as in the source.

docs/intro-to-coalton.md
```diff
@@ -4,15 +4,63 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
 
 This document is aimed toward people who are already familiar with functional programming languages.
 
-## Program Structure
+## Systems and Packages
+
+Coalton uses ordinary Common Lisp packages (and ASDF systems) to organize code. If you're starting a new project, add `#:coalton` to your ASD's `:depends-on` list. For improved error messages, also depend on `#:named-readtables`.
```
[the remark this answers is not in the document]
stylewarning: They don't need to but is this not a bad practice?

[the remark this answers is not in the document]
gefjon: Directly using your dependencies' dependencies is bad practice when those transitive dependencies are an implementation detail and subject to change, but I don't think that's the case here. Coalton's use of and integration with `named-readtables` is a documented and stable part of our exposed interface. It would be gross and bad if we encouraged our users to interact with `fset` without explicitly depending on it, but I think explicitly depending on `named-readtables` is just unnecessary noise and cognitive load.

docs/intro-to-coalton.md
stylewarning: I thought about this a little. I like consistency, but then I thought: If you're going to change the reader, why not do that as soon as possible, so you don't have "pre-reader" code and "post-reader" code?

gefjon: Eh, I could go either way. I'm not sure an `in-package` form counts as "code," and I think it's nice that the `defpackage` and `in-package` go right next to each other. But I don't think it really matters, just that there's some small value in having the docs be consistent with the library.

docs/intro-to-coalton.md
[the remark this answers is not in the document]
stylewarning: I feel like there's cognitive load when a project uses `foo:bar` explicitly but does not `:depend-on` the resident system of `foo`.

docs/intro-to-coalton.md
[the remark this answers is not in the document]
stylewarning: the paragraph does mention the benefits, but i'm fine with more concision

docs/intro-to-coalton.md
```diff
@@ -4,15 +4,63 @@ Coalton is a statically typed language that is embedded in, and compiles to, Com
...
+
+```lisp
+(named-readtables:in-readtable coalton:coalton)
+
+(in-package #:my-package)
+```
+
+The `named-readtables` liine is optional, but (assuming you've included it as a dependency to your system definitition) you will have vastly better error messages. Using Coalton's reader allow for compiler errors to accurately refer to source code and provide correct line numbers.
+
+Alternatively, you will have switched to `#:coalton-user` in the REPL. Remember that Coalton packages do *not* `:use` the `#:common-lisp`/`#:cl` package, so you must prepend Common Lisp symbols with `cl:` if you need them.
```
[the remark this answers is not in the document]
stylewarning: i think it's too premature to declare writing Lisp in Coalton as rare, as everybody who has written Coalton has needed to do it extensively so far

docs/intro-to-coalton.md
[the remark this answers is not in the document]
gefjon: Ah, I somehow missed the end of that paragraph. Oops. I still think this is a good change, because "vastly better" is pretty much filler. Also, if we're still encouraging people to `:depends-on "named-readtables"`, the parenthetical is somewhat distracting. I'd propose:

> The named-readtables:in-readtable form is optional but encouraged. Using Coalton's reader allows compiler errors to accurately refer to source code and provide correct line numbers. Only include it if your system `:depends-on "named-readtables"`

docs/intro-to-coalton.md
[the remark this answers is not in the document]
gefjon: Yeah, true. I'm looking for some way to prevent people from trying to call `cl:print-line` in Coalton code without a `lisp` form and wondering why it doesn't work, when the docs call out that you can refer to CL operators using the `cl:` prefix.

docs/intro-to-coalton.md
colescott: `in-readtable` should happen after `in-package` so that the editor hints with SWANK can associate the readtable with the package for the file

stylewarning: im fine to change it since the stdlib is that way, but I don't personally think it's "correct"

eliaslfox (approved):

eliaslfox:
I'm merging this. It's a good improvement to the docs. Further changes can happen in subsequent prs.

PR #864, opened 2023-04-04
Add unused variable check

### LINKED ISSUE #12: Add unused variable warning

eliaslfox:
Add warnings for unused variables:

```
(coalton-toplevel
  (define (f x) 5))
```

```
; caught COMMON-LISP:STYLE-WARNING:
;   warn: Unused variable
;     --> /private/var/tmp/slimedunOe0:2:13
;      |
;    2 |    (define (f x) 5))
;      |               ^ variable defined here
```

Variables can be marked as intentionally used by prefixing them with `_`. If an intentionally ignored variable is used, then a warning is generated:

```
(coalton-toplevel
  (define (f _x) _x))
```

```
;   warn: Use of ignored variable
;     --> /private/var/tmp/slimeQxQzkn:2:17
;      |
;    2 |    (define (f _x) _x))
;      |               -- definition here
;      |                   ^^ first usage here

```

Fixes #12


stylewarning:
FWIW, I've enjoyed the distinction of IGNORE and IGNORABLE in CL declarations, mostly for macro-generated code. But I do like the `_x` convention for most practical uses of ignoring a variable.

gefjon:
> But I do like the _x convention for most practical uses of ignoring a variable.

I'm a fan of Rust's using `_` (the underscore with nothing after) as equivalent to CL's `ignore`, and `_foo` (an underscore with any valid identifier after) as `ignorable`.

eliaslfox:
> I'm a fan of Rust's using _ (the underscore with nothing after) as equivalent to CL's ignore, and _foo (an underscore with any valid identifier after) as ignorable.

I like this. I'll update this pr to make `_foo` equivalent to `ignorable`. Underscore bindings will be part of argument pattern matching (now tracked at #865).

src/analysis/unused-variables.lisp
```diff
@@ -0,0 +1,96 @@
...
+
+    ;; Check for unused parameters in the binding
+    (unless (tc:binding-nullary binding)
+      (loop :for var :in (tc:binding-parameters binding)
+            :do (variable-binding var used-variables file)))
+
+    ;; Check for unused variables in the body
+    (tc:traverse
+     (tc:binding-value binding)
+     (tc:make-traverse-block
```
colescott, line 42: This needs to check `do-bind`

colescott (approved):

PR #866, opened 2023-04-04
Iterator improvements

eliaslfox:
* add size tracking
* add filter-map!


gefjon:
The size hint of an `Iterator` should probably be an `(Optional (Cell UFix))`, so that `next!` can decrement it.

eliaslfox:
> The size hint of an Iterator should probably be an (Optional (Cell UFix)), so that next! can decrement it.

I don't think so. We don't need an exact size for the iterator, we just want something that's on the same order of magnitude to avoid resizing without over allocating too much.

colescott (approved):

PR #873, opened 2023-04-05
document unused variable syntax

stylewarning:


docs/intro-to-coalton.md
```diff
@@ -175,8 +178,9 @@ Here is an example of using a curried function to transform a list.
   (map (+ 2) nums)) ;; 4 5 6 7
 ```
 
-There are convenient *syntaxes* for composing functions with the
-`pipe` and `nest` macros.
+### Pipelining Syntax and Function Composition
+
+There are convenient *syntaxes* for composing functions with the `pipe` and `nest` macros.
```
gefjon, line 183: Why is "*syntaxes*" italicized?

docs/intro-to-coalton.md
```diff
@@ -192,9 +196,32 @@ There are convenient *syntaxes* for composing functions with the
...
+  --> COALTON-TOPLEVEL (NIL):1:31
+   |
+ 1 |  (COALTON-TOPLEVEL (DEFINE (F X Y)
+   |                                 ^ variable defined here
+help: prefix the variable with '_' to decare it unused
+ 1 | (COALTON-TOPLEVEL (DEFINE (F X _Y)
+   |                                --
+```
+
+As suggested, one can replace `y` with `_y`, which tells the Coalton compiler that the parameter might be intentionally unused.
```
gefjon, line 224: Should we document that it's still possible to use `_y`, though users are discouraged from doing that except in corner cases like macros? Something like:

> Variables prefixed with `_` like `_y` are still normal variables, and can be read. The following is valid Coalton:
> ```
> (define (f _x) _x)
> ```
> You should treat underscore prefixed variables as ignored whenever possible, and use a name not prefixed with `_` if you intend to use it. Reading from underscore-prefixed variables is permitted so that generated code (e.g. using macros or read-conditionals) may avoid unused variable warnings for variables which will be used in some compilation contexts but not others.

gefjon (commented):
I'm a little nervous about documenting this when we're still expecting a change in the near future to how ignored variables work, namely re: `_` patterns. Still, this is a good, descriptive change.

eliaslfox (approved):

docs/intro-to-coalton.md
```diff
@@ -192,9 +196,32 @@ There are convenient *syntaxes* for composing functions with the
...
+```lisp
+(coalton-toplevel
+  (define (f x y)
+    x))
+```
+
+Here, `y` is unused and will produce a warning:
+
+```
+COMMON-LISP:WARNING: warn: Unused variable
```
colescott, line 214: This warning looks like it was generated without the readtable enabled. Could you update this with that fixed?

stylewarning: This was in the REPL FWIW. Yes I'll update it.

docs/intro-to-coalton.md
stylewarning: well, for emphasis of course. they're not functions for composing functions :)

stylewarning:
Made all the suggested changes.

eliaslfox (approved):

PR #874, opened 2023-04-05
document describing a simple usage of `coalton-testing`

gefjon:
Note that there are a handful of TODO lines in there.

### DIFF
--- a/docs/intro-to-coalton-testing.md
+++ b/docs/intro-to-coalton-testing.md
@@ -7,25 +7,23 @@ The fact that your code type-checks and compiles should assure you that it's fre
 large class of bugs, but that doesn't necessarily mean that it does what you expect. If
 you're worried about whether you've got the logic right, it's time to write a test suite!
 
+You can find an example project containing the code from this document in [examples/coalton-testing-example-project](../examples/coalton-testing-example-project)
+
 ## Adding a Test System to Your System Definition
 
 For a project `foo`, you should already have a system definition file `foo.asd` located in
 the project's root directory, which looks something like:
 
-```
+```lisp
 (defsystem "foo"
   :depends-on ("coalton" "named-readtables" ...other dependencies?)
   :components ((:file "main")
                ...other components?))
 ```
 
-Don't worry if your system looks a little different. It might have `:serial t`, and it
-might specify an `:author` or a `:license`, it might have a `:pathname`, and if you're
-really cool it might have `:class :package-inferred-system`.
-
 To that `defsystem` form, add the following line:
 
-```
+```lisp
   :in-order-to ((test-op (test-op "foo/test")))
 ```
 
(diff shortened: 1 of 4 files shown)


docs/intro-to-coalton-testing.md
```diff
@@ -0,0 +1,220 @@
...
+the project's root directory, which looks something like:
+
+```
+(defsystem "foo"
+  :depends-on ("coalton" "named-readtables" ...other dependencies?)
+  :components ((:file "main")
+               ...other components?))
+```
+
+Don't worry if your system looks a little different. It might have `:serial t`, and it
```
eliaslfox, line 22: I think this note is a little superfluous.

docs/intro-to-coalton-testing.md
```diff
@@ -0,0 +1,220 @@
...
+(defsystem "foo/test"
+  :depends-on ("foo" "coalton/testing")
+  :pathname "test/"
+  :serial t
+  :components ((:file "test"))
+  :perform (test-op (o s)
+             (symbol-call '#:foo/test '#:run-tests)))
+```
+
+TODO: do we want to tell users to explicitly depend on `fiasco`, the same way we tell them
```
eliaslfox, line 50: I guess so

docs/intro-to-coalton-testing.md
```diff
@@ -0,0 +1,220 @@
...
+`define-test` is a macro that was generated by the `coalton-fiasco-init` form you wrote
+earlier. (This is an implementation detail; try not to think too much about it.) For your
+very first `define-test`, try writing:
+
+```
+(define-test my-first-test ()
+  (is True))
+```
+
+`my-first-test` is the name of your test. The empty list `()` is its argument list. I
```
eliaslfox, line 138: Probably an unnecessary note for a tutorial.

docs/intro-to-coalton-testing.md
```diff
@@ -0,0 +1,220 @@
...
+(define-test test-one-element-list ()
+  (match (one-element-list 0)
+    ((Nil) (is False "ONE-ELEMENT-LIST returned an empty list!"))
+    ((Cons elt (Nil)) (is (== 0 elt)))
+    ((Cons _ _) (is False "ONE-ELEMENT-LIST returned a list with more than 1 element!"))))
+```
+
+## Running your tests
+
+Now that you've defined a test system and some tests to go in it, you can run your tests
```
eliaslfox, line 219: There should be a note here that running asdf commands should not be done in a coalton package. Doing so tends to corrupt asdf.

eliaslfox (commented):
Thanks for taking the time to document this. I think it would be worth also having an example project skeleton, something like [coalton-example](https://github.com/eliaslfox/coalton-example).

docs/intro-to-coalton-testing.md
gefjon: Is this still the case now that `coalton` reexports `defpackage` and `in-package`? And if so, is it a breakage between Coalton and ASDF, or is it that some other systems expect to be in a package that uses CL before being loaded, and contain code with that expectation before defining and entering their own packages?

gefjon:
Still one TODO: Fiasco and FiveAM's `is` macros both privilege the argument order `(is (PREDICATE EXPECTED-VALUE COMPUTED-VALUE))` when writing `is` tests of binary comparisons, generating error messages that describe the first argument as the expected value, and the second argument as the value computed by the tests. `coalton/testing` does not do this. Currently, the intro-to-coalton-testing recommends that users obey that convention anyways. Do we want to document preferring this convention? Personally, I lean towards yes (I mean, I wrote it that way, duh) because I find it easy to read, and I value consistency with the rest of the CL world, but I can see an argument for removing the recommendation from the document, or adding language that specifies to prefer it because it's a convenient convention, rather than just saying to prefer it.

eliaslfox:
I always found the expected/actual designations to be a bit arbitrary. I also frequently get it "wrong" in my own code. The [fiasco readme](https://github.com/joaotavora/fiasco) seems to use both styles.

I personally wouldn't recommend a specific style, but I think the wording in the document is fine as is.

eliaslfox:
Sorry, I completely forgot about this.

eliaslfox (approved):

PR #889, opened 2023-04-13
Add runtime-repr-of to extract type-entry-runtime-type

simendsjo:
This is necessary to get the underlying `repr` of a coalton type without having a value available.

### DIFF
--- a/src/debug.lisp
+++ b/src/debug.lisp
@@ -175,6 +175,11 @@
   "Lookup the kind of type SYMBOL in the global environment"
   (tc:kind-of (coalton-impl/typechecker::type-entry-type (tc:lookup-type entry:*global-environment* symbol))))
 
+(defun coalton:runtime-type-of (symbol)
+  "Lookup the runtime type of value SYMBOL in the global environment"
+  (let ((type (tc:lookup-type entry:*global-environment* symbol)))
+    (coalton-impl/typechecker::type-entry-runtime-type type)))
+
 (defun coalton:lookup-code (name)
   "Lookup the compiled code of a given definition"
   (declare (type symbol name))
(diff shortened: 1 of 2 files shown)


colescott:
Does `(runtime-repr (the (Proxy <Type>) Proxy))` work for your use-case?

simendsjo:
> Does (runtime-repr (the (Proxy <Type>) Proxy)) work for your use-case?

Yes, thanks! Guess this PR doesn't make sense then? I don't understand why this works though. It matches the following instance? And then there's some compiler magic which will actually implement an actual instance for each type I define (guessing from the comment?)

```lisp
  (define-class (RuntimeRepr :a)
    "Types which have a runtime LispType representation.

`runtime-repr` corresponds to the type emitted by the Coalton compiler for the type parameter to the given Proxy.

The compiler will auto-generate instances of `RuntimeRepr` for all defined types."
    (runtime-repr (Proxy :a -> LispType)))
```

colescott:
When you write `(runtime-repr (the (Proxy MyType) Proxy))` the compiler looks for a method matching `runtime-repr :: Proxy MyType -> LispType`, which is provided by the `(RuntimeRepr MyType)` instance (filling the type variable `:a` to `MyType`).

This `RuntimeRepr` instance is automatically generated by the Coalton compiler as part of the `define-type` for `MyType` based on that `type-entry-runtime-type` ([this is done manually](https://github.com/coalton-lang/coalton/blob/45b0833d4b51a9cbb29f65ef891e5e2054df3bef/library/types.lisp#L52) for some early types defined within the compiler since they exist without explicit `define-type` forms). The compiler will write `RuntimeRepr` instances for all user-defined types.

PR #890, opened 2023-04-14
Dual numbers

Jobhdez:
here is an example:

```
* (in-package #:coalton-user)
* (coalton (mul-duals (Dual (*Real* 3) (*Dual* 4)) (Dual (*Real* 1) (*Dual* 2))))
#.(DUAL #.(*REAL* 3) #.(*DUAL* 10))
* (coalton (add-duals (Dual (*Real* 3) (*Dual* 4)) (Dual (*Real* 1) (*Dual* 2))))
#.(DUAL #.(*REAL* 4) #.(*DUAL* 6))
* (coalton (sub-duals (Dual (*Real* 3) (*Dual* 4)) (Dual (*Real* 1) (*Dual* 2))))
#.(DUAL #.(*REAL* 2) #.(*DUAL* 2))
*

### CODE REVIEW

File: coalton.asd [hunk 1]
```diff
@@ -156,7 +156,8 @@
                              (:file "real")
                              (:file "complex")
                              (:file "elementary")
-                             (:file "dyadic")))
+                             (:file "dyadic")
+			     (:file "dual")))
```
stylewarning, line 160: fix indentation

library/math/dual.lisp
```diff
@@ -0,0 +1,65 @@
...
+
+(coalton-toplevel
+
+ (define-type *Real*
+     (*Real* Integer))
+
+ (define-type *Dual*
+     (*Dual* Integer))
+ 
+ (define-type Dual
```
stylewarning, line 30: I would define a type

```
(define-type (Dual :t) 
  (Dual :t :t)) 
```

so that it can work for any underlying number type.

library/math/dual.lisp
```diff
@@ -0,0 +1,65 @@
...
+     (Dual *Real* *Dual*))
+ 
+ (declare dual-real (Dual -> Integer))
```
stylewarning, line 33: i'd name it `real-part`

with the above change the function would have type `Dual :t -> :t`

library/math/dual.lisp
```diff
@@ -0,0 +1,65 @@
...
+ (define (dual-real d1)
```
stylewarning, line 34: you can use argument pattern matching

```
(define (real-part (Dual r _))
  r)
```

library/math/dual.lisp
```diff
@@ -0,0 +1,65 @@
...
+     (match d1
+	    ((Dual (*Real* x) (*Dual* y)) x)))
+ 
+ (declare dual-dual (Dual -> Integer))
+ (define (dual-dual d1)
+     (match d1
+	    ((Dual (*Real* x) (*Dual* y)) y)))
+
+ (declare add-duals (Dual -> Dual -> Dual))
```
stylewarning, line 43: You ought to implement an instance of the `Num` type class.

```
(define-instance (Num (Dual :t)) 
  (define (+ ...)
    ...)

  (define (- ...)
    ...)

  (define (* ...)
    ...)

  (define (from-int ...)
    ...))
```

Then the normal arithmetic operators will work for dual numbers.

stylewarning (changes requested):
Great job so far. On the right track.

library/math/dual.lisp
```diff
@@ -0,0 +1,66 @@
...
+ 
+ (declare primal-part (Dual :t -> :t))
+ (define (primal-part (Dual r _)) r)
+
+ (declare dual-part (Dual :t -> :t))
+ (define (dual-part (Dual _ d))
+     d)
+
+ (define-instance (Eq :t => Eq (Dual :t))
+     (define (== a b)
```
stylewarning, line 40:
```
(define (== (Dual a b) (Dual p q))
  (and (== a p)
       (== b q)))
```

library/math/dual.lisp
```diff
@@ -0,0 +1,66 @@
...
+
+(in-package #:coalton-library/math/dual)
+
+(coalton-toplevel
+
+ (define-type *Real*
+     (*Real* Integer))
+
+ (define-type *Dual*
+     (*Dual* Integer))
```
stylewarning, line 27: these can be removed now

library/math/dual.lisp
```diff
@@ -0,0 +1,66 @@
...
+   #:coalton-library/builtin
+   #:coalton-library/classes
+   #:coalton-library/functions
+   #:coalton-library/math/arith)
+  (:export
+   #:Dual
+   #:primal-part
+   #:dual-part
+   #:*Dual*
+   #:*Real*))
```
stylewarning, line 17: these can be removed now

library/prelude.lisp
```diff
@@ -84,6 +85,20 @@
...
    #:even?
    #:odd?)
+  
+  (:import-from
+   #:coalton-library/math/dual
+   #:Dual
+   #:primal-part
+   #:dual-part
+   #:*Real*
+   #:*Dual*)
```
stylewarning, line 95: remove

library/prelude.lisp
```diff
@@ -84,6 +85,20 @@
...
+  (:export
+   #:Dual
+   #:primal-part
+   #:dual-part
+   #:*Real*
+   #:*Dual*)
```
stylewarning, line 101: remove

library/math/dual.lisp
```diff
@@ -0,0 +1,66 @@
...
+ (define (dual-part (Dual _ d))
+     d)
+
+ (define-instance (Eq :t => Eq (Dual :t))
+     (define (== a b)
+	 (lisp Boolean (a b)
+	       (cl:equal a b))))
+
+ (define-instance (Num :t => Num (Dual :t))
+     (define (+ dual1 dual2)
```
stylewarning, line 45: You might be able to make these functions more readable by doing:

```
(define (+ (Dual p1 d1) (Dual p2 d2))
  ...)
```

so you don't need as many calls go `primal-part` and `dual-part`

stylewarning (changes requested):
I know this is a WIP still.

Three things will still be necessary:

1. Add doc strings to all types and new functions.

2. Fix indentation to not use tabs (you can do `(setq indent-tabs-mode nil)` in Emacs), and use correct levels of indentation (SLIME will do this automatically when you hit the tab key).

3. Implement Reciprocable, Exponentiable, and Trigonometric. (There might be more, but that's a
good start.)

Great update, looking forward to this PR developing.

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+(named-readtables:in-readtable coalton:coalton)
+
+(coalton-toplevel
+
+  (define-type (Dual :t)
+    "Representation of a Dual number in the form (a + bE)."
+    (Dual :t :t))
+  
+  (declare primal-part (Dual :t -> :t))
+  (define (primal-part (Dual r _)) r)
```
stylewarning, line 28: r -> p

put the p on a newline for clarity

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    d)
+
+  ;; base-part to get the base of a given log
+  (declare base-part (Dual :t -> :t))
+  (define (base-part (Dual b _))
+    b)
+
+  (declare exponent-part (Dual :t -> :t))
+  (define (exponent-part (Dual expt _))
+    expt)
```
stylewarning, line 41: Not needed.

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+  
+  (define-instance (Eq :t => Eq (Dual :t))
+    
+    (define (== (Dual a b) (Dual p q))
+      "Check that the given Dual numbers are the same."
```
stylewarning, line 46: I don't think any of the *instance* doc strings are necessary *unless* the implementation is doing something special, unexpected, or noteworthy. I'd delete these personally.

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    
+    (define (- (Dual p1 d1) (Dual p2 d2))
+      "Subtraction of Dual numbers."
+      (Dual (- p1 p2)
+            (- d1 d2)))
+
+    (define (* (Dual p1 d1) (Dual p2 d2))
+      "Multiplication of Dual numbers."
+      (Dual (* p1 p2)
+            (+ (* p1  d2)
```
stylewarning, line 65: extra stray space

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    (define (/ (Dual p1 d1) (Dual p2 d2))
+      "Division of Dual numbers."
+      (Dual (/ p1 p2)
+            (/ (- (* d1 p2)
+                  (* p1 d2))
+               (* p2 p2))))
+    
+    (define (reciprocal (Dual p1 d1))
+      "Reciprocal of given Dual number."
+      (Dual (/ 1 p1)
```
stylewarning, line 82: (reciprocal p1)

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+            (/ d1
```
stylewarning, line 83: I think you need to (negate ...) this. derivative of 1/x is -1/x^2.

stylewarning, line 83: no need for newline here. doesn't offer additional clarity.

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    
+    (define (cos (Dual p1 d1))
+      "Cos of a given Dual Number."
+      (Dual (cos p1)
+            (* (* -1 d1) (sin p1))))
+    
+    (define (tan (Dual p1 d1))
+      "Tan of the given Dual number."
+      (Dual (tan p1)
+            (/ d1 (pow (cos p1) 2))))
```
stylewarning, line 102: use ^ not pow

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    
+    (define (asin (Dual p1 d1))
+      "Asin of the given Dual number."
+      (Dual (asin p1)
+            (* d1
+               (/ 1 (sqrt (- 1 (pow p1 2)))))))
```
stylewarning, line 108: use ^ not pow

stylewarning, line 108: use reciprocal not (/ 1 ...

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    
+    (define (acos (Dual p1 d1))
+      "Acos of the given Dual number."
+      (Dual (acos p1)
+            (* d1
+               (/ -1 (sqrt (- 1 (pow p1 2)))))))
```
stylewarning, line 114: ^ not pow

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+    
+    (define (atan (Dual p1 d1))
+      "Atan of the given Dual number."
+      (Dual (atan p1)
+            (/ d1
+               (+ 1 (pow p1 2))))))
```
stylewarning, line 120: ^ not pow

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+
+  (define-instance ((Num :t) (Exponentiable :t) (Reciprocable :t) => (Exponentiable (Dual :t)))
+    
+    (define (exp (Dual p1 d1))
+      "Exp of the given Dual number."
+      (Dual (exp p1)
+            (* (exp p1)
```
stylewarning, line 127: `(* d1 (exp p1))`

easier to read

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+               d1)))
+
+    (define (pow (Dual p1 d1) (Dual n dummy))
```
stylewarning, line 130: `dummy` is not actually a dummy. POW should let us take arbitrary powers of duals to duals. 

implement `(pow a b)` as `(exp (* b (ln a)))`

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+      "Pow of the given Dual number."
+      (Dual (pow p1 n)
+            (* d1
+               (* n
+                  (pow p1 (- n 1))))))
+    
+    (define (ln (Dual p1 d1))
+      "ln of the the given Dual number."
+      (Dual (ln p1)
+            (* (/ 1 p1)
```
stylewarning, line 140: `(* d1 (reciprocal p1))`

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+               d1)))
+    
+    (define (log (Dual base dummy) (Dual p1 d1))
```
stylewarning, line 143: dummy is not a dummy. implement `(log b x)` as `(/ (ln x) (ln b))`

library/math/dual.lisp
```diff
@@ -0,0 +1,166 @@
...
+      "Log of the given Dual number."
+      (Dual (log base p1)
+            (/ p1 d1))))
+
+  (define-instance ((Num :t) (Radical :t) (Reciprocable :t) => (Radical (Dual :t)))
+    
+    (define (nth-root n (Dual p1 d1))
+      "N-th root of the given Dual."
+      (Dual (nth-root n p1)
```
stylewarning, line 152: this math is incorrect. it should be the derivative of x^(1/n).

stylewarning (changes requested):
some small math mistakes and mostly style simplifications

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
...
+    (define (- (Dual p1 d1) (Dual p2 d2))
+      (Dual (- p1 p2)
+            (- d1 d2)))
+
+    (define (* (Dual p1 d1) (Dual p2 d2))
+      (Dual (* p1 p2)
+            (+ (* p1  d2) (* d1 p2))))
+    
+    (define (fromInt z)
+      (Dual (fromInt z) (fromInt 0))))
```
eliaslfox, line 60: You can just write `0` here. The `fromInt` call is implicit on integer literals.

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
...
+     #:coalton-library/functions
+     #:coalton-library/math/arith
+     #:coalton-library/math/elementary
+     #:coalton-library/math/integral)
+  (:export
+   #:Dual
+   #:primal-part
+   #:dual-part))
+
+(in-package #:coalton-library/math/dual)
```
eliaslfox, line 19: Add a newline after `in-package`.

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
...
+(named-readtables:in-readtable coalton:coalton)
+
```
eliaslfox, line 21: Add the stdlib optimize declaim after `in-readtable`
```
#+coalton-release
```

library/prelude.lisp
```diff
@@ -84,6 +115,16 @@
    #:mod
    #:even?
    #:odd?)
+  
+  (:import-from
```
eliaslfox, line 119: Don't add dual numbers to the prelude.

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
+;;;; dual.lisp
+;;;;
+;;;; dual numbers to enable automatic differentiation
+
+(coalton-library/utils:defstdlib-package #:coalton-library/math/dual
```
eliaslfox, line 5: `defstdlib-package` unfortunately needs to be manually indented

library/prelude.lisp
```diff
@@ -3,25 +3,26 @@
...
+     #:coalton-library/math/integral
+     #:coalton-library/math/real
+     #:coalton-library/math/complex
+     #:coalton-library/math/elementary
+     #:coalton-library/math/dual))
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/MATH")
 
 (uiop:define-package #:coalton-prelude
```
eliaslfox, line 21: `uiop:define-package` also needs to be manually indented

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
...
+      p)
+
+  (declare dual-part (Dual :t -> :t))
+  (define (dual-part (Dual _ d))
+      d)
+  
+  (define (sq x)
+      (* x x))
+  
+  (define-instance (Eq :t => Eq (Dual :t))
```
eliaslfox, line 39: Don't put newlines after `define-instance`.

library/math/dual.lisp
```diff
@@ -0,0 +1,135 @@
...
+(named-readtables:in-readtable coalton:coalton)
+
+(coalton-toplevel
+
+  (define-type (Dual :t)
+    "Representation of a Dual number in the form (a + bE)."
+    (Dual :t :t))
+  
+  (declare primal-part (Dual :t -> :t))
+  (define (primal-part (Dual p _))
```
eliaslfox, line 29: I think your emacs indentation is misconfigured. Forms following a define should be indented two spaces not four. You might need [something like this](https://github.com/eliaslfox/dotfiles/blob/main/nixos/users/elf/files/init.el#L113-L114) to configure it.

eliaslfox (commented):
I had a few nitpicks, but this looks pretty good. I think `Dual` should also implement `Ord` and `Hash`. The instances can be pretty similar to [the ones for `Tuple`](https://github.com/coalton-lang/coalton/blob/main/library/tuple.lisp#L55).

stylewarning:
> I had a few nitpicks, but this looks pretty good. I think `Dual` should also implement `Ord` and `Hash`. The instances can be pretty similar to [the ones for `Tuple`](https://github.com/coalton-lang/coalton/blob/main/library/tuple.lisp#L55).

Hash, definitely. Ord is debatable (outside of Coalton even). Some say it's most practical to have Ord implement comparison on just the "primal" part so almost any code can be auto differentiated "for free", but it comes with caveats. (Same with Eq.)

(P.S., Sorry, I accidentally edited your answer instead of replying.)

eliaslfox:
That makes sense. In that case, Eq should only compare primal and Hash should only hash primal. Ord can do the same. FWIW I looked at [this Haskell library for automatic differentiation](https://hackage.haskell.org/package/fad-1.1.0.1/docs/Numeric-FAD.html) and it seems to do the same.

library/prelude.lisp
Jobhdez: I added the import because otherwise I get the error: 

```
help: Did you mean COALTON-LIBRARY/MATH/DUAL:DUAL?
```

when I type `(coalton (+ (Dual 2 3) (Dual 3 4)))` inside the `coalton-user` package. it says that `Dual` is an unknown package.

library/math/dual.lisp
Jobhdez: the link is broken.

Jobhdez:
> That makes sense. In that case, Eq should only compare primal and Hash should only hash primal. Ord can do the same. FWIW I looked at [this Haskell library for automatic differentiation](https://hackage.haskell.org/package/fad-1.1.0.1/docs/Numeric-FAD.html) and it seems to do the same.

so just to confirm `Eq` and `Hash` and `Ord` should only work on primals? I just want to make sure before I push.

library/prelude.lisp
gefjon: Use the `math` nickname, like `(coalton (+ (math:Dual 2 3) (math:Dual 3 4)))`.

Jobhdez: Thanks

gefjon:
> Hash, definitely. Ord is debatable (outside of Coalton even). Some say it's most practical to have Ord implement comparison on just the "primal" part so almost any code can be auto differentiated "for free", but it comes with caveats. (Same with Eq.)

> Eq should only compare primal and Hash should only hash primal. Ord can do the same.

This is scary to me; I think `(== (Dual 1 0) (Dual 1 1)) ;=> True` violates the principle of least surprise.

If it's really the most useful set of `Eq`, `Hash`, `Ord` instances, I think we need to be really careful in documenting that behavior and the motivations behind it.

I think personally, my design intuition would be to have `Dual` implement `Eq` and `Hash` by both parts, and not implement `Ord`, and then to offer a newtype that has `Eq`, `Ord` and `Hash` by only the primal part.

library/math/dual.lisp
eliaslfox: Not sure what's going on. I'll just copy and paste. My slime use-package looks like:

```
(use-package slime
    :ensure t
    :config
    (progn
        (load (expand-file-name "~/.local/share/quicklisp/slime-helper.el"))
        (setq lisp-indent-function 'common-lisp-indent-function)
        (push 'slime-indentation slime-contribs)
        (setq inferior-lisp-program "sbcl --noinform")))
```

My guess is that you need the lines mentioning indentation to have eMacs format lisp as Common Lisp.

Jobhdez: Ok - I have just pushed. I had to manually indent. I added your config to my `.emacs` but indentation was off a little still.

eliaslfox:
I think having `Eq` look at the dual part breaks equality more generally. Requiring newtypes to get the desired instances would mean that numerical code could no longer be auto-differentiated for free.

eliaslfox:
@Jobhdez there's been a lot of back and forth for style changes in this pr. In general the code should be formatted to match the other standard library files. See [vector](https://github.com/coalton-lang/coalton/blob/main/library/vector.lisp) as an example.

Jobhdez:
> @Jobhdez there's been a lot of back and forth for style changes in this pr. In general the code should be formatted to match the other standard library files. See [vector](https://github.com/coalton-lang/coalton/blob/main/library/vector.lisp) as an example.

@eliaslfox I have just fixed indentation. I used two spaces and made it look like `vector.lisp`

Jobhdez:
What else do I have to do to merge this pr? Thanks

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+    (define (exp (Dual p1 d1))
+      (Dual (exp p1)
+            (* d1 (exp p1))))
+
+    (define (pow dual1 dual2)
+      (exp (* dual2 (ln dual1))))
+     
+    (define (ln (Dual p1 d1))
+      (Dual (ln p1)
+            (* d1 (reciprocal p1))))
```
stylewarning, line 109: (/ d1 p1)

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+   
+  (define-instance ((Num :t) (Radical :t) (Reciprocable :t) (Exponentiable :t) => (Radical (Dual :t)))
+    (define (nth-root n (Dual p1 d1))
+      (let ((n* (fromInt n)))
+        (Dual (pow  p1 (/ 1 n*))
+  	      (pow (* p1 (/ d1 n*)) (- (reciprocal n*) 1)))))
+     
+    (define (sqrt (Dual p1 d1))
+      (Dual (sqrt p1)
+            (/ (* (/ 1 2) p1)
```
stylewarning, line 122: indentation etc here.

also

(/ p1 (* 2 (sqrt ...

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+      (Dual (ln p1)
+            (* d1 (reciprocal p1))))
+     
+    (define (log dual1 dual2)
+      (/ (ln dual2) (ln dual1))))
+   
+  (define-instance ((Num :t) (Radical :t) (Reciprocable :t) (Exponentiable :t) => (Radical (Dual :t)))
+    (define (nth-root n (Dual p1 d1))
+      (let ((n* (fromInt n)))
+        (Dual (pow  p1 (/ 1 n*))
```
stylewarning, line 117: (pow x (/ 1 n*)) -> (nth-root n x)

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+    (define (sqrt (Dual p1 d1))
+      (Dual (sqrt p1)
+            (/ (* (/ 1 2) p1)
+  	       (sqrt p1)))))
+
+  (define-instance ((Ord :t) (Ord :t) => Ord (Dual :t))
+    (define (<=> (Dual p1 d1) (Dual p2 d2))
+      (match (<=> p1 p2)
+        ((LT) LT)
+  	((GT) GT)
```
stylewarning, line 129: indentation

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+  	((EQ) (<=> p1 d2)))))
```
stylewarning, line 130: this is not correct

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+        (Dual (pow  p1 (/ 1 n*))
+  	      (pow (* p1 (/ d1 n*)) (- (reciprocal n*) 1)))))
+     
+    (define (sqrt (Dual p1 d1))
+      (Dual (sqrt p1)
+            (/ (* (/ 1 2) p1)
+  	       (sqrt p1)))))
+
+  (define-instance ((Ord :t) (Ord :t) => Ord (Dual :t))
+    (define (<=> (Dual p1 d1) (Dual p2 d2))
```
stylewarning, line 126: I think the definition of this will
just be

    (<=> p1 p2)

ignore the other vars with _

library/math/dual.lisp
```diff
@@ -0,0 +1,139 @@
...
+      (match (<=> p1 p2)
+        ((LT) LT)
+  	((GT) GT)
+  	((EQ) (<=> p1 d2)))))
+
+  (define-instance ((Hash :t) (Hash :t) => (Hash (Dual :t)))
+    (define (hash (Dual p1 d1))
```
stylewarning, line 133: ignore d1 with _

library/prelude.lisp
```diff
@@ -3,25 +3,26 @@
...
-   #:coalton-library/math/arith
-   #:coalton-library/math/num
-   #:coalton-library/math/bounded
-   #:coalton-library/math/conversions
-   #:coalton-library/math/fraction
-   #:coalton-library/math/integral
-   #:coalton-library/math/real
-   #:coalton-library/math/complex
-   #:coalton-library/math/elementary))
+    (:use-reexport
```
stylewarning, line 6: these shouldn't have got indented

stylewarning (changes requested):
some math stuff

eliaslfox:
Is there a reasonable implementation of [Complex](https://coalton-lang.github.io/reference/#complex-class) for Dual?

library/math/dual.lisp
```diff
@@ -0,0 +1,176 @@
+;;;; dual.lisp
+
+"
```
eliaslfox, line 3: Good doc. Add it as the `:documentation` part of the `defpackage` so that it shows up on the doc generator. The docs for `Hash`, `Eq`, and `Ord` should be moved to the docstrings for their respective `define-instance` forms.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+
+    consider you have the given expression f(x) = 3x+2 and you want to calculate
+    f(4) and f'(4).
+
+    You then convert 4 into a dual number which gives:
+
+        4+1ε
+
+    then the computation is as follows:
+
```
stylewarning, line 32: This math isn't stated very clearly, because you've written:

$$f(x)=3x+2$$

as

$$f(x) = (3+0\varepsilon)x+(2+0\varepsilon).$$

This isn't bad in and of itself, but the exposition could be clearer, or this could be eliminated altogether in the following way:

We have $f(x)=3x+2$. So:
$$f(4+\varepsilon) = 3(4+\varepsilon)+2 = 12 + 3\varepsilon+2 = 14+3\varepsilon.$$

In this case, we don't need to make use of the fact that $\varepsilon^2 = 0$.

So to move forward, I'd recommend one of three things:

- Option 1: Keep the current exposition, but be clear that you are now evaluating the function $f(x) = (3+0\varepsilon)x+(2+0\varepsilon)$.
- Option 2: Change the current exposition to use the simplified approach to evaluating $f(x) = 3x+2$.
- Option 3: (My preference:) Use an example like $f(x) = 3x^2-1$ instead, with the simpler style of exposition, just so we can see the $\varepsilon^2$ law in action.

Choosing any one of these three options is required for this PR.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+   in this result, the primal 14 is the value of f(4) and the dual is the value of
+   of f'(4).
+
+   Haskell has an automatic differentiation library and you can find it here
+   https://hackage.haskell.org/package/ad.
+
+   limitations:
+
+     some limitations is that `Ord` and `Eq` and `Hash` only work on primals.
```
stylewarning, line 50: > We have decided to implement `Ord`, `Eq`, and `Hash` to look at only the primal part of numbers. This is so the `Dual` type can be used primarily for the purpose of automatic differentiation of existing code, and not for general abstract mathematics. If you need these type classes acting in the usual way (i.e., on both primal and dual parts), then we recommend making your own data type which wraps a dual number.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+
+(named-readtables:in-readtable coalton:coalton)
+
+#+coalton-release
+(cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
+
+(coalton-toplevel
+
+  (define-type (Dual :t)
+    "Representation of a Dual number in the form (a + bε) where a and b are  real numbers and ε satisfie ε**2 = 0 and ε != 0.  Note: `Eq`, and `Ord` and `Hash` only work on the primal component."
```
stylewarning, line 69: Some more typographic mistakes. I don't want to hold up this PR, but these should be fixed in a follow-up PR.

- lowercase "Dual" in the description
- extra unnecessary spaces between "are real"
- no need for parens around $a+b\varepsilon$
- misspelling "satisfie"
- inconsistent use of `^` and `**`. We recommend using `^` to be consistent with our standard library.
- "only work on" -> "only make use of"

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+    (Dual :t :t))
+   
+  (declare primal-part (Dual :t -> :t))
+  (define (primal-part (Dual p _))
```
stylewarning, line 73: Every exported function should have documentation. I'm fine if we do this in a future PR.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+      (Dual (- p1 p2) (- d1 d2)))
+
+    (define (* (Dual p1 d1) (Dual p2 d2))
+      (Dual (* p1 p2)
+            (+ (* p1  d2) (* d1 p2))))
+     
+    (define (fromInt z)
+      (Dual (fromInt z) 0)))
+   
+  (define-instance (Reciprocable :t => Reciprocable (Dual :t))
```
stylewarning, line 102: In many functions below, I think it would be clearer to have functions like

```
(defun (d-recip x)
  (negate (recip (sq x))))

(defun (d-sin x)
  (cos x))

(defun (d-cos x)
  (negate (d-sin x)))
```

and so on. It would make implementing the instances a lot more clear and easier to review.

It's not required for this PR but I'd recommend it for future work.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+    (define (/ (Dual p1 d1) (Dual p2 d2))
+      (Dual (/ p1 p2)
+            (/ (- (* d1 p2)
+                  (* p1 d2))
+  	       (* p2 p2))))
```
stylewarning, line 107: use `sq`

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+     
+    (define (reciprocal (Dual p1 d1))
+      (Dual (reciprocal p1)
+            (/ (negate d1) (* p1 p1)))))
```
stylewarning, line 111: use `sq`

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+      (Dual (acos p1)
+            (negate (/ d1 (sqrt (- 1 (sq p1)))))))
+     
+    (define (atan (Dual p1 d1))
+      (Dual (atan p1)
+            (/ d1 (+ 1 (sq p1))))))
+
+  (define-instance ((Num :t) (Exponentiable :t) (Reciprocable :t) => (Exponentiable (Dual :t)))
+    (define (exp (Dual p1 d1))
+      (Dual (exp p1)
```
stylewarning, line 140: @eliaslfox Do you have thoughts on us trying to avoid multiple calculation of expressions like `(exp p1)`?

- We could do it because it 1/2's the cost
- We could avoid it if we plan on trying to do common subex. elim. analyses.

My Common Lisp instinct is to lift this into a `let` because I know CL generally is unable to do CSE on function calls it doesn't know about.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+     
+    (define (ln (Dual p1 d1))
+      (Dual (ln p1)
+            (/ d1 p1)))
+     
+    (define (log dual1 dual2)
+      (/ (ln dual2) (ln dual1))))
+   
+  (define-instance ((Num :t) (Radical :t) (Reciprocable :t) (Exponentiable :t) => (Radical (Dual :t)))
+    (define (nth-root n (Dual p1 d1))
```
stylewarning, line 154: this needs to be fixed:

$$
\begin{align*}
\frac{d}{dx}\sqrt[n]{x}
&= \frac{d}{dx}x^{1/n}\\
&= \frac{1}{n}x^{1/n - 1}\\
&= \frac{1}{n}x^{1/n - n/n}\\
&= \frac{1}{n}x^{(1-n)/n}\\
&=  \frac{1}{n}x^{-(n-1)/n}\\
&= \frac{1}{nx^{(n-1)/n}}
\end{align*}
$$

So for $\sqrt[n]{a+b\varepsilon}$, we have:

$$
\sqrt[n]{a}+\frac{b}{na^{(n-1)/n}}\varepsilon.
$$

N.B., I do **not** recommend "simplifying" $x^{(n-1)/n}$ to $\sqrt[n]{x^{n-1}}$, as tempting as it may be to make this recursive call. This is because $x^{n-1}$ may be inadvertently large and we could lose precision.

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+      (let ((n* (fromInt n)))
+        (Dual (nth-root n p1)
+  	      (pow (* p1 (/ d1 n*)) (- (reciprocal n*) 1)))))
+     
+    (define (sqrt (Dual p1 d1))
+      (Dual (sqrt p1)
+            (/ p1 (* 2 (sqrt p1))))))
```
stylewarning, line 161: typo: $d_1/2\sqrt{p_1}$

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+
+  (define-instance ((Ord :t) (Ord :t) => Ord (Dual :t))
+    "Note: Ord only works on the primal components."
```
stylewarning, line 164: "works on" -> "uses"

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+    (define (<=> (Dual p1 _) (Dual p2 _))
+      (<=> p1 p2)))
+      
+  (define-instance ((Hash :t) (Hash :t) => (Hash (Dual :t)))
+    "Note: Hash only works on the primal component."
```
stylewarning, line 169: "works on" -> "uses"

library/math/dual.lisp
```diff
@@ -0,0 +1,177 @@
...
+    (define (hash (Dual p1 _))
+      (hash p1))))
```
stylewarning, line 171: I would actually recommend mixing a constant hash value in with this here, so that the hash value of `(Dual :t)` is explicitly different from `:t`. Look at how other hash functions are implemented to see how.

stylewarning (changes requested):
Not all of the comments are required to fix in this PR. But the ones that are mistakes should be.

I also think there should be a couple unit tests written that check that derivatives are indeed properly calculated for a couple examples, and also that the laws of dual numbers are satisfied (i.e., that $\varepsilon^2 = 0$).

library/math/dual.lisp
eliaslfox: I don't think we should mix in a constant. We have lots of other cross type hash collisions.

stylewarning: Not doing it saves two or three assembly instructions.

Doing it makes debugging ever so marginally easier.

eliaslfox: It makes hash-tables asymptotically worse by reducing entropy in the hash. We [have discusses this previously](https://github.com/coalton-lang/coalton/pull/357#issuecomment-1006241282).

eliaslfox: If we do hash type information here, we should probably do it everywhere else for consistency.

gefjon: Given that we don't (intend to) support mixed-type hash tables, I see no reason to mix type ids into hashes. In dynamically typed contexts, it's valuable to make sure that e.g. the double-float `0d0` hashes differently from the single-float `0f0`, but in Coalton that doesn't matter; those values won't co-exist as keys of the same hash table. And @eliaslfox is right; every additional value you mix into a hash reduces the "weight" of each other value in the resulting hash, so mixing in a constant reduces overall entropy. I doubt that the entropy loss matters significantly for numeric types except for very large hash-tables, but there's no reason at all to mix in a constant, as compared to a small reason not to.

library/math/dual.lisp
gefjon: I would personally prefer "Dual numbers are <typeclass operator simple past tense> <on/by>" rather than "Note: <typeclass name> only works on." The current version is more succinct, but I'd rather have docstrings that read as complete sentences.

library/math/dual.lisp
stylewarning: 1. Coalton still lives inside of Common Lisp and is debugged inside of Common Lisp. I hope we don't forget that.

2. I don't understand why you believe that mixing a constant reduces entropy. Even if the hash were `hash(x)+C mod 2^n`, we still have the same information content, for we can always unmix this.

gefjon: Are we prepared to assume that `mix-hashes` is invertible? I know that `logxor` is, which I think we use as a fallback, and I haven't checked whether SBCL's is, but I know that for many hashing algorithms it is not. For example, I recently implemented FxHash, which uses a `wrapping-mul` and is therefore not invertible.

I could be doing my math wrong, but it seems to me that hashing a value of `N` bits into a hash of `H` bits captures `(/ H N)` bits of entropy from each bit of the value. Adding a constant of `C` bits means you instead capture `(/ H (+ N C))` bits from each bit of the value, which is less than `(/ H N)`.

stylewarning: I don't know about `mix-hashes` but I'm not advocating for any function in particular, just one that lets us make hash codes for different object types distinguished.

It's all very minor which is why I let go of it before, and I'm fine to let go of it now.

Jobhdez:
@eliaslfox could you please review this if you have time?

stylewarning (approved):
I think there are a few cosmetic and documentation fixes that need to be made, but I suggest we simplify, merge this PR, and I can follow on with another PR with some of these changes.

eliaslfox:
Closed in favor of #926.

Issue #877, opened 2023-04-06
Add missing tryinto instances for integer types to floats


eliaslfox:

Jobhdez:
Hello can someone please give more details on this one; is the goal here to convert integer types to floats? An example would be nice; thanks

stylewarning:
I also think more Dividable instances can be added for fixed integer types and single/doubles.

stylewarning:
> Hold off on this. I want reconsider our approach to division.

Im curious to hear your thoughts. Haskell, OCaml, Elm, are kind of sloppy about it imho; Rust is very explicit about casts/conversions/`From`; and none seem to natively admit that division is a field operation, not a ring-like one.

Probably not a discussion for this issue, but if we want to make division/numbers nicer within the context of this issue, then additional instances will be needed.

eliaslfox:
I think Rust has the right approach. I want to make all of addition, multiplication, subtraction, and division `Class :left :right :result`. Num would then be defined as something like:
`(Add :a :a :a) (Subtract :a :a :a) (Multiply :a :a :a) => Num :a`

The stdlib should only define operations in the form of `C :a :a :a`. Having instances to divide two integers and end up with a float is infrequently desired. If an alternative output type is desired, the inputs should be explicitly type cast first.

I also want to remove the `Div` class entirely. I don't find the distinction between integer and fractional division helpful.

PR #900, opened 2023-04-19
add tryinto instances for integer types to floats

Jobhdez:
this PR is about this [issue](https://github.com/coalton-lang/coalton/issues/877).

### DIFF
--- a/library/math/conversions.lisp
+++ b/library/math/conversions.lisp
@@ -182,5 +182,41 @@ cannot be represented in :TO. These fall into a few categories:
                  (Err "Integer to Double-Float conversion out-of-range")
                  (Ok y)))))))
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:defmacro integer-tryinto-float (integer lisp-float float pow)
+    `(define-instance (TryInto ,integer ,float)
+       (define (tryInto x)
+	 (lisp (Result String ,float) (x)
+           (cl:if (cl:and (cl:> x (cl:expt -2 ,pow)) (cl:< x (cl:expt 2 ,pow)))	 
+	          (cl:let ((y (cl:coerce x ',lisp-float)))
+	            (cl:if (cl:null y)
+		           (cl:error `,(cl:concatenate 'string (cl:symbol-name ,integer) "to " (cl:symbol-name ,float) " conversion out-of-range"))
+		           (Ok y)))
+	          (Err `,(cl:concatenate 'string "Given integer" "is not within " "(-2^" (write-to-string ,pow) ", " "2^" (write-to-string ,pow) ")"))))))))
+
+(coalton-toplevel
+  ;; Single Float
+  (integer-tryinto-float I64 cl:single-float Single-Float 24)
+
+  (integer-tryinto-float U64 cl:single-float Single-Float 24)
+
+  (integer-tryinto-float IFix cl:single-float Single-Float 24)
+
+  (integer-tryinto-float UFix cl:single-float Single-Float 24)
+
+  (integer-tryinto-float U32 cl:single-float Single-Float 24)
+
+  (integer-tryinto-float I32 cl:single-float Single-Float 24)
+
+;; Double Float
+  (integer-tryinto-float I64 cl:double-float Double-Float 53)
+
+  (integer-tryinto-float U64 cl:double-float Double-Float 53)
+
+  (integer-tryinto-float IFix cl:double-float Double-Float 53)
+
+  (integer-tryinto-float UFix cl:double-float Double-Float 53))
+
+
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/MATH/CONVERSIONS")


eliaslfox:
I don't think this is the right approach. All of these types have known bounds, the conversion code should just check if the input value is exactly representable in the output type. Also all of these conversions are infallible, so they should use `Into` not `TryInto`.  Looking at the [instances for `Double-Float`](https://coalton-lang.github.io/reference/#double-float-type) it looks like these already exist. 

The ones that are missing are `TryInto` from `I64`, `U64`, `IFix`, and `UFix`. It's worth noting that we don't want `TryInto` conversions to lose precision, so if a value is within a floats range, but outside the safe range the conversion should fail.

Jobhdez:
> I don't think this is the right approach. All of these types have known bounds, the conversion code should just check if the input value is exactly representable in the output type. Also all of these conversions are infallible, so they should use `Into` not `TryInto`. Looking at the [instances for `Double-Float`](https://coalton-lang.github.io/reference/#double-float-type) it looks like these already exist.
> 
> The ones that are missing are `TryInto` from `I64`, `U64`, `IFix`, and `UFix`. It's worth noting that we don't want `TryInto` conversions to lose precision, so if a value is within a floats range, but outside the safe range the conversion should fail.

@eliaslfox should I start working on this again?

eliaslfox:
If you want to. The issue is still open.

Does what I said about how to proceed make sense?

Jobhdez:
> If you want to. The issue is still open.
> 
> Does what I said about how to proceed make sense?

so I should do TryInto from I64, U64, IFix, and UFix?

but I dont get this quote “so if a value is within a floats range, but outside the safe range the conversion should fail”.

eliaslfox:
Floats have a range of values they can represent, but only some of that range represents numbers exactly. Larger values (or larger negative values) are imprecise. TryInto shouldn't lose precision, so when converting a value a float, if the values is outside the float's "safe range" then the conversion should fail.

gefjon:
I could be misremembering, but I think the safe range for f64 (our `Double-Float`) is (-2^53 ..= 2^53), and for f32 (our `Single-Float`) (-2^24 ..= 2^24).

Jobhdez:
> Floats have a range of values they can represent, but only some of that range represents numbers exactly. Larger values (or larger negative values) are imprecise. TryInto shouldn't lose precision, so when converting a value a float, if the values is outside the float's "safe range" then the conversion should fail.

Thanks. Can you provide a link to an example where there’s a test to check whether the values are outside the safe range? Thanks

eliaslfox:
I think @gefjon is right about the ranges. 

Just check that the value is within (-2^53, 2^53) before coercing to a double float.

The stack overflow answer will give confusing results because some values will coerce to floats, but that value plus or minus one won't. This is because some integers above 2^53 can be represented exactly as a double float, but not all of them.

See [floating point](https://en.m.wikipedia.org/wiki/Floating-point_arithmetic) for more information.

Also the Coalton code you linked should probably be updated to check integer ranges directly. I'm not sure if erroring when a value is out of range in coerce is guaranteed by the standard.

Jobhdez:
> I think @gefjon is right about the ranges.
> 
> Just check that the value is within (-2^53, 2^53) before coercing to a double float.
> 
> The stack overflow answer will give confusing results because some values will coerce to floats, but that value plus or minus one won't. This is because some integers above 2^53 can be represented exactly as a double float, but not all of them.
> 
> See [floating point](https://en.m.wikipedia.org/wiki/Floating-point_arithmetic) for more information.
> 
> Also the Coalton code you linked should probably be updated to check integer ranges directly. I'm not sure if erroring when a value is out of range in coerce is guaranteed by the standard.

@eliaslfox want to take a look at my changes? let me know if I have to make any changes please. thanks

library/math/conversions.lisp
```diff
@@ -182,5 +182,32 @@ cannot be represented in :TO. These fall into a few categories:
...
+  `(coalton-toplevel
+     (define-instance (TryInto ,integer ,float)
+       (define (tryInto x)
+         (lisp (Result String ,float) (x)
+	   (cl:if (cl:and (cl:> x (cl:expt -2 ,pow)) (cl:< x (cl:expt 2 ,pow)))	 
+	          (cl:let ((y (cl:ignore-errors (cl:coerce x (cl:quote ,lisp-float)))))
+	            (cl:if (cl:null y)
+		           (Err (cl:concatenate 'string ,integer-name "to " ,float-name " conversion out-of-range"))
+		           (Ok y)))
+		  (Err (cl:concatenate 'string "Given integer" is not within "(-2^" (write-to-string ,pow) ", " "2^" (write-to-string ,pow) ")"))))))))
```
eliaslfox, line 195: Write this as `(Err ,(cl:concatenate ...))` so that it runs during macro expansion.

library/math/conversions.lisp
```diff
@@ -182,5 +182,32 @@ cannot be represented in :TO. These fall into a few categories:
                  (Err "Integer to Double-Float conversion out-of-range")
                  (Ok y)))))))
 
+(cl:defmacro integer-tryinto-float (integer integer-name lisp-float float float-name pow)
+  `(coalton-toplevel
+     (define-instance (TryInto ,integer ,float)
+       (define (tryInto x)
+         (lisp (Result String ,float) (x)
+	   (cl:if (cl:and (cl:> x (cl:expt -2 ,pow)) (cl:< x (cl:expt 2 ,pow)))	 
+	          (cl:let ((y (cl:ignore-errors (cl:coerce x (cl:quote ,lisp-float)))))
```
eliaslfox, line 191: Don't ignore errors here. The conversion should never fail, and if it does we want to know.

library/math/conversions.lisp
```diff
@@ -182,5 +182,32 @@ cannot be represented in :TO. These fall into a few categories:
                  (Err "Integer to Double-Float conversion out-of-range")
                  (Ok y)))))))
 
+(cl:defmacro integer-tryinto-float (integer integer-name lisp-float float float-name pow)
```
eliaslfox, line 185: Have the macro generate a `define-instance` without a  `coalton-toplevel`. Then call the macro inside a toplevel like:
```
(coalton-toplevel
  (integer-tryinto-float ...)
  ...)
```
This allows the Coalton compiler to gives better error messages.

library/math/conversions.lisp
```diff
@@ -182,5 +182,32 @@ cannot be represented in :TO. These fall into a few categories:
...
+	          (cl:let ((y (cl:ignore-errors (cl:coerce x (cl:quote ,lisp-float)))))
+	            (cl:if (cl:null y)
+		           (Err (cl:concatenate 'string ,integer-name "to " ,float-name " conversion out-of-range"))
+		           (Ok y)))
+		  (Err (cl:concatenate 'string "Given integer" is not within "(-2^" (write-to-string ,pow) ", " "2^" (write-to-string ,pow) ")"))))))))
+	   
+
+  
+;; Single Float
+(integer-tryinto-float I64 "I64" cl:single-float Single-Float "Single-Float" 24)
```
eliaslfox, line 200: You don't need to pass the type names as strings. Symbols can be converted to strings with `cl:symbol-name`.

eliaslfox (commented):
A couple minor things. Also single-floats should have `TryInto U32|I32 Single-Float`.

Jobhdez:
> A couple minor things. Also single-floats should have `TryInto U32|I32 Single-Float`.

@eliaslfox Ok I have addressed your review

eliaslfox (commented):
A couple minor things

eliaslfox:
Also you need to wrap the macro definition in an `eval-when` like [this](https://github.com/coalton-lang/coalton/blob/main/library/math/num.lisp#L42-L48) because otherwise ccl and allegro get confused.

eliaslfox (approved):
Looks good

eliaslfox:
It looks like there's merge conflicts. Resolve them locally with `git rebase main`, then I can merge this.

Jobhdez:
> Looks good

@eliaslfox do i update the branch or do you take over from here?

eliaslfox:
Don't update with the web ui, it'll create a merge commit. It needs to be done locally with `git rebase` to preserve a linear commit history.

Jobhdez:
> It looks like there's merge conflicts. Resolve them locally with `git rebase main`, then I can merge this.

do i do this from my current branch `patch-7` or from main?

eliaslfox:
from `patch-7`

eliaslfox:
Also you should squash your commits to a single commit.

Jobhdez:
> Also you should squash your commits to a single commit.

@eliaslfox how do I squash them? ive been trying and cant

Jobhdez:
@eliaslfox hey i think the other day robert told me that he could squash the commits himself. can you squash the commits? i need to learn some git. i wasnt able to rebase it either.

Jobhdez:
@eliaslfox ok i have successfully rebased -- ie resolved all the conflicts manually. i was having trouble but after eating some chips i was able to to do it

Jobhdez:
hey so i dont think theres merge conflicts now. the other day robert said he could squash the commits on his end. could you squash the commits? im fighting git right now. it seems like an endless cycle

eliaslfox:
Yeah I can squash it tomorrow.

Jobhdez:
> Yeah I can squash it tomorrow.

thanks. there shouldnt be merge conflicts. i resolved all the conflicts manually

PR #908, opened 2023-04-25
Refactor library/math/num.lisp

eliaslfox:
Includes the refactors from #905 without the breaking changes.

### DIFF
--- a/library/math/num.lisp
+++ b/library/math/num.lisp
@@ -1,6 +1,6 @@
 ;;;; num.lisp
 ;;;;
-;;;; Num instances for primitive numerical types
+;;;; Instances for primitive numerical types
 
 (coalton-library/utils:defstdlib-package #:coalton-library/math/num
   (:use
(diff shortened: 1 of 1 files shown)


colescott (approved):
This looks great! I'm definitely coming around to macros in the standard library as long as they appear within `coalton-toplevel` forms

library/math/num.lisp
```diff
@@ -79,139 +88,137 @@
...
+   (cl:dpb 0 (cl:byte (cl:1- bits) 0) x)))
+
+(cl:defmacro %define-overflow-handler (name bits)
+  `(cl:progn
+     (cl:declaim (cl:inline ,name))
+     (cl:defun ,name (value)
+       (cl:typecase value
+         ((cl:signed-byte ,bits) value)
+         (cl:otherwise
+          (cl:cerror "Continue, wrapping around."
```
gefjon, line 149: Any consideration to, in release mode, eliding the test and the cerror, and just wrapping unconditionally?

eliaslfox: I think that would just hide bugs.

PR #910, opened 2023-04-26
Add default typeclass

macrologist:
A draft of a Default typeclass. More standard types could be added. Making this draft seeking comments.


eliaslfox:
This looks good.  Default doesn't need its own file, it can just go in `library/classes.lisp`. Instances for types should go in the files for those types. 

Also `Vector`, `Hash-Table`, `Cell`, and tuples should also implement `Default`.

stylewarning:
What was the rationale for making `default` a function and not a value?

gefjon:
> What was the rationale for making default a function and not a value?

I can't speak to @macrologist 's thinking, but I'll point out that if `default` is a value, then `HashTable`, `Vector` and `Cell` cannot have meaningful instances.

stylewarning:
What's not meaningful about `(new default)`, etc.? That those kinds of things cause trouble with compilation/evaluation or something?

stylewarning:
That is too say

```
(define-instance (Default :a => Default (Cell :a)) 
  (define default (new default)))
```

will cause some evaluation issue that

```
(define-instance (Default :a => Default (Cell :a)) 
  (define (default) (new (default))))
```

doesn't?

stylewarning:
Sorry for the train of thought.

I guess you mean that these are mutable objects. Not that they're not meaningful, but rather that it's dangerous and not useful.

If that's what you meant, then I might suggest we rename it to be explicit about making a new default: like `make-default` (though we use that nomenclature for construction macros, which we probably ought to change).

gefjon:
Also, we should think about how to deal with existing operators that talk about defaulting, like `with-default` on `Unwrappable`. Should we rename that function to `unwrap-or`, and have `with-default` (or possibly `unwrap-or-default`) return the `default` inner type?

macrologist:
> Also, we should think about how to deal with existing operators that talk about defaulting, like `with-default` on `Unwrappable`. Should we rename that function to `unwrap-or`, and have `with-default` (or possibly `unwrap-or-default`) return the `default` inner type?

I would favor something like `unwrap-or` or your suggested `upwrap-or-default` over changing the definition of `with-default`.

macrologist:
> I'd seriously consider the possibility that `default` should be a value and that inherently mutable types should not have `Default` instances. I can't imagine any useful polymorphism coming from being able to construct an empty hash-table, an empty vector, `Nil` and `0` all from the same interface. It seems to me that algorithms which use mutation will always be pretty closely tied to the specific type. (If someone has a counterexample, I could be convinced otherwise.)

There could easily be another class to produce "empty" mutable things. 
```
   (define-class (Fresh :a)
       (define fresh (Unit -> :a)))
```
Or something.

My initial impulse was to have `default` be a value, but at the suggestion of @eliaslfox I altered it.

eliaslfox:
I don't want to have multiple type classes for constructing default versions of types.

Bifurcating default construction would require duplicating helper methods and instances. Do we now need `fresh-unwrap`? Does every type that implements `Default` also need to implement `Fresh`?

None of this seems worth being able to sometimes write `default` instead of `(default)`.

macrologist:
> Does every type that implements `Default` also need to implement `Fresh`?

My thinking is no. The two concepts, while potentially compatible, would be conventionally distinct. One for immutable values - something with an easy-to-write-down literal representation - and the other for things like `Cell` and `Vector`.  But if "don't do it because that's not what *we do here*" is not a good enough barrier to mis-instantiation of `Fresh` or `Default` on the wrong data types, I'm cool with dispensing with the distinction and landing on `(default)`.

Echoing @gefjon , I also do not see a *great* utility default `HashMap`s, but it can't really hurt to include implementations for them.

library/classes.lisp
```diff
@@ -324,7 +325,20 @@ Typical `fail` continuations are:
...
+  ;; Default
+  ;;
+
+  (define-class (Default :a)
+    (default (Unit -> :a)))
+
+  (declare defaulting-unwrap ((Unwrappable :container) (Default :element) =>
+                              (:container :element) -> :element))
+  (define (defaulting-unwrap container)
+    (with-default (default) container)))
```
eliaslfox, line 341: One last thing. `defaulting-unwrap` should call `unwrap` directly so that the default value is only constructed if needed.

library/classes.lisp
```diff
@@ -324,7 +325,20 @@ Typical `fail` continuations are:
...
-                    container)))
+                    container))
+
+
+  ;;
+  ;; Default
+  ;;
+
+  (define-class (Default :a)
+    (default (Unit -> :a)))
```
eliaslfox, line 336: Also `Default` the class and `defaulting-unwrap` should have docstrings.

eliaslfox (approved):

PR #918, opened 2023-05-05
Add warning for underapplied values

### LINKED ISSUE #882: Re-add warning for discarded partially applied values

eliaslfox:
Fixes #882

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -90,6 +90,7 @@
                 :serial t
                 :components ((:file "pattern-exhaustiveness")
                              (:file "unused-variables")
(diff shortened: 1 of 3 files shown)


stylewarning:
Can you write an example and update the documentation somewhere?

eliaslfox:
I don't think it makes sense to document this in the intro doc. It's not a type system behavior, the compiler is just trying to warn about something that might be a bug. Also the warning message is fairly self explanatory.

stylewarning:
What about the style guide then?

colescott (approved):

PR #923, opened 2023-05-19
deprecated function/typo in intro

Izaakwltn:
Just a typo/update fix, function `(make-vector)` no longer exists and has been replaced by `(coalton-library/vector:new)`


eliaslfox (approved):
Thanks for the fix

PR #927, opened 2023-05-22
improve and extend readme example

stylewarning:
I extended the README example with some more of Coalton's features, and included a glamor shot of a type error.

This is a mere (constructive) suggestion. I'm OK if we prefer to keep the README simpler in style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,52 +18,87 @@ Coalton can be written in files:
 ```lisp
 (in-package #:coalton-user)
 
+(named-readtables:in-readtable coalton:coalton)
+
 (coalton-toplevel
-  (define-type Symbol
-    (Symbol String))
+  ;; Define Coalton `Symbol`s as Lisp `cl:keyword`s.
+  (repr :native cl:keyword)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,52 +18,87 @@ Coalton can be written in files:
...
+  ;; Define Coalton `Symbol`s as Lisp `cl:symbol`s.
+  (repr :native cl:symbol)
+  (define-type Symbol)
+
+  ;; Bind a Lisp function into Coalton.
+  (declare sym (String -> Symbol))
+  (define (sym s)
+    "Create a new symbol named `s`."
+    (lisp Symbol (s)
+      (cl:intern s)))
```
eliaslfox, line 33: Could this be something like `(cl:intern s :keyword)`. A bare `inern` in example code makes me nervous.

eliaslfox (approved):
Looks good

README.md
stylewarning: yeah that sounds reasonable; done

PR #928, opened 2023-05-22
Add `vector:clear!`

eliaslfox:
Also rename `vector:set-capacity` to `vector:set-capacity!`.

### DIFF
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -18,7 +18,8 @@
    #:element-type
    #:empty?
    #:copy
-   #:set-capacity
+   #:set-capacity!
+   #:clear!
    #:push!
    #:pop!
    #:pop-unsafe!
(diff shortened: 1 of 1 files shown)


library/vector.lisp
```diff
@@ -114,6 +115,11 @@
              (cl:adjust-array v new-capacity :fill-pointer cl:t))
       Unit))
 
+  (declare clear! (Vector :a -> Unit))
+  (define (clear! v)
+    "Set the capity of V to 0."
```
stylewarning, line 120:
```suggestion
    "Set the capacity of V to 0."
```

library/vector.lisp
```diff
@@ -102,10 +103,10 @@
     (lisp (Vector :a) (v)
       (alexandria:copy-array v)))
 
-  (declare set-capacity (UFix -> Vector :a -> Unit))
-  (define (set-capacity new-capacity v)
+  (declare set-capacity! (UFix -> Vector :a -> Unit))
+  (define (set-capacity! new-capacity v)
     "Set the capacity of V to NEW-CAPACITY. Setting the capacity to lower then the length will remove elements from the end."
-    (let shrinking = (< new-capacity (length v)))
+    (let shrinking = (> new-capacity (length v)))
```
stylewarning, line 109: I would write this as:

```
    (let shrinking = (< new-capacity (length v)))
    (lisp Unit (v shrinking new-capacity)
      ;; If the array is getting larger then dont change the
      ;; fill pointer
      (cl:adjust-array v new-capacity :fill-pointer shrinking)
```

Right now, the variable `shrinking` doesn't match the behavior (shrinking is True when it's actually expanding), though the code that follows it is correct.

stylewarning (changes requested):

stylewarning (approved):

PR #930, opened 2023-05-23
Added Nth-cdr function, and corresponding test

Izaakwltn:
Added nth-cdr, modeled after the index definition.

(Please ignore the aforementioned "'intercalatexf' typo change",  that was a personal emacs keyboard snafu)


library/list.lisp
```diff
@@ -234,6 +235,16 @@
     "Like INDEX, but errors if the index is not found."
     (from-some "There is no NTH" (index n l)))
 
+  (declare nth-cdr (Ufix -> List :a -> (Optional (List :a))))
```
stylewarning, line 238: To be consistent with CDR this ought to just do CDR at most N times, shorting out when it reaches the empty list. 

Alternatively, to be consistent with NTH, we could error on (nth-cdr n Nil) for n>0. I don't recommend it though.

stylewarning (changes requested):

Izaakwltn:
I ran into an error using 'decleration' so I found and fixed all uses of 'decleration' instead of 'declaration' in the codebase

library/list.lisp
```diff
@@ -234,6 +235,15 @@
     "Like INDEX, but errors if the index is not found."
     (from-some "There is no NTH" (index n l)))
 
+  (declare nth-cdr (Ufix -> List :a -> List :a))
+  (define (nth-cdr n l)
+    "Returns the nth-cdr of a list."
+    (cond ((null? l)
```
stylewarning, line 241: style: put condition on new line

library/list.lisp
```diff
@@ -234,6 +235,15 @@
...
+	   nil)
```
stylewarning, line 242: style: capitalize Nil

library/list.lisp
```diff
@@ -234,6 +235,15 @@
...
+	  ((== 0 n)
```
stylewarning, line 243: style: use `zero?`

library/list.lisp
```diff
@@ -234,6 +235,15 @@
...
+	   l)
+	  (True (nth-cdr (- n 1) (cdr l)))))
```
stylewarning, line 245: style: use `1-`

stylewarning (approved):
only minor style things

in general the typo fixes should remain in a separate commit. will you be rebasing your commits to clean the history?

PR #932, opened 2023-05-23
Fix warnings

eliaslfox:


stylewarning (approved):
we should add an `unreachable` function to Coalton

PR #935, opened 2023-05-23
Added nthcdr, corresponding tests

Izaakwltn:


eliaslfox (approved):
Looks good. This needs to be rebased from main, and then squashed.

PR #939, opened 2023-05-24
Add structs with accessors

eliaslfox:
Adds the ability to define structs:

```lisp
(coalton-toplevel
  (define-struct Point
    (x Integer)
    (y Integer)))
```

Defining a struct is equivalent to defining a single constructor ADT. Structs can also use the new accessor syntax to get individual fields.

```lisp
(coalton-toplevel
  (define p (Point 1 2))

  (define x (.x p))

  (define y (.y p)))
```

Accessors can be passed by value like functions:

```lisp
(coalton (map .x (make-list (Point 1 2) (Point 3 4) (Point 5 6))))
```

Accessors must access a field from a single unambiguous struct type. The following are invalid:

```lisp
(coalton-toplevel
  (define (f p)
    (.x p)) ;; the type of p is unknown

  (define-struct Point3
    (x Integer)
    (y Integer)
    (z Integer))
    
  (define g
    (let h = .x)
    (h (Point 1 2))
    (h (Point3 1 2 3))) ;; h is used on different structs

  (define i (.x "q")) ;; accessors can only be used on structs

  (define j (.q (Point 1 2)))) ;; accessors must be for valid fields
```

Structs can have type variables:

```lisp
(coalton-toplevel
  (define-struct (ThreeOfThem :a)
    (the-first :a)
    (the-second :a)
    (the-third :a)))
```

## Still TODO:
- [x] updating documentation
- [x] updating the doc generator to display struct fields

## Future work:
* Add pattern matching and constructors that are named rather than positional:
```lisp
(match p
  ((Point { x y }) (+ x y)))
```
* Add a syntax for immutable partial updates
* Add mutable struct fields

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -51,6 +51,7 @@
                              (:file "collect")
                              (:file "renamer")
                              (:file "binding")
+                             (:file "type-definition")
                              (:file "package")))
                (:module "runtime"
                 :serial t
@@ -76,6 +77,7 @@
                              (:file "traverse")
                              (:file "toplevel")
                              (:file "binding")
+                             (:file "accessor")
                              (:file "partial-type-env")
                              (:file "parse-type")
                              (:file "define-type")
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -14,7 +14,9 @@ When making a new project, you will want to define a new package and establish y
 
 ```lisp
 (defpackage #:my-package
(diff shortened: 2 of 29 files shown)


eliaslfox:
Should we allow fields to be numbers? Then accessors for tuples could be `.0`, `.1`, etc.

colescott (approved):

PR #941, opened 2023-05-24
Add `Seq`, a new persistent sequence type

stylewarning:
Seq is implemented as a portable, stand-alone Common Lisp library. This has been put into a new folder
src/persistent-data-structures/, anticipating future such data structures to be implemented there.

The performance of Seqs is about 5-10x slower than Vectors at this time. There is a lot of opportunity for optimization, since there's no use of type declarations.

TODO:

- [ ] Get agreement on the name of this data structure in Coalton.
- [ ] Write Lisp and Coalton tests.
- [ ] Get feature parity with existing sequence functions.
- [ ] Implement additional type classes.
- [ ] Run some performance benchmarks.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -132,6 +132,7 @@
                       (funcall compile)))
   :depends-on (#:coalton/compiler
                #:coalton/hashtable-shim
+               #:coalton/persistent-data-structures
                #:trivial-garbage
                #:alexandria)
   :pathname "library/"
@@ -166,6 +167,7 @@
                (:file "tuple")
                (:file "list")
                (:file "vector")
+               (:file "seq")
                (:file "char")
                (:file "string")
                (:file "slice")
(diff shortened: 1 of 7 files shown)


macrologist:
As regards the name. If `Seq` seems too presumptuous, what's wrong with `PVec`?

stylewarning:
> As regards the name. If `Seq` seems too presumptuous, what's wrong with `PVec`? 

A presumptuous choice might be good if we want to encourage it to be a quasi-default. A reason against "vector" is that that name is entrenched as meaning a linear array of memory.

gefjon:
I would consider moving mutable data structures into a subdir `mutable`, and having persistent data structures in a subdir `persistent`, with package names mirroring the fs, so that this is `coalton-library/persistent/vec:Vec`. And ideally updating the prelude to refer to the persistent structures.

[RRB-trees](https://github.com/clojure/core.rrb-vector) might be a better data structure as a general-purpose persistent vector than bit-partitioned binary tries with tail buffers, since they have much better concatenation at the cost of slightly worse indexing.

I have a Common Lisp implementation of bit-partitioned binary tries with tail buffers, and also hash array-mapped tries, at [https://github.com/gefjon/immutable/], which is reasonably optimized. I don't foresee having the time to finish and publish the library in the near future, but I'd be happy to add a maintainer, or have it moved in-tree into Coalton. I haven't benchmarked, but I suspect that the absence of generic functions and the presence of type declarations in my version lead to better performance.

PR #946, opened 2023-05-25
Add `vector:with-initial-element`

### LINKED ISSUE #876: Add missing stdlib functions
* create a vector with an initial size and element
* extend a vector with an iterator
* string views (like slices but for strings)

eliaslfox:
Fixes #876

### DIFF
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -13,6 +13,7 @@
    #:Vector
    #:new
    #:with-capacity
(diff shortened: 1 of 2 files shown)


stylewarning (approved):
i'm not too much of a fan of the WITH- naming; Lispers are used to that being a "RAII"-style macro.

eliaslfox:
I don't mind renaming it.

PR #947, opened 2023-05-26
Added iterator tests: for-each!,  count!, and find!

Izaakwltn:


tests/iterator-tests.lisp
```diff
@@ -178,14 +178,35 @@
...
+    (is (== (iter:collect! (iter:into-iter (populate-vector)))
+	    (make-list 0 1 2 3 4)))))
+
+(define-test iter-count ()
+  (is (== (iter:count! (iter:into-iter (make-list 1 2 3 4 5)))
+	  5)))
+
+(define-test iter-find ()
+  (let ((even-in-iter (fn (iter)
+			(iter:find! (fn (x) (coalton-library/math/arith:zero? (mod x 2)))
```
eliaslfox, line 198: This should be written as `math:zero?`

eliaslfox (approved):
Thanks for adding tests

tests/iterator-tests.lisp
eliaslfox: All of the `math/*` packages get pulled together [here](https://github.com/coalton-lang/coalton/blob/main/library/prelude.lisp#L5-L16).

stylewarning (commented):
On mobile GitHub, indentation is still showing as being incorrect. @Izaakwltn

eliaslfox:
I missed that. Maybe we should add a CI rule to ban tab characters from the repository.

PR #948, opened 2023-05-26
Added iter:flat-map! and corresponding test

Izaakwltn:


eliaslfox:
The type of flat-map should be `(:a -> Iterator :b) -> Iterator :a -> Iterator :b`

eliaslfox (approved):

PR #950, opened 2023-05-28
String functions First and Rest, and tests

Izaakwltn:
I thought these might be useful substring wrappers, though they probably cater to less functional approaches.


library/string.lisp
```diff
@@ -64,6 +66,16 @@
       (lisp String (real-start real-end str)
         (cl:subseq str real-start real-end))))
 
+  (declare first (String -> String))
+  (define (first str)
+    "Returns the first character of a String."
```
stylewarning, line 71: Technically this doesn't return a character.

Maybe merge these functions and have

```
(define (split n s)
  (Tuple (substring s 0 n)
         (substring s n (length s))))
```

?

Izaakwltn: That sounds good to me, I switched those out.

eliaslfox (approved):

PR #956, opened 2023-06-08
Fix tree:make macro

tojoqk:
I've identified and fixed an issue in the tree:make macro. It was mistakenly expanding to use a non-existent `collect-tree!` function, when it should be using `collect!`.

### DIFF
--- a/library/ord-tree.lisp
+++ b/library/ord-tree.lisp
@@ -534,4 +534,4 @@ B'. Which one is chosen for the result is undefined."
   "Construct a tree containing the ELEMENTS.
 
 e.g. (tree:make 5 6 1 8 9) => tree containing 1, 5, 6, 8, 9."
-  `(collect-tree! (iter:into-iter (make-list ,@elements))))
+  `(collect! (iter:into-iter (make-list ,@elements))))


eliaslfox (approved):
Thanks for the fix

library/ord-tree.lisp
```diff
@@ -534,4 +534,4 @@ B'. Which one is chosen for the result is undefined."
   "Construct a tree containing the ELEMENTS.
 
 e.g. (tree:make 5 6 1 8 9) => tree containing 1, 5, 6, 8, 9."
-  `(collect-tree! (iter:into-iter (make-list ,@elements))))
+  `(collect! (iter:into-iter (make-list ,@elements))))
```
eliaslfox, line 537: This should probably be:
```
(the (Tree :elt) (collect! ...))
```

So that it always collects to a tree.

tojoqk: This `collect!` symbol belongs to `coalton-library/ord-tree` package, it is not `iter:collect!`.

Definition of `coalton-library/ord-gree:collect!` is as follows:
https://github.com/coalton-lang/coalton/blob/5ff5179d07b80a142bbcd17894c98c90646bc98f/library/ord-tree.lisp#L493-L498

So, in the current implementation, it always collects to a tree.

PR #957, opened 2023-06-08
Add: delete-if to library/list

macrologist:
B/c sometimes you don't want to traverse the whole list via `filter`

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -41,8 +41,8 @@
    #:intersection
    #:lookup
    #:remove-duplicates
-   #:delete-if
-   #:delete
+   #:remove-if
+   #:remove
    #:difference
    #:zipWith
    #:zipWith3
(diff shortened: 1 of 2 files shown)


gefjon:
Common Lisp uses `remove-if` for the immutable version of this operator, and `delete-if` for the version which may destructively modify the list. Coalton should probably do the same.

macrologist:
The suggestion makes sense but is more radical than it seems if we care at all about consistency. Assuming we do, we'd also have to rename `delete` to `remove`, which is a breaking change (in the fantasy world where people are using Coalton "out there"). I'm not sure cohesion with CL is "worth it", but I'm open to further arguments. 

My addition was proposed b/c I had a direct need for such a function this morning; @gefjon 's concern make sense, but in order to take it seriously, we must begin to consider this "CL convention parity" issue in the wider context of the whole standard libarary. This might be smell test for a wider technical/culture debt issue. It might also just be frivilous bikeshedding 🤷

macrologist:
Oh to be clear, I'm fine with just letting this PR die and just using my own private utility library that contains all these little "missing functions" that I'd like to use.

eliaslfox:
I'm in favor of renaming `delete` to `remove` and `delete-if` to `remove-if`.

macrologist:
Cool, so renamed.

eliaslfox (approved):

PR #969, opened 2023-07-13
Add: impossible macro

macrologist:
Pattern matches on ADTs often require matching on cases that are never intended to be encountered in reality, but would fail exhaustiveness checks if not matched. The `impossible` macro produces an expression of type `:a` and signals an error at runtime.


stylewarning:
bike shed: I prefer "unreachable" but I'm not married to it.

macrologist:
ahh unreachable already exists

PR #971, opened 2023-07-14
Rrb tree pseq

macrologist:
@stylewarning @eliaslfox @colescott  

This PR adds a Relaxed Radix Balanced Tree implementation of persistent sequences to the coalton standard library. 

The rebalancing step attempts to minimize memory allocation, reusing already allocated contiguous memory whenever possible.

### TODO

- [x] Need to fix a bug in the radix search (change +1 to linear search) 
- [x] Test that branching factor invariants are preserved in various cases 
- [x] More typeclass impls (e.g. Eq)

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -1,0 +1,559 @@
+(coalton-library/utils:defstdlib-package #:coalton-library/seq
+  (:use
+   #:coalton
+   #:coalton-library/builtin
+   #:coalton-library/functions
+   #:coalton-library/classes)
+  (:local-nicknames
+   (#:types #:coalton-library/types)
+   (#:math #:coalton-library/math/integral)
+   (#:optional #:coalton-library/optional)
+   (#:cell #:coalton-library/cell)
+   (#:vector #:coalton-library/vector)
+   (#:iter #:coalton-library/iterator))
+  (:export
+   #:Seq
+   #:new
+   #:push
+   #:pop
+   #:size
+   #:get
+   #:put
+   #:empty?
+   #:conc
+   #:make))
+
+(in-package #:coalton-library/seq)
+
+(named-readtables:in-readtable coalton:coalton)
+
+#+coalton-release
+(cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
+
+;;;;
+;;; This module implements a persistent sequence data type based on
+;;; Relaxed Radix Balanced Trees, introduced by Phillip Bagwell
+;;; See: 
+;;; https://www.semanticscholar.org/paper/RRB-Trees%3A-Efficient-Immutable-Vectors-Bagwell-Rompf/30c8c562f6421ab6b00d0b7faebd897c407de69c
+;;; 
+
+;;; Private utility macro for convenient matching of a single case of a ADT
+(cl:eval-when (:compile-toplevel :load-toplevel :execute)
+  (cl:defmacro match1 (pattern form cl:&body body)
+    `(match ,form
+       (,pattern ,@body)
+       (_ (unreachable)))))
+
+(coalton-toplevel 
+  ;;
+  ;; Seq
+  ;; 
+  
+  (declare max-branching UFix)
+  (define max-branching 32)
+
+  (declare min-branching UFix)
+  (define min-branching 24)
+
+  (define-type (Seq :a)
+    (RelaxedNode
+     UFix                               ; height
+     UFix                               ; cached full subtree size
+     (vector:Vector UFix)               ; cumulative size table
+     (vector:Vector (Seq :a)))         ; subtrees
+    (LeafArray (vector:Vector :a)))
+
+  (declare new (types:RuntimeRepr :a => Unit -> Seq :a))
+  (define (new)
+    "Create a new empty `Seq`."
+    (LeafArray (vector:new)))
+
+  (declare size (Seq :a -> UFix))
+  (define (size seq)
+    "Return the number of elements in the `seq`."
+    (match seq
+      ((RelaxedNode _ _ cst _) (vector:last-unsafe cst))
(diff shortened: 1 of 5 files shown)


stylewarning:
Note: This would supersede my Seq PR.

eliaslfox (approved):
This looks really good.

The package should be added to [this list](https://github.com/coalton-lang/coalton/blob/main/src/doc/generate-documentation.lisp#L101-L135) so that it shows up in the generated documentation.

Approving but waiting on @colescott and @stylewarning's reviews before merging.

library/pseq.lisp
```diff
@@ -0,0 +1,513 @@
...
+    (define <> conc))
+  
+  (define-instance (iter:IntoIterator (PSeq :a) :a)
+    (define (iter:into-iter seq)
+      (match seq
+        ((LeafArray v) (iter:into-iter v))
+        ((RelaxedNode _ _ _ subs)
+         (fold iter:chain! iter:empty (map iter:into-iter subs))))))
+
+  (define-instance (Functor PSeq)
```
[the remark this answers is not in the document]
macrologist: I can't seem to do it.  

`RuntimeRepr :a` evidently needs to be satisfied for `PSeq :a` but I don't know how to say that the constraint needs to be satisfied for the arguments of the type constructor `PSeq`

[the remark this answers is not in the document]
macrologist: (I imagine that this is part of the reason that `Vector` isn't a `Monad`?)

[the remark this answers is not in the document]
eliaslfox: Oh yeah, nvm then.

stylewarning:
Bike Shed (TM): Could I suggest just the name "Seq"? Matches Clojure, also "p" can be interpreted as "parallel" (cf. plet pdotimes etc).

stylewarning:
Bike Shed (TM): Could I suggest just the name "Seq"? Matches Clojure, also "p" can be interpreted as "parallel" (cf. plet pdotimes etc).

macrologist:
if @colescott and @stylewarning have any (more) complaints, lemme know. Otherwise @eliaslfox (or whomever) can merge.

library/seq.lisp
```diff
@@ -0,0 +1,517 @@
...
+      ;;make a node from rebalanced branches
+      (make-node-from-branches
+       (iter:collect!
+        (iter:filter! (compose (< 0) size)
+                      (map branch-rebalancer
+                           (iter:up-through stop))))))))
+
+(cl:defmacro make (cl:&rest elems)
+  "Create a new Seq containing ELEMS."
+  `(fold push (new) (make-list ,@elems)))
```
stylewarning, line 514: I would avoid consing the list. Maybe just expand into a chain of PUSH, to match how MAKE-LIST works?

library/seq.lisp
```diff
@@ -0,0 +1,517 @@
...
+       (,pattern ,@body)
+       (_ (unreachable)))))
+
+(coalton-toplevel 
+  ;;
+  ;; Seq
+  ;; 
+  
+  (declare max-branching UFix)
+  (define max-branching 32)
```
stylewarning, line 53: I wish we had a way to define constants in Coalton.

library/seq.lisp
```diff
@@ -0,0 +1,517 @@
...
+    (match seq
+      ((RelaxedNode _ _ cst _) (vector:last-unsafe cst))
+      ((LeafArray leaves) (vector:length leaves))))
+  
+  (define (empty? seq)
+    (== 0 (size seq)))
+
+  (declare get (Seq :a -> UFix -> Optional :a))
+  (define (get seq idx)
+    "Get the member of SEQ at index IDX, or None if IDX is larger
```
stylewarning, line 83: Coalton docstring style is to use markdown syntax and lowercase names

    Get the member of `seq` at ...

library/seq.lisp
```diff
@@ -0,0 +1,517 @@
...
+     UFix                               ; height
+     UFix                               ; cached full subtree size
+     (vector:Vector UFix)               ; cumulative size table
+     (vector:Vector (Seq :a)))         ; subtrees
+    (LeafArray (vector:Vector :a)))
+
+  (declare new (types:RuntimeRepr :a => Unit -> Seq :a))
+  (define (new)
+    "Create a new empty Seq."
+    (LeafArray (vector:new)))
```
stylewarning, line 69: If we get rid of the runtimerepr stuff, we should make this LOAD-TIME-VALUEd

stylewarning (approved):
My only grievance is the impl of MAKE. Eventually it would be nice to "statically" optimize MAKE and other conversion functions to allocate the tree structure wholesale without intermediate consing—but not for this PR.

I'm fine if we punt changing the definition of MAKE in the interest of moving this forward, and having follow up PRs to improve things. m

library/seq.lisp
macrologist: We'd also be able to define `Monad` for `Vector` and consequently `Seq`.

macrologist:
I Tried writing an optimized macro - but it still uses `conc` to make it somewhat sane. However, the `runtimerepr` constraint make sit not typecheck on high-arity invocations of `make`.  Still working on it.

macrologist:
@stylewarning Ok I added an experimental extra commit that changes `seq:make` to build sequences at "compile time". 

The reason it creates such a massive quantity of `coalton:let` bindings is because I was not sure that I could build vectors with coalton expressions inside of them and have it do the right thing. 

That is, I want to be able to write `(seq:make (+ 1 1) (+ 2 2))`.

If there's a better way than the approach I took, I'm all ears. The final case in the macro is the "just freaking concatenate" case.

If nobody objects to the horror of this macro, then I'll squash the commits that add it into the branch to mrege, otherwise I'll abandon them.

PR #972, opened 2023-07-14
Fix codegen of nullary classes with superclasses

### LINKED ISSUE #968: Runtime error caused by defining a typeclass that has no methods
Define `make-bar` function that creates the values of an instance of the Bar typeclass by using the method of the Foo typeclass as follows:

```
(in-package :coalton-user)

(coalton-toplevel
  (define-class (Foo :a)
    (make-foo (String -> :a)))

  (define-class (Foo :a => Bar :a))
    
  (declare make-bar (Bar :a => String -> :a))
  (define (make-bar x)
    (make-foo x))

  (define-type Baz Baz)
  (define-instance (Foo Baz)
    (define (make-foo _) Baz))
  (define-instance (Bar Baz)))

(coalton-toplevel
  (define (main)
    (the Baz (make-bar "Test"))))
```

In REPL:
```
CL-USER> (in-package :coalton-user)
#<COMMON-LISP:PACKAGE "COALTON-USER">
COALTON-USER> (coalton (main))
```

Below is the occurring error:

```
The value
  #<COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY :ARITY 1>
is not of type
  COALTON-USER::CLASS/BAR
when binding #:DICT53
   [Condition of type COMMON-LISP:TYPE-ERROR]
```

In this case, it may be necessary to implement a type error or take measures to prevent runtime errors.

eliaslfox:
Fixes #968

### DIFF
--- a/src/codegen/translate-instance.lisp
+++ b/src/codegen/translate-instance.lisp
@@ -83,7 +83,7 @@
 
          ;; If the instance has methods then apply them
          (app-node
-           (if unqualified-method-definitions 
+           (if (or unqualified-method-definitions superclass-preds)
                (make-node-application
                 :type (pred-type pred env)
                 :rator var-node
--- a/tests/runtime-tests.lisp
+++ b/tests/runtime-tests.lisp
@@ -144,3 +144,21 @@
 (coalton-toplevel
   (define-class (Mult :a :b :c (:a :b -> :c))
     (.* (:a -> :b -> :c))))
+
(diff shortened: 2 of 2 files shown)


stylewarning (approved):

stylewarning:
(a test would be nice)

colescott (approved):

PR #979, opened 2023-08-04
Added seq get-unsafe and put-unsafe, corresponding tests

Izaakwltn:
Added unsafe options for seq:get and seq:put, with verbose errors.

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -18,7 +18,9 @@
    #:pop
    #:size
    #:get
+   #:get-unsafe
    #:put
+   #:put-unsafe
    #:empty?
    #:conc
    #:make))
@@ -60,7 +62,7 @@
      UFix                               ; height
      UFix                               ; cached full subtree size
      (vector:Vector UFix)               ; cumulative size table
-     (vector:Vector (Seq :a)))         ; subtrees
+     (vector:Vector (Seq :a)))          ; subtrees
     (LeafArray (vector:Vector :a)))
 
   (declare new (types:RuntimeRepr :a => Unit -> Seq :a))
@@ -80,8 +82,7 @@
 
   (declare get (Seq :a -> UFix -> Optional :a))
   (define (get seq idx)
-    "Get the member of `seq` at index `idx`, or `None` if `idx` is larger
-than `(size seq)`"
+    "Get the member of `seq` at index `idx`, or `None` if `idx` >= `(size seq)`"
     (match seq
       ((LeafArray leaves)
        (vector:index idx leaves))
@@ -92,6 +93,15 @@ than `(size seq)`"
         ((Tuple subtree-idx offset) <- (cst-search guess cst idx))
          (subtree <- (vector:index subtree-idx sts))
          (get subtree (- idx offset))))))
+
+  (declare get-unsafe (Seq :a -> UFix -> :a))
+  (define (get-unsafe seq idx)
+    "Gets the member of `seq` at index `idx`, triggers error if `idx` >= `(size seq)`."
+    (match (get seq idx)
+      ((Some x) x)
+      ((None) (error (let ((l (math:1- (size seq))))
+                   (lisp String (l idx)
+                         (cl:format cl:nil "Invalid index ~a for Seq with max index ~a." idx l)))))))
   
   (declare put (Seq :a -> Ufix -> :a -> Optional (Seq :a)))
   (define (put seq idx a)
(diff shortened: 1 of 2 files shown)


library/seq.lisp
```diff
@@ -1,9 +1,9 @@
 (coalton-library/utils:defstdlib-package #:coalton-library/seq
   (:use
-   #:coalton
-   #:coalton-library/builtin
-   #:coalton-library/functions
-   #:coalton-library/classes)
+    #:coalton
```
stylewarning, line 3: nit: not right indentation

library/seq.lisp
```diff
@@ -92,6 +93,15 @@ than `(size seq)`"
         ((Tuple subtree-idx offset) <- (cst-search guess cst idx))
          (subtree <- (vector:index subtree-idx sts))
          (get subtree (- idx offset))))))
+
+  (declare get-unsafe (Seq :a -> UFix -> :a))
+  (define (get-unsafe seq idx)
+    "Gets the member of `seq` at index `idx`, triggers error if `idx` >= `(size seq)`."
+    (match (get seq idx)
+      ((Some x) x)
+      (_None (error (let ((l (math:1- (size seq))))
```
stylewarning, line 102: `((None) ...)`

library/seq.lisp
```diff
@@ -115,6 +125,16 @@ contains `a`."
            (vector:set! subtree-idx new-subtree newsts)
            (pure (RelaxedNode h fss cst newsts)))))))
 
+  (declare put-unsafe ((Seq :a) -> UFix -> :a -> (Seq :a)))
+  (define (put-unsafe seq idx a)
+    "If `idx` is less than `(size seq)`, Return a new `seq` whose `idx` position
+contains `a`, otherwise errors."
+    (match (put seq idx a)
+      ((Some x) x)
+      (_None (error (let ((l (math:1- (size seq))))
```
stylewarning, line 134: `((None) ...)`

stylewarning (changes requested):

library/seq.lisp
```diff
@@ -92,6 +93,15 @@ than `(size seq)`"
...
          (subtree <- (vector:index subtree-idx sts))
          (get subtree (- idx offset))))))
+
+  (declare get-unsafe (Seq :a -> UFix -> :a))
+  (define (get-unsafe seq idx)
+    "Gets the member of `seq` at index `idx`, triggers error if `idx` >= `(size seq)`."
+    (match (get seq idx)
+      ((Some x) x)
+      ((None) (error (let ((l (math:1- (size seq))))
+                   (lisp String (l idx)
```
stylewarning, line 103: style nit: indentation

library/seq.lisp
```diff
@@ -115,6 +125,16 @@ contains `a`."
...
            (pure (RelaxedNode h fss cst newsts)))))))
 
+  (declare put-unsafe ((Seq :a) -> UFix -> :a -> (Seq :a)))
+  (define (put-unsafe seq idx a)
+    "If `idx` is less than `(size seq)`, Return a new `seq` whose `idx` position
+contains `a`, otherwise errors."
+    (match (put seq idx a)
+      ((Some x) x)
+      ((None) (error (let ((l (math:1- (size seq))))
+                   (lisp String (l idx)
```
stylewarning, line 135: style nit: indentation

stylewarning (approved):

Izaakwltn:
@macrologist suggested reworking `get-unsafe` and `put-unsafe` as altered copies of `get` and `put` so that they're hopefully faster. 

@stylewarning Do these look like better than the existing PR?
```
  (define (get-unsafe seq idx)
    "Get the member of `seq` at index `idx`, or `None` if `idx` >= `(size seq)`"
    (match seq
      ((LeafArray leaves)
       (vector:index-unsafe idx leaves))
      
      ((RelaxedNode _ fss cst sts)
       (unwrap
         (do
          (let guess = (math:div idx fss))
          ((Tuple subtree-idx offset) <- (cst-search guess cst idx))
           (subtree <- (vector:index subtree-idx sts))
           (get subtree (- idx offset)))))))
```
```
  (define (put-unsafe seq idx a)
    "If `idx` is less than `(size seq)`, Return a new `seq` whose `idx` position
contains `a`."
    (match seq
      ((LeafArray leaves)
       (if (<= (vector:length leaves) idx)
           (error (let ((l (math:1- (size seq))))
                    (lisp String (l idx)
                      (cl:format cl:nil "Invalid index ~a for Seq with max index ~a." idx l))))
           (let ((newleaves (vector:copy leaves)))
             (vector:set! idx a newleaves)
             (LeafArray newleaves))))

      ((RelaxedNode h fss cst sts)
       (unwrap
        (do
         (let guess = (math:div idx fss))
         ((Tuple subtree-idx offset) <- (cst-search guess cst idx))
          (subtree <- (vector:index subtree-idx sts))
          (new-subtree <- (put subtree (- idx offset) a))
          (let ((newsts (vector:copy sts)))
            (vector:set! subtree-idx new-subtree newsts)
            (pure (RelaxedNode h fss cst newsts))))))))
```

macrologist:
These are still just calling `get` and `put` respectively. In your definition for the unsafe get, for example, it's just calling the safe `get` at every step except in the special case that you start with a leaf node. So rather than wrapping the second match case in an `unwrap` - instead get in there and eliminate all the dependence on the `Maybe` monad, using `vector:index-unsafe` whenever you can.

stylewarning:
Closing; can be opened as a new PR with a more sharp-edge approach.

Issue #963, opened 2023-07-07
Fixnums should signal on underflow and overflow


eliaslfox:

macrologist:
Wanted to note that the standard library now depends on underflow [here](https://github.com/coalton-lang/coalton/blob/58856fc9ba7929444f8a1b048a21dd9042ddb4ee/library/seq.lisp#L279) 

Obviously not hard to fix - just wanted to raise awareness.

PR #981, opened 2023-08-13
check unsigned values for underflow.

Jobhdez:
Hello you all,

This fixes this issue https://github.com/coalton-lang/coalton/issues/921.

I ran the code and it indeed signals when unsigned integers underflow:

```
		(declare add-8bit (U8 -> U8 -> U8))
		(define (add-8bit x y)
		  (+ x y)))
; No value
COALTON-USER> (coalton (add-8bit 2 3))
5
COALTON-USER> (coalton (add-8bit 2 -3))
; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {1002A81003}>.
		(declare sub-8bit (U8 -> U8 -> U8))
		(define (sub-8bit x y)
		  (- x y)))
; No value
COALTON-USER> (coalton (sub-8bit 2 1))
1
COALTON-USER> (coalton (sub-8bit 2 3))
; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {1003D96173}>.

		(declare mul2-8bit (U8 -> U8 -> U8))
		(define (mul2-8bit x y)
		  (* x y)))
			 
; No value
COALTON-USER> (coalton (mul2-8bit 2 3))
6
COALTON-USER> (coalton (mul2-8bit 2 -3))
; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {1007ECE083}>.

		(declare from-8bit (Integer -> U8))
		(define (from-8bit n)
		  (fromInt n)))
; No value
COALTON-USER> (coalton (from-8bit 3))
3
COALTON-USER> (coalton (from-8bit -4))

-- Job

### DIFF
--- a/library/math/num.lisp
+++ b/library/math/num.lisp
@@ -157,10 +157,43 @@
 (%define-overflow-handler %handle-64bit-overflow 64)
 (%define-overflow-handler %handle-fixnum-overflow #.+fixnum-bits+)
 
-;;;
-;;; Num instances for integers
+;;; 
+;;; Underflow checks for unsigned values
 ;;;
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:defmacro define-unsigned-num-underflow (type bits)
+    "Define a `Num' instance for Type which signals an error on underflow."
+    `(define-instance (Num ,type)
+       (define (+ a b)
+         (lisp ,type (a b)
+	   (cl:if (cl:and (cl:< b 0) (cl:< a (cl:- 0 b)))
+		  (cl:cerror "Unsigned value underflowed." ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  (cl:+ a b))))
+       
+       (define (- a b)
+         (lisp ,type (a b)
+	   (cl:if (cl:and (cl:>= b 0) (cl:< a (cl:+ 0 b)))
+		  (cl:cerror "Unsigned value underflowed." ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  (cl:- a b))))
+       
+       (define (* a b)
+	 (lisp ,type (a b)
+	   (cl:if (cl:or (cl:and (cl:and (cl:> b 0) (cl:< a 0)) (cl:< a (cl:/ 0 b)))
+			 (cl:and (cl:and (cl:< b 0) (cl:> a 0)) (cl:> a (cl:/ 0 b))))
+		  (cl:cerror "Unsigned value undeflowed." ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  (cl:* a b))))
+
+       (define (fromInt x)
+	 (lisp ,type (x)
+	   (cl:if (cl:< x 0)
+		  (cl:cerror "Unsigned value underflowed." ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  x))))))
+
+
+;;;
+;;; Num instances for integers
+;;;	   
 (cl:eval-when (:compile-toplevel :load-toplevel)
   (cl:defmacro define-num-checked (type overflow-handler)
     "Define a `Num' instance for TYPE which signals on overflow."
@@ -214,8 +247,12 @@
   (define-num-wrapping U16 16)
   (define-num-wrapping U32 32)
   (define-num-wrapping U64 64)
-  (define-num-wrapping UFix #.+unsigned-fixnum-bits+))
+  (define-num-wrapping UFix #.+unsigned-fixnum-bits+)
 
+  (define-unsigned-num-underflow U8 8)
+  (define-unsigned-num-underflow U16 16)
+  (define-unsigned-num-underflow U32 32)
+  (define-unsigned-num-underflow U64 64))
 ;;;
 ;;; Num instances for floats
 ;;;

### CODE REVIEW

File: library/math/num.lisp [hunk 1]
```diff
@@ -157,29 +157,47 @@
 (%define-overflow-handler %handle-64bit-overflow 64)
 (%define-overflow-handler %handle-fixnum-overflow #.+fixnum-bits+)
 
+;;;
```
Izaakwltn, line 160: Shouldn't there also be overflow handlers for unsigned values?

Jobhdez: Unsigned values don’t overflow

Izaakwltn: If the result is supposed to be an unsigned 8-bit Integer, and it evaluates to > 127, wouldn't that be an overflow? 

Even if it wraps around and modulos it or somewhat, that still seems like potentially problematic behavior.

Jobhdez: A quote from the blog post I shared above;

“ An unsigned 8-bit integer can store any value between 0 and 255…”

Jobhdez: Unsigned Ints wrap around but don’t overflow. https://stackoverflow.com/questions/9193880/overflowing-of-unsigned-int

Izaakwltn: Whoops I got turned around, I was thinking in signed for a second, but still... 

The blog post you shared talks about the different ways to handle overflow and underflow, and talks about how wrapping around is less expensive but can cause problems, same with saturation. 

For example, a U8 derived from `(+ 255 1)` should trigger an overflow error, since it can't return 256 without changing to a larger Integer type.
The other two options return questionable results:
With wrap around: `(+ 255 1)` => `1`.
With saturation `(+ 255 1)` => `255`.

The point of the handler is to return an error so you don't get the **wrong** answer somewhere in your code.

Jobhdez: I see what you’re saying.

Here’s the  original issue and it says to signal an error if unsigned numbers underflow. So I a was assuming I didn’t need to check whether they overflowed. https://github.com/coalton-lang/coalton/issues/921

Izaakwltn: That makes sense, I was looking at this issue: #963 which is about signed numbers anyways.

I guess the jury's still out.

Izaakwltn: You are right about wrap around in Coalton-

In SBCL, this errors because 256 is too big for unsigned 8-bit:
```
(declaim (ftype (function ((unsigned-byte 8) (unsigned-byte 8)) (unsigned-byte 8)) add-8bit))

(defun add-8bit (x y)
  (+ x y))

(add-8bit 255 1)
```
But in Coalton, this evaluates to 0:
```
(coalton-toplevel

  (declare add-8bit (U8 -> U8 -> U8))
  (define (add-8bit x y)
    (+ x y)))

(coalton (add-8bit 255 1))
```

Jobhdez: So is my code correct?

Izaakwltn: So for the four cases, unsigned and signed, overflow and underflow, this is what I'm currently getting with your code, let me know if it's consistent with your results:

For signed 8-bit, it's throwing an overflow error for both overflow and underflow. For unsigned, it's wrapping around in both directions, which seems to be what you were trying to change. 

I did check, and these happen to be the same results for the main branch as well.

```
(coalton-toplevel
  (declare add-8bit (U8 -> U8 -> U8))
  (define (add-8bit x y)
    (+ x y))

  (declare sub-8bit (U8 -> U8 -> U8))
  (define (sub-8bit x y)
    (- x y))

  (declare add-I8 (I8 -> I8 -> I8))
  (define (add-I8 x y)
    (+ x y))
  
  (declare sub-I8 (I8 -> I8 -> I8))
  (define (sub-I8 x y)
    (- x y)))
```
```
COALTON-USER> (sub-I8 -50 100)

 Signed value overflowed 8 bits.
```
```
COALTON-USER> (add-I8 50 100)

 Signed value overflowed 8 bits.
```
```
COALTON-USER> (add-8bit 255 1)
0
```
```
COALTON-USER> (sub-8bit 1 10)
247
```

Jobhdez: Thanks - I’ll look into it

Izaakwltn:
It's definitely throwing an error when it needs to now, though I'm not sure the error messages are quite clear for potential users:

Using the example functions from above:
```
COALTON-USER> (sub-8bit 1 255)
; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {10091946B3}>.
```
This probably should say "Unsigned value underflowed 8 bits".

```
COALTON-USER> (add-8bit 255 1)
; Evaluation aborted on #<TYPE-ERROR expected-type: (UNSIGNED-BYTE 8) datum: 256>.
```
This one is pretty clear, but it might be better to have something from Coalton instead of depending on the Lisp implementation's error messages, likewise "Unsigned value overflowed 8 bits."

Jobhdez:
> It's definitely throwing an error when it needs to now, though I'm not sure the error messages are quite clear for potential users:
> 
> Using the example functions from above:
> 
> ```
> COALTON-USER> (sub-8bit 1 255)
> ; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {10091946B3}>.
> ```
> 
> This probably should say "Unsigned value underflowed 8 bits".
> 
> ```
> COALTON-USER> (add-8bit 255 1)
> ; Evaluation aborted on #<TYPE-ERROR expected-type: (UNSIGNED-BYTE 8) datum: 256>.
> ```
> 
> This one is pretty clear, but it might be better to have something from Coalton instead of depending on the Lisp implementation's error messages, likewise "Unsigned value overflowed 8 bits."

I used `cl-cerror` because the surrounding code uses this. I tried to follow the patterns in the surrounding code.
What do you think?

Jobhdez:
> It's definitely throwing an error when it needs to now, though I'm not sure the error messages are quite clear for potential users:
> 
> Using the example functions from above:
> 
> ```
> COALTON-USER> (sub-8bit 1 255)
> ; Evaluation aborted on #<SB-INT:SIMPLE-PROGRAM-ERROR "invalid number of arguments: ~S" {10091946B3}>.
> ```
> 
> This probably should say "Unsigned value underflowed 8 bits".
> 
> ```
> COALTON-USER> (add-8bit 255 1)
> ; Evaluation aborted on #<TYPE-ERROR expected-type: (UNSIGNED-BYTE 8) datum: 256>.
> ```
> 
> This one is pretty clear, but it might be better to have something from Coalton instead of depending on the Lisp implementation's error messages, likewise "Unsigned value overflowed 8 bits."

Those are actually not the errors. If you try out if it underflows it signals a proper error with a similar error message as you suggested.

Izaakwltn:
They're the errors it's showing me 🤷 .

I think the problem is that `cerror` requires at least two arguments: a continue-format-control, and a datum.

For example, earlier in the file:
```
(cl:defmacro %define-overflow-handler (name bits)
  `(cl:progn
     (cl:declaim (cl:inline ,name))
     (cl:defun ,name (value)
       (cl:typecase value
         ((cl:signed-byte ,bits) value)
         (cl:otherwise
          (cl:cerror "Continue, wrapping around."
                     ,(cl:format cl:nil "Signed value overflowed ~D bits." bits))
          (%unsigned->signed ,bits (cl:mod value ,(cl:expt 2 bits))))))))
```
`cl:cerror` takes "Continue, wrapping around." and the error message. 

You may also want to add that `~D bits` portion just for verbosity.

Jobhdez:
> They're the errors it's showing me 🤷 .
> 
> I think the problem is that `cerror` requires at least two arguments: a continue-format-control, and a datum.
> 
> For example, earlier in the file:
> 
> ```
> (cl:defmacro %define-overflow-handler (name bits)
>   `(cl:progn
>      (cl:declaim (cl:inline ,name))
>      (cl:defun ,name (value)
>        (cl:typecase value
>          ((cl:signed-byte ,bits) value)
>          (cl:otherwise
>           (cl:cerror "Continue, wrapping around."
>                      ,(cl:format cl:nil "Signed value overflowed ~D bits." bits))
>           (%unsigned->signed ,bits (cl:mod value ,(cl:expt 2 bits))))))))
> ```
> 
> `cl:cerror` takes "Continue, wrapping around." and the error message.
> 
> You may also want to add that `~D bits` portion just for verbosity.
Ok sounds good. I’ll do that. Thanks

Jobhdez:
> They're the errors it's showing me 🤷 .
> 
> I think the problem is that `cerror` requires at least two arguments: a continue-format-control, and a datum.
> 
> For example, earlier in the file:
> 
> ```
> (cl:defmacro %define-overflow-handler (name bits)
>   `(cl:progn
>      (cl:declaim (cl:inline ,name))
>      (cl:defun ,name (value)
>        (cl:typecase value
>          ((cl:signed-byte ,bits) value)
>          (cl:otherwise
>           (cl:cerror "Continue, wrapping around."
>                      ,(cl:format cl:nil "Signed value overflowed ~D bits." bits))
>           (%unsigned->signed ,bits (cl:mod value ,(cl:expt 2 bits))))))))
> ```
> 
> `cl:cerror` takes "Continue, wrapping around." and the error message.
> 
> You may also want to add that `~D bits` portion just for verbosity.

Hey @Izaakwltn,

I fixed the error messaging; now the error appears as:

``` 
Unsigned value underflowed 8 bits.
   [Condition of type COMMON-LISP:SIMPLE-ERROR]

Restarts:
 0: [CONTINUE] Unsigned value underflowed.
 1: [RETRY] Retry SLIME REPL evaluation request.
 2: [*ABORT] Return to SLIME's top level.
 3: [ABORT] abort thread (#<THREAD tid=4205 "new-repl-thread" RUNNING {10019A0003}>)
```

Izaakwltn:
The underflow error is correct now! 

I still wonder whether there should be a corresponding overflow error for unsigned types (which I know wasn't specifically mentioned in the issue you're responding to (and is up for debate)).

You'll need to squash your commits so there's only one going in, but otherwise I'll approve this.

@eliaslfox maybe you could take a look at this.

PR #986, opened 2023-09-12
loop sampler

macrologist:
This PR introduces looping constructs into Coalton. 

Some algorithms are more naturally expressed with imperative and go-to style iteration than they are with recursion. This is especially the case with complex search algorithms and scheduling algorithms that utilize some form of backtracking.  It is not that these algorithms *cannot* be expressed in a recursive style, only that sometimes it is awkward to do so. 

This PR takes a crack at addressing this need. It enlarges the toolbox of Coalton programmers without sacrificing the things that make Coalton great: excellent error messages, and statically typed programs.

Here's what this PR lets you do:

``` lisp

;; loop while a predicate is true
(while <predicate> 
   <action1> 
   <action2>
   ... )

;; loop while a pattern matches
(while-let  pattern = expression 
   <action1> 
   <action2>
   ...)

;; loop over values in an intoiterator instance
(for <pattern> in <intoiterator-value>
   <action1> 
   <action2>
   ...)

;; loop forever
(loop [optional-label] 
    <action1>
    <action2> 
    ...)


```

Moreover, `(continue)` and `(break)` are available in all of the above loops. If in a `loop`, both `continue` and `break` accept an optional `label` form.  A parse error will signal when the provided label does not match the label of a lexically enclosing loop.  (i.e. you cannot escape from a loop by calling `(break :moo)` in some function that a loop happens to call).

- [x] `while`
- [x] `while-let` 
- [x] `for` (iterators)
- [x] `loop`
- [x] a `break` and `continue` for use in `while` and `loop`
- [x] `break` and `continue` in `while`-style loops
- [x] labelled `loop` with labelled `break` and `continue`

Any and all of these can be scrapped.

### DIFF
--- a/src/codegen/codegen-expression.lisp
+++ b/src/codegen/codegen-expression.lisp
@@ -13,12 +13,21 @@
    (#:settings #:coalton-impl/settings)
    (#:util #:coalton-impl/util)
    (#:rt #:coalton-impl/runtime)
-   (#:tc #:coalton-impl/typechecker))
+   (#:tc #:coalton-impl/typechecker)
+   (#:const #:coalton-impl/constants))
   (:export
    #:codegen-expression                 ; FUNCTION
    ))
 
 (in-package #:coalton-impl/codegen/codegen-expression)
+
+(defun continue-label (lаbеl)
+  (declare (type symbol lаbеl))
+  (alexandria:format-symbol :keyword "~a-CONTINUE" lаbеl))
+
+(defun break-label (lаbеl)
+  (declare (type symbol lаbеl))
+  (alexandria:format-symbol :keyword "~a-BREAK" lаbеl))
 
 (defgeneric codegen-expression (node current-function env)
   (:method ((node node-literal) current-function env)
@@ -107,10 +116,65 @@
                 ,inner)
           inner)))
 
-  (:method ((expr node-match) current-function env)
+  (:method ((expr node-while) current-function env)
     (declare (type tc:environment env)
              (type (or null symbol) current-function))
 
+    (let ((pred-expr (codegen-expression (node-while-expr expr) current-function env))
+          (body-expr (codegen-expression (node-while-body expr) current-function env))
+          (label (node-while-label expr)))
+      `(loop
+         :named ,(break-label label)
+         :while ,pred-expr
+         :do
+            (block ,(continue-label label) ,body-expr))))
+
+  (:method ((expr node-while-let) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
+
+    (let ((match-expr (codegen-expression (node-while-let-expr expr) current-function env))
+          (body-expr (codegen-expression (node-while-let-body expr) current-function env))
+          (label (node-while-let-label expr))
+          (match-var (gensym "MATCH")))
+
+      (multiple-value-bind (pred bindings)
+          (codegen-pattern (node-while-let-pattern expr) match-var env)
+        `(loop
+           :named ,(break-label label)
+           :for ,match-var
+             := ,(if settings:*emit-type-annotations*
+                     `(the ,(tc:lisp-type (node-type (node-while-let-expr expr)) env) ,match-expr)
+                     match-expr)
+           :while ,pred
+           :do (block ,(continue-label label)
+                 ,(cond ((null bindings) body-expr)
+                        (t `(let ,bindings
+                              (declare (ignorable ,@(mapcar #'car bindings)))
+                              ,body-expr))))))))
+
+  (:method ((expr node-loop) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
+    (let ((body-expr (codegen-expression (node-loop-body expr) current-function env))
+          (label (node-loop-label expr)))
+      `(loop :named  ,(break-label label)
+             :do (block ,(continue-label label)
+                   ,body-expr))))
+
+  (:method ((expr node-break) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
+    `(return-from ,(break-label (node-break-label expr))))
+
+  (:method ((expr node-continue) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
+    `(return-from ,(continue-label (node-continue-label expr))))
+
+  (:method ((expr node-match) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
     ;; If possible codegen a cl:if instead of a trivia:match
     (when (and (equalp (node-type (node-match-expr expr)) tc:*boolean-type*)
                (= 2 (length (node-match-branches expr)))
--- a/src/codegen/pattern.lisp
+++ b/src/codegen/pattern.lisp
@@ -6,7 +6,7 @@
    (#:tc #:coalton-impl/typechecker))
   (:export
    #:pattern                            ; STRUCT
-   #:pattern-source                     ; ACCESSOR
+   #:pattern-type                       ; ACCESSOR
    #:pattern-list                       ; TYPE
    #:pattern-var                        ; STRUCT
    #:make-pattern-var                   ; ACCESSOR
--- a/src/codegen/translate-expression.lisp
+++ b/src/codegen/translate-expression.lisp
@@ -141,7 +141,8 @@ Returns a `node'.")
 
              (num-class (util:find-symbol "NUM" classes-package))
 
-             (num-pred (tc:make-ty-predicate :class num-class :types (list (tc:qualified-ty-type qual-ty))))
+             (num-pred (tc:make-ty-predicate :class num-class
+                                             :types (list (tc:qualified-ty-type qual-ty))))
 
              (from-int-method (util:find-symbol "FROMINT" classes-package)))
         
@@ -549,6 +550,159 @@ Returns a `node'.")
                    :body (make-node-variable
                           :type tc:*unit-type*
                           :value unit-value))))))
+
+  (:method ((expr tc:node-while) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    
+    (make-node-while 
+     :type tc:*unit-type*
+     :label (tc:node-while-label expr)
+     :expr (translate-expression (tc:node-while-expr expr) ctx env)
+     :body (translate-expression (tc:node-while-body expr) ctx env)))
+
+  (:method ((expr tc:node-while-let) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    (make-node-while-let
+     :type tc:*unit-type*
+     :pattern (translate-pattern (tc:node-while-let-pattern expr))
+     :label (tc:node-while-let-label expr)
+     :expr (translate-expression (tc:node-while-let-expr expr) ctx env)
+     :body (translate-expression (tc:node-while-let-body expr) ctx env)))
+
+  (:method ((expr tc:node-for) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    (let* ((pat-arg
+             (translate-pattern (tc:node-for-pattern expr)))
+
+           (pat-arg-ty
+             (pattern-type pat-arg))
+
+           (classes-package
+             (find-package "COALTON-LIBRARY/CLASSES"))
+
+           (some
+             (util:find-symbol "SOME" classes-package))
+
+           (optional
+             (util:find-symbol "OPTIONAL" classes-package))
+
+           (optional-pat-arg-ty
+             (tc:apply-type-argument
+              (tc:make-tycon :name optional
+                             :kind (tc:make-kfun :from tc:+kstar+ :to tc:+kstar+))
+              pat-arg-ty))
+
+           (some-pattern
+             (make-pattern-constructor
+              :type optional-pat-arg-ty
+              :name some
+              :patterns (list pat-arg)))
+
+           (into-iter-arg
+             (translate-expression (tc:node-for-expr expr) ctx env))
+
+           (into-iter-arg-ty
+             (node-type into-iter-arg))
+
+           (iterator-package
+             (find-package "COALTON-LIBRARY/ITERATOR"))
+
+           (into-iter-method
+             (util:find-symbol "INTO-ITER" iterator-package))
+
+           (intoiter-class
+             (util:find-symbol "INTOITERATOR" iterator-package))
+
+           (iterator
+             (util:find-symbol "ITERATOR" iterator-package))
+
+           (iter-ty
+             (tc:apply-type-argument
+              (tc:make-tycon :name iterator
+                             :kind (tc:make-kfun :from tc:+kstar+ :to tc:+kstar+))
+              into-iter-arg-ty))
+
+           (intoiterator-pred
+             (tc:make-ty-predicate
+              :class intoiter-class
+              :types (list into-iter-arg-ty pat-arg-ty)
+              :source (tc:node-source expr)))
+
+           (into-iter-node
+             (make-node-application
+              :type optional-pat-arg-ty
+              :rator (make-node-variable
+                      :type (tc:make-function-type*
+                             (list (pred-type intoiterator-pred env)
+                                   (node-type into-iter-arg))
+                             iter-ty)
+                      :value into-iter-method)
+              :rands (list
+                      (resolve-dict intoiterator-pred ctx env)
+                      into-iter-arg)))
+
+           (next-method
+             (util:find-symbol "NEXT!" iterator-package))
+
+           (into-iter-binding-var-name
+             (gensym "ITER-"))
+
+           (iter-next-node
+             (make-node-application
+              :type optional-pat-arg-ty
+              :rator (make-node-variable
+                      :type (tc:make-function-type*
+                             (list iter-ty)
+                             optional-pat-arg-ty)
+                      :value next-method)
+              :rands (list (make-node-variable
+                            :type iter-ty
+                            :value  into-iter-binding-var-name))))
+
+           (while-let-node
+             (make-node-while-let
+              :type tc:*unit-type*
+              :label (tc:node-for-label expr)
+              :pattern some-pattern
+              :expr iter-next-node
+              :body (translate-expression (tc:node-for-body expr) ctx env))))
+
+      (make-node-bind
+       :type tc:*unit-type* 
+       :name into-iter-binding-var-name
+       :expr into-iter-node
+       :body while-let-node)))
+
+  (:method ((expr tc:node-loop) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    (make-node-loop
+     :type tc:*unit-type*
+     :label (tc:node-loop-label expr)
+     :body (translate-expression (tc:node-loop-body expr) ctx env)))
+
+  (:method ((expr tc:node-break) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    (make-node-break
+     :type tc:*unit-type*
+     :label (tc:node-break-label expr)))
+
+  (:method ((expr tc:node-continue) ctx env)
+    (declare (type pred-context ctx)
+             (type tc:environment env)
+             (values node))
+    (make-node-continue
+     :type tc:*unit-type*
+     :label (tc:node-continue-label expr)))
 
   (:method ((expr tc:node-cond) ctx env)
     (declare (type pred-context ctx)
--- a/src/constants.lisp
+++ b/src/constants.lisp
@@ -1,0 +1,15 @@
+;;;; constants.lisp
+;;;;
+;;;; This file contains constant values used throughout compilation.
+
+(defpackage #:coalton-impl/constants
+  (:use #:cl)
+  (:export
+   #:+default-loop-label+               ; VARIABLE
+   ))
+
+(in-package #:coalton-impl/constants)
+
+(defconstant +default-loop-label+ :coalton-loop
+  "Suррlіеԁ аs а lоор lаbеl оr wһіlе, whіlе-lеt, fоr, lоор, brеаk, аnԁ
+соntіnuе whеn а lаbеl іs nоt suррlіеԁ bу thе usеr.")
--- a/src/parser/expression.lisp
+++ b/src/parser/expression.lisp
@@ -11,7 +11,8 @@
    #:parse-error)
   (:local-nicknames
    (#:cst #:concrete-syntax-tree)
-   (#:util #:coalton-impl/util))
+   (#:util #:coalton-impl/util)
+   (#:const #:coalton-impl/constants))
   (:export
    #:node                               ; STRUCT
    #:node-source                        ; ACCESSOR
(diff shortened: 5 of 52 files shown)


macrologist:
Here's an awkward use of iterators

```

(coalton 
               (let ((itr (iter:range-increasing 10 0 100))
                     (item (cell:new (iter:next! itr)))
                     (sum (cell:new 0)))
                 (while (optional:some? (cell:read item)) 
                   (cell:update! (+ (defaulting-unwrap (cell:read item))) sum) 
                   (cell:swap! item (iter:next! itr))
                   Unit)
                 (cell:read sum)))
450

```

eliaslfox:
The code looks really good!

macrologist:
Added `while-let`

```

(coalton
               (let ((xs (cell:new (make-list 1 2 3 4))))
                 (while-let (Some x) = (cell:pop! xs)
                            (traceobject "x = " x))))

x = : 1
x = : 2
x = : 3
x = : 4

```

macrologist:
`loop` with `break` and `continue`


``` lisp

COALTON-USER> (coalton 
               (let ((rand (fn () (lisp Integer () (cl:random 10)))))
                 (loop 
                   (let n = (rand))
                   (cond ((== n 5) 
                          (trace "got a five, moving on") 
                          (continue)) 
                         ((== n 0) 
                          (trace "got a zero, we're done")
                          (break))
                         (True 
                          Unit))
                   (traceobject "n" n))))
n: 1
n: 4
got a five, moving on
n: 9
n: 4
n: 4
got a zero, we're done
COMMON-LISP:NIL
COALTON-USER> 

```

macrologist:
for loops 

```

COALTON-USER> (coalton 
               (let ((iter (iter:up-to 10)))
                 (for x in iter
                      (when (== x 8) 
                        (trace "BREAK!")
                        (break))
                      (when (== 0 (mod x 2)) 
                        (continue))
                      (traceobject "x" x))))
x: 1
x: 3
x: 5
x: 7
BREAK!
```

``` lisp

COALTON-USER> (coalton 
               (let ((iter "cool"))
                 (for x in iter
                      (traceobject "x" x))))
x: c
x: o
x: o
x: l
```

macrologist:
labelled loops 

``` lisp

(coalton 
               (let ((sum (cell:new 0)))
                 (named-loop 
                  :outer 
                  (let random = (fn (n) (lisp Integer (n) (cl:random n))))
                  (named-loop 
                   :inner 
                   (let n = (random 100)) 
                   (when (== 0 n) 
                     (trace "Rolled a zero. Abort!")
                     (break-from :outer))
                   (when (< 1000 (cell:update! (+ n) sum)) 
                     (trace "We're bigger than 1000. Start summing again")
                     (break-from :inner)))
                   (cell:swap! sum 0) Unit) 
                 (traceobject "final sum was" (cell:read sum))))
We're bigger than 1000. Start summing again
We're bigger than 1000. Start summing again
We're bigger than 1000. Start summing again
We're bigger than 1000. Start summing again
Rolled a zero. Abort!
final sum was: 463

```

eliaslfox:
This is looking really good. I have a few minor thoughts about the syntax.

* I don't think `named-loops` should be a special form, but instead every loop should be label-able. 
* `break` and `break-from` should be unified to a single form with an optional argument.
* `continue` should also take a label as an optional argument.

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+             (and (cst:consp (cst:rest form))
+                  (if (and (cst:atom (cst:second form))
+                           (keywordp (cst:raw (cst:second form)))) 
+                      (cst:raw (cst:second form))
+                      ;; (break not-a-keyword)
+                      (error 'parse-error
+                             :err (coalton-error
+                                   :span (cst:source form)
+                                   :file file
+                                   :highlight :end
```
eliaslfox, line 1070: This shouldn't `:highlight :end`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :message "Invalid break expression"
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
+                        :file file
+                        :highlight :end
```
eliaslfox, line 1079: This shouldn't `:highlight :end`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                        :message "Invalid break expression"
+                        :primary-note (format nil "Label ~s not found in any enclosing loop" label)))))
+
+       (unless *escapable-loop-context*
+         (error 'parse-error
+                :err (coalton-error
+                      :span (cst:source form)
+                      :file file
+                      :highlight :end
```
eliaslfox, line 1088: same here

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+             (and (cst:consp (cst:rest form))
+                  (if (and (cst:atom (cst:second form))
+                           (keywordp (cst:raw (cst:second form)))) 
+                      (cst:raw (cst:second form))
+                      ;; (continue not-a-keyword)
+                      (error 'parse-error
+                             :err (coalton-error
+                                   :span (cst:source form)
+                                   :file file
+                                   :highlight :end
```
eliaslfox, line 1108: and here

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :message "Invalid continue expression"
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
+                        :file file
+                        :highlight :end
```
eliaslfox, line 1117: and here

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                        :message "Invalid continue expression"
+                        :primary-note (format nil "Label ~s not found in any enclosing loop" label)))))
+
+       (unless *escapable-loop-context*
+         (error 'parse-error
+                :err (coalton-error
+                      :span (cst:source form)
+                      :file file
+                      :highlight :end
```
eliaslfox, line 1126: and here

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                      :message "Invalid continue expression"
+                      :primary-note "no enclosing loop")))
+
+       (make-node-continue :source (cst:source form) :label label)))
+    
+
+    ((and (cst:atom (cst:first form))
+          (eq 'coalton:for (cst:raw (cst:first form))))
+     (macrolet ((for-error (note)
```
eliaslfox, line 1135: Don't macrolet here

tests/parser/parse-break.bad.coalton
```diff
@@ -0,0 +1,4 @@
+;; BAD: Parse-Break
+(package test-parser)
+
+(define f (break))
```
eliaslfox, line 4: All parser test files should have a trailing newline

src/codegen/codegen-expression.lisp
```diff
@@ -107,6 +110,92 @@
...
+  
+  (:method ((expr node-loop) current-function env)
+    (declare (type tc:environment env)
+             (type (or null symbol) current-function))
+    (let ((body-expr (codegen-expression (node-loop-body expr) current-function env))
+          (label (node-loop-label expr)))
+      `(block ,+break-label+
+         (loop ,@(when label (list :named label))
+               :do ,(if label
+                        `(tagbody ,label
```
eliaslfox, line 179: Is there a reason this can't be
```
`(block ,label
    (block ,+continue-label
      ,body-expr))
```

src/codegen/pattern.lisp
```diff
@@ -6,7 +6,8 @@
    (#:tc #:coalton-impl/typechecker))
   (:export
    #:pattern                            ; STRUCT
-   #:pattern-source                     ; ACCESSOR
+   #:pattern-source                     ; ACCESSOR THIS DOESN'T SEEM TO EXIST
```
eliaslfox, line 9: Delete it. It was probably a copy-paste mistake.

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+        :pattern (parse-pattern (cst:second form) file) 
+        :expr (parse-expression (cst:fourth form) file)
+        :body (parse-body (cst:nthrest 4 form) form  file))))
+
+
+    ((and (cst:atom (cst:first form))
+          (eq 'coalton:loop (cst:raw (cst:first form))))
+
+     (let* ((label
+              ;; (loop)
```
eliaslfox, line 1016: misplaced comment

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+     (let ((label
+             ;; (break)
+             (and (cst:consp (cst:rest form))
+                  (if (and (cst:atom (cst:second form))
+                           (keywordp (cst:raw (cst:second form)))) 
+                      (cst:raw (cst:second form))
+                      ;; (break not-a-keyword)
+                      (error 'parse-error
+                             :err (coalton-error
+                                   :span (cst:source form)
```
eliaslfox, line 1068: `:span (cst:second form)`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :file file
+                                   :highlight :end
+                                   :message "Invalid break expression"
```
eliaslfox, line 1071: `"Invalid label in break"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :primary-note "expected a keyword label"))))))
```
eliaslfox, line 1072: `"expected a keyword"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :message "Invalid break expression"
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
+                        :file file
+                        :highlight :end
+                        :message "Invalid break expression"
```
eliaslfox, line 1080: `"Invalid label in break"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :span (cst:source form)
+                                   :file file
+                                   :highlight :end
+                                   :message "Invalid break expression"
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
```
eliaslfox, line 1077: `:span (cst:second form)`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                  (if (and (cst:atom (cst:second form))
+                           (keywordp (cst:raw (cst:second form)))) 
+                      (cst:raw (cst:second form))
+                      ;; (continue not-a-keyword)
+                      (error 'parse-error
+                             :err (coalton-error
+                                   :span (cst:source form)
+                                   :file file
+                                   :highlight :end
+                                   :message "Invalid continue expression"
```
eliaslfox, line 1109: `"Invalid label in continue"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :primary-note "expected a keyword label"))))))
```
eliaslfox, line 1110: `"expected a keyword"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+     (let ((label
+             ;; (continue)
+             (and (cst:consp (cst:rest form))
+                  (if (and (cst:atom (cst:second form))
+                           (keywordp (cst:raw (cst:second form)))) 
+                      (cst:raw (cst:second form))
+                      ;; (continue not-a-keyword)
+                      (error 'parse-error
+                             :err (coalton-error
+                                   :span (cst:source form)
```
eliaslfox, line 1106: `:span (cst:second form)`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :file file
+                                   :highlight :end
+                                   :message "Invalid continue expression"
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
```
eliaslfox, line 1115: `:span (cst:second form)`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                        :file file
+                        :highlight :end
+                        :message "Invalid continue expression"
```
eliaslfox, line 1118: `"Invalid label in continue"`

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
+                        :file file
+                        :highlight :end
+                        :message "Invalid break expression"
+                        :primary-note (format nil "Label ~s not found in any enclosing loop" label)))))
```
eliaslfox, line 1081: `"label not found in any enclosing loop"`

no need to pretty print the label name when it can be highlighted

src/parser/expression.lisp
```diff
@@ -875,6 +945,225 @@
...
+                                   :primary-note "expected a keyword label"))))))                 
+       (when label
+         (unless (member label *loop-label-context*)
+           (error 'parse-error
+                  :err (coalton-error
+                        :span (cst:source form)
+                        :file file
+                        :highlight :end
+                        :message "Invalid continue expression"
+                        :primary-note (format nil "Label ~s not found in any enclosing loop" label)))))
```
eliaslfox, line 1119: `"label not found in any enclosing loop"`

src/typechecker/define.lisp
```diff
@@ -1212,14 +1157,217 @@ Returns (VALUES INFERRED-TYPE PREDICATES NODE SUBSTITUTIONS)")
...
+
+        (multiple-value-bind (expr-ty preds accessors expr-node subs)
+            (infer-expression-type (parser:node-for-expr node) (tc:make-variable) subs env file)
+
+          (multiple-value-bind (body-ty preds_ accessors_ body-node subs)
+              (infer-expression-type (parser:node-for-body node) (tc:make-variable) subs env file)
+
+            (declare (ignore body-ty))
+
+            (setf
```
eliaslfox, line 1274: One assignment per `setf`

src/codegen/codegen-expression.lisp
macrologist: This is to let `(continue :my-special-little-loop)` do the right thing. We use `tagbody` so that we can use `go`, otherwise we'd only hav `return-from` at our disposal and `(continue :my-special-little-loop)` would generate the same code as `(break :my-special-little-loop)`. 


Consider:

``` lisp

`(block ,+break-label+
    (loop :named :my-special-little-loop 
         :do (block ,:my-special-little-loop
                  (block ,+continue-label+
                       (return-from :my-special-little-loop) ;; is this a break or a continue ???
                        (print "not gonna happen")))))
```

If that `return-from` were generated from a `continue`, then it behaves as we expect, but if was generated from a `break` then it doesn't. That is `return-from` returns from the closest enclosing block with the name. If we want to `break` from the loop by label, then we can't; we'll always end up breaking from the inner block named `:my-special-little-loop`.

eliaslfox: That makes sense. I wonder if it would be easier to have all loops and all breaks/continues be labelled. I think that would reduce the number of code paths.

macrologist: I sort of compromised. 

If a loop is given a label, then we generate something like the above (but without the use of tagbody and go, instead I'm turning a label into a "continue label" or a "break label" and inserting them properly at codegen).  If a loop isn't labelled, then we generate less wrapped code using default labels. 

e.g. 

```
(loop (break))
```

generates something like 

``` 
(loop :named :coalton-loop-break 
          :do (block :coalton-loop-continue 
                      (return-from :coalton-loop-break)))

```

whereas 

``` 

(loop :x (break))

```

generates something like 

``` 

(block :coalton-loop-break 
   (loop :named :x-break 
             :do (block :coalton-loop-continue 
                        (block :x-continue 
                              (return-from :coalton-loop-break)))))
```

And the reason is I hope clear: we want to support both labelled and unlabelled breaks inside labelled loops. 

My expectation is that for most code that makes use of break and continue, unlabelled variants will be sufficient.

macrologist:
take it or leave it

stylewarning:
> take it or leave it

Is this complete and ready for review?

stylewarning:
@chrsnjk from Twitter suggests creating a fresh lexical binding for iterator variables to avoid the

    (loop :for i :below 10 :collect (lambda () i))

footgun. Go just recently adopted this.

macrologist:
Already there on acount of the way pattern matching works. The `for` and `while-let` loops both recycle pattern-match binding codegen functions, which produces fresh bindings per match. 

```

(coalton 
 (let ((v (vector:new)))
   (for x in "abcd" (vector:push! (fn () x) v))
   (map (fn (f) (f)) v)))

```

returns 

```
#(#\a #\b #\c #\d)
```

as expected.

src/codegen/translate-expression.lisp
```diff
@@ -550,6 +551,139 @@ Returns a `node'.")
...
+             (type tc:environment env)
+             (values node))
+    (make-node-while-let
+     :type tc:*unit-type*
+     :pattern (translate-pattern (tc:node-while-let-pattern expr))
+     :label (tc:node-while-let-label expr)
+     :expr (translate-expression (tc:node-while-let-expr expr) ctx env)
+     :body (translate-expression (tc:node-while-let-body expr) ctx env)))
+
+  (:method ((expr tc:node-for) ctx env)
```
eliaslfox, line 576: Can `tc:node-for` translate into an `ast:node-while-let`?

src/constants.lisp
```diff
@@ -0,0 +1,15 @@
...
+
+(defpackage #:coalton-impl/constants
+  (:use #:cl)
+  (:export
+   #:+default-loop-label+               ; VARIABLE
+   ))
+
+(in-package #:coalton-impl/constants)
+
+(defparameter +default-loop-label+ :coalton-loop
```
eliaslfox, line 13: `defconstant`

src/parser/expression.lisp
```diff
@@ -140,6 +168,12 @@
 
 (defvar *macro-expansion-count* 0)
 
+(defvar *loop-label-context* nil
```
eliaslfox, line 171: `(declaim (type util:symbol-list *loop-label-context*))`

src/typechecker/define.lisp
```diff
@@ -1212,14 +1157,219 @@ Returns (VALUES INFERRED-TYPE PREDICATES NODE SUBSTITUTIONS)")
...
+
+      (multiple-value-bind (body-ty preds_ accessors_ body-node subs)
+          (infer-expression-type (parser:node-while-body node)
+                                 (tc:make-variable)
+                                 subs
+                                 env
+                                 file)
+        (declare (ignore body-ty))
+
+        (setf preds     (append preds     preds_))
```
eliaslfox, line 1185: Don't align these

eliaslfox (commented):
I think the label tracking is still more complex then it needs to be. It should be sufficient to track the stack of non-default labels and the label of the most inner loop for unlabeled breaks. That should eliminate the difference between labelled and unlabelled loops in codegen.

A couple other minor things:
* your tests aren't added to the asd file, and they should be expanded to include labelled for and while loops.
* the intro document needs to a section on loops
* it shouldn't be possible to break or continue from a lambda inside a loop

stylewarning:
> it shouldn't be possible to break or continue from a lambda inside a loop

I think it should be OK, as long as the lambda is called within the dynamic extent of the loop. See [here](http://clhs.lisp.se/Body/03_ae.htm) and [here](http://clhs.lisp.se/Body/03_af.htm) for how CL thinks about it, including a few examples of RETURN-FROM within lambdas.

src/constants.lisp
macrologist: Ah yeah I never use defconstant b/c I basically only hack on things with interactive development in mind and defconstant is too much of a pain.

macrologist:
> > it shouldn't be possible to break or continue from a lambda inside a loop
> 
> I think it should be OK, as long as the lambda is called within the dynamic extent of the loop. See [here](http://clhs.lisp.se/Body/03_ae.htm) and [here](http://clhs.lisp.se/Body/03_af.htm) for how CL thinks about it, including a few examples of RETURN-FROM within lambdas.

As it is, this this does what you want:

```lisp

(coalton
 (for x in "ahoy there"
      (let ((breakproxy (fn (z) (when (== z #\Space) (break)))))
        (breakproxy x)
        (traceobject "x" x))))

```

But the following will compile, but error at runtime.

```lisp

(coalton-toplevel 
  (define (ffff)
    (let ((breaker (cell:new None)))
      (loop :ohno (cell:swap! breaker (Some (fn () (break :ohno)))) (break))
      (for :ohno x in "abcd"
           (match (cell:read breaker)
             ((None) Unit)
             ((Some b) (b)))
           (traceobject "x" x)))))

```

For exactly the reasons discussed in 3.1.6 of the hyperspec linked above. We have to produce the `breaker` function in such a convoluted way b/c it is a parse error to include `break` outside of the lexical span of an encolsing looping form (one of `loop`, `for`, `while`, or `while-let`). So that covers what are likely to be "naturally occurring" abuses of `break` and `continue`.   

With effort I think we could cause an error at codegen during translate-expression and cribbing a version of variable renaming, adapting it to the case of loop labels. I'm still thinking about it this morning, but that is my plan.

src/codegen/translate-expression.lisp
macrologist: this is not very convenient

eliaslfox:
I don't want to support this.

> With effort I think we could cause an error at codegen during translate-expression by tracking label contexts and ensuring that unique block names are generated for each context, throwing an error whenever an out-of-context name appears in a break enclosed by an abstraction.

Translation and codegen should never error. User errors must be caught in or before type checking, when it's still possible to generate good errors.

I don't think it's possible to ensure that breaks from subroutines are always valid. You can always generate a valid lambda that outlives the dynamic extent of its loop:

```
(define (f)
  (loop :label
    (return (lambda () (break)))))
```

eliaslfox:
Treat unlabelled loops as loops whose label is the default label. Then every break/continue can be labelled.

macrologist:
> Treat unlabelled loops as loops whose label is the default label. Then every break/continue can be labelled.

That's already what's happening. Each loop is labelled at parse time. If a loop is not supplied a label by the user, then it is given the default label.

So then I can now see two different possible things that you mean.

1) you would like for bare `(break)` and `(continue)` forms to receive the label of the inner-most loop instead of the default label. 
2) you would like for `(break)` and `(continue)` forms to appear *only* in "unlabelled" loops; I.e. those that were not explicitly labelled by a user.

eliaslfox:
> you would like for bare (break) and (continue) forms to receive the label of the inner-most loop instead of the default label.

This is what I meant.

stylewarning:
> In this case, however, my disagreement with this principle doesn't have any bearing. Although I can imagine somebody momentarily siezed by madness thinking its a good idea to break while iterating via map or fold, I doubt very much it is in general ever going to be good practice. 

Why not? Maybe you're looping over a set of iterator values and each of those values is searched for via an FMAP-like call. I do it all the time in Lisp when I need to exit conditional on a property of the interior of a structure.

macrologist:
@eliaslfox

macrologist:
@eliaslfox

eliaslfox (approved):
This looks great 🎉

Issue #985, opened 2023-09-02
Unexpected performance of coalton while running `(microbench1:run)`


digikar99:
Hello,

Something seems off regarding performance and inlining in the current master branch version of coalton, or may be I'm missing something. 

I started SBCL with `rlwrap sbcl --no-userinit --load quicklisp/setup.lisp`:

```lisp
* (asdf:load-system "coalton" :force t)
...
* (ql:quickload "small-coalton-programs")
To load "small-coalton-programs":
  Load 1 ASDF system:
    small-coalton-programs
; Loading "small-coalton-programs"
[package small-coalton-programs]..................
[package small-coalton-programs.primes-native]....
[package small-coalton-programs.primes-iterator]..
[package freecalc]................................
[package lisp-microbench1]........................
[package coalton-microbench1].....................
[package microbench1].
("small-coalton-programs")
* (microbench1:run)
Lisp took 37084 ms
Coalton took 25723 ms
Coalton was faster by 11361 ms (30.64%)
The hand-optimized function took 8020 ms
NIL
```

I expected coalton to take about the same time as the hand-optimized function, but it took significantly longer and seems closer to the lisp version than the hand-optimized version. (I also set the minimum and maximum clock speeds for my CPU at 0.8GHz.)

So, I checked the disassembly of the `add-two-caller` below, and it turns out to be not inlined:

```lisp
(in-package :coalton-user)

(named-readtables:in-readtable coalton:coalton)

(coalton-toplevel
  (declare add-two (single-float -> single-float))
  (define (add-two x) (+ x 2.0)))

(cl:defun add-two-caller (x)
  (cl:declare (cl:optimize cl:speed (cl:safety 0) (cl:debug 0))
              (cl:type cl:single-float x))
  (add-two x))
```

```lisp
(cl:disassemble 'add-two-caller)
; disassembly for ADD-TWO-CALLER
; Size: 13 bytes. Origin: #x54078956                          ; ADD-TWO-CALLER
; 56:       B902000000       MOV ECX, 2
; 5B:       FF7508           PUSH QWORD PTR [RBP+8]
; 5E:       E9DFCE36FC       JMP #x503E5842                   ; #<FDEFN ADD-TWO>
```

Am I missing something here - or is this a (known) bug?

stylewarning:
Did you try building with [`:coalton-release`](https://github.com/coalton-lang/coalton/blob/main/docs/coalton-lisp-interop.md#interaction-mode) on your `*features*`?

[...]

PR #987, opened 2023-09-13
If the method is declared inline, then inline its method instances

digikar99:
Here is an attempt to fix https://github.com/coalton-lang/coalton/issues/985

The idea is this: if a method is known to have been declared inline, then while defining the method instances, we also proclaim the method instance to be inline. If the method is not declared inline, then its method instances will not be declared inline either. Thus this maintains the default lispy behavior of not requiring call sites to be recompiled when the function being called is redefined, but allows for inlining when the user demands so.

One problem I'm facing is that the inlining works as expected on the first compile-load. But when the lisp image is subsequently restarted, and coalton is loaded without compilation again, then the inlining does not work. EDIT: I got why this is happening - I'm proclaiming during a macroexpansion, so the effects do not reoccur after the expanded macro form is loaded. But, at the moment, I don't know what would be the appropriate place to make the change proposed in this PR.

### DIFF
--- a/library/math/num.lisp
+++ b/library/math/num.lisp
@@ -229,6 +229,7 @@
            (Some x))))
 
 (cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:declaim (inline + - * fromInt))
   (cl:defmacro define-num-float (type lisp-type)
     "Define `Num' for TYPE"
 
--- a/src/typechecker/define-instance.lisp
+++ b/src/typechecker/define-instance.lisp
@@ -53,6 +53,14 @@
         :for unparsed-instance :in unparsed-instances
         :collect (typecheck-instance instance unparsed-instance env file)))
 
+(defun method-inline-p (method-name)
+  #+sbcl
+  (multiple-value-bind (type locally-bound-p decl)
+      (sb-cltl2:function-information method-name)
+    (and (eq type :function)
+         (null locally-bound-p)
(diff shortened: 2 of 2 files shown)


eliaslfox:
I don't think this is the best way to handle instance method inlining. For one thing class method functions are [always declared inline](https://github.com/coalton-lang/coalton/blob/main/src/codegen/codegen-class.lisp#L68). So I think [this predicate](https://github.com/coalton-lang/coalton/pull/987/files#diff-fe2693a3c95bcfd4e407d54b7e1705778fb52006caa2ca24c7917fc65751f1caR128) will always evaluate to `t`.

I also don't like the idea of Coalton's codegen being dependent on inspecting the host lisp environment. The compiler output is currently an almost pure function of the input source form(s) and the global environment. This makes debugging and especially finding minimal reproductions of bugs much easier.

Lastly, because the compiler runs in a macro, any side effects is preforms won't be repeated when its output is dumped to a fasl and reloaded. Performing side effects requires generating code to do so at load time.

I think the immediate fix for #985 is to explicitly mark some stdlib method functions as inline.

Eventually we should have `(inline)` declarations in the language proper so that they can be tracked in the global environment:
```
(define-instance (Num Integer)
  (inline)
  (define (+ a b) ...)
   ...)
```

digikar99:
Thank you for reviewing! I agree with all of it, so I'm closing this PR, and will wait for the better fix of having a `coalton:inline` declaration.

PR #988, opened 2023-09-14
Removed deprecated Early Types doc

Izaakwltn:
A large portion of this doc was deprecated
- `Lisp-object` no longer exists
- many of the filenames were no longer accurate

It also seems that at this stage the Early Types are well-rooted in the implementation and don't necessarily need amendments.


eliaslfox (approved):

stylewarning:
> It also seems that at this stage the Early Types are well-rooted in the implementation and don't necessarily need amendments.

I don't agree with this necessarily. I would have much rather seen the document updated to be accurate than be deleted.

Izaakwltn:
I'm happy to draft an updated version this week.

PR #991, opened 2023-09-20
When and Unless body type ignored; implicitly Unit

macrologist:
This PR makes a small change to the typechecker code for `when` and `unless` forms;  it ignores the inferred type of the bodies of both, keeping the type of the overall expression as `Unit`. 

This is an ergonomic change - and I'm happy to hear a good reason why it (or something like it that achieves the same developer-frendliness) ought not be included. 

Concretely, what was once this:

```
(when (< (cell:read x) 10) 
   (cell:swap! x 10)
   Unit)
```

becomes 

```

(when (< (cell:read x) 10) 
   (cell:swap! x 10))

```

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -339,12 +339,14 @@ to `len`."
          (vector:push! elem acc)
          (when (== len (vector:length acc))
            (vector:push! (vector:copy acc) result)
-           (vector:clear! acc)))
+           (vector:clear! acc) ))
        (iter:into-iter vec))
       ;; push the last if non-empty
+
       (unless (== 0 (vector:length acc)) 
         (vector:push! acc result)
         Unit)
+
       result))
 
   (declare build-cumulative-size-table (vector:Vector (Seq :a) -> vector:Vector UFix))
@@ -427,10 +429,21 @@ table of relaxed nodes may be incaccurate."
   (define (make-relaxed-node ht branches)
     (RelaxedNode ht (perfect-subtree-size-at-height ht) (build-cumulative-size-table branches) branches))
   
+  (define (pop-while vec pred)
+    (when (pred vec)
+      (vector:pop! vec)
+      (pop-while vec pred)))
+
   (declare make-node-from-branches (types:runtimerepr :a => vector:Vector (Seq :a) -> Seq :a))
   (define (make-node-from-branches branches)
     "Makes a `Seq` tall enough to contain all the branches. Branches are
 all assumed to have the same height."
+
+    (pop-while branches
+               (fn (v)
+                 (and (< 0 (vector:length v))
+                      (== 0 (virtual-size (vector:last-unsafe v))))))
+
     (let branches-length = (vector:length branches))
     (cond ((== 0 branches-length)
            (new))
@@ -444,6 +457,7 @@ all assumed to have the same height."
                        (group-vectors max-branching branches))
                      (taller-branches
                        (map (make-relaxed-node (+ ht 1)) groups)))
+
                  (make-node-from-branches taller-branches))))))
 
   (define (copy seq)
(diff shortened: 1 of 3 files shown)


stylewarning (approved):
I'm ok with it. @eliaslfox ?

eliaslfox:
I don't think this is valid, see:

```
(coalton-toplevel
  (declare x Unit)
  (define x (when True 1)))
```

This would type check and define `x` as `1`.

macrologist:
I opted to add a collection of new node types called `node-unit-body` that is just like the `node-body` nodes except it doesn't have a slot for its last node. This node captures the idea of a sequence of expressions whose last type is ignored. 

`when` and `unless` have been refactored to use `node-unit-body`. If this is acceptable, I will recycle this same node type in my looping PR as the bodies of loops are also all unit typed.

eliaslfox:
If you do this then `when` and `unless` can't be used for tail recursion.

macrologist:
Why not?

macrologist:
As I refuse to supply useless `Unit`s in the future when the syntax alone insists on `Unit` typed forms, as it does for `when` and `unless`, but also as it will for `loop` et al, I'd appreiciate some pointers. 

I'm trying to avoid the obvious and thoroughly lispy way to accomplish this: just generate a Unit value at parse time and insert it into the parse tree even though there was no corresponding Unit represented in the source. Such a "node from nowhere" would violate conventions that I'm trying to respect.

Inserting a Unit at codegen time (in translate-expression) would require ignoring the body type of `when` and `unless` during typechecking (otherwise, the there's no point). I can't tell if emitting tyepchecker nodes with wrongly typed bodies is acceptable or not, even if just until the typechecker tree hits codgen.

If it is, then that's what I'll do. If not... then I'm not sure how best to achieve this truly tiny change in a way that is both semantically correct and the implementation of which conforms to convention.

eliaslfox:
I don't think there's a valid way to insert a Unit without breaking tail calls. I also don't think there's a valid way to pretend the return type is Unit when it isn't.

macrologist:
I think the reason this smells bad to me is this.

The expression

```
(when (condition?) (do-stuff))

```

is always `Unit`. 

If `(do-suff)` happens to not be `Unit`, then we should not get a type error. It is not a matter of pretending: when I see a `when` I already know the type of this form. The compiler already knows it too: that's why we return `tc:*unit-type*` instead of `body-ty` during typechecking. 

Since everybody already knows the type of this `when` form, why should I need to include a form that returns `Unit` excplictly?

stylewarning:
I wonder if we could have an explicit idea of a value not being returned (and whether this would even help here...). The tail call fact is very important, but also WHEN is used a lot with side effects and early returns, making it more of a statement than an expression in a lot of cases. Maybe it should be recognized as a statement (say, as syntax in a PROGN) and not an expression? If so, we'd have

```
<progn> := 
(progn <control>* <expr>)

<control> := 
  (when ...)
| (unless ...)
| (if ...)
| (return ...)
| (discard <expr>)
```

in this case, the typing rules could be made so WHEN can have anything in it, at the expense that WHEN is no longer a valid expression.

stylewarning:
(we do something similar for `(let x = y)` syntax)

eliaslfox:
> Since everybody already knows the type of this when form, why should I need to include a form that returns Unit explicitly?

Because you need to return Unit at runtime.

eliaslfox:
> in this case, the typing rules could be made so WHEN can have anything in it, at the expense that WHEN is no longer a valid expression.

This would work. I think we would still want to allow `if` in tail position.

macrologist:
> Because you need to return Unit at runtime.

But I already know that. Unit is a type with one value, and just as I already know the type before anything runs it follows that I already know the value. I'm not really trying to escape expressiondom, I'm trying to make a compiler that is nice to users.

Setting aside syntax ala `let x = y`, and thinking back to this issue of tail calls:

Couldn't I just make a decision about whether to insert a Unit at codegen time by examining the type of the last node at typechecking time, storing it in a slot, then passing it along to codegen. At codegen, I look to see if the last form returns unit, if it does, then I leave it alone, if it does not, then I insert a unit node?

macrologist:
However, and in accord with @stylewarning 's suggestions. I do think that `when` and its ilk are poor expressions. I almost suggested we remove them altogether and content ourselves with if and user-code macros that wrap if for those times when `when` is really desired. They don't add anything to the language; they're really just syntactic conveniences that aren't as convenient as they should be.

macrologist:
Closing this and opening discussions about some of topics raised. ('Oops' version of comment)

PR #997, opened 2023-09-26
Bugfix in sequence branch rebalance step

macrologist:
There was a bug that would produce invalid empty branches during rebalancing that produced an array indexing error.

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -339,12 +339,14 @@ to `len`."
          (vector:push! elem acc)
          (when (== len (vector:length acc))
            (vector:push! (vector:copy acc) result)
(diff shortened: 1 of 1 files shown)


eliaslfox (approved):
Can you add a regression test?

library/seq.lisp
```diff
@@ -339,12 +339,14 @@ to `len`."
          (vector:push! elem acc)
          (when (== len (vector:length acc))
            (vector:push! (vector:copy acc) result)
-           (vector:clear! acc)))
+           (vector:clear! acc) ))
```
stylewarning, line 342: nit

stylewarning (approved):

PR #1005, opened 2023-10-05
    don't specialize Lisp vector underlying VECTOR/SLICE :A

stylewarning:
We remove the type specialization of the VECTOR :A representation
    because it provides no (or worse) speed benefit, and only sometimes
    a space benefit. The space benefit is actually unpredictable from
    the programmer's perspective (it's not obvious when an array will
    specialize), and fragile (changing the vector type might change
    the space performance).
    
    Making this change is also beneficial because the user no longer
    has to write "RuntimeRepr" every time they use vectors.


stylewarning:
Still need to ensure I didn't break the tests... They don't work locally.

colescott (approved):

PR #1011, opened 2023-10-06
Synchronized names for uninintialized slot reporters

jbouwman:
util:require raises a condition that reports the slot name for for uninitialized slots: make sure that all cases match the real slot name


macrologist (approved):
We should add a contributing to coalton heading in the readme that specifically endorses copy-and-paste development

PR #1012, opened 2023-10-10
Move pi to trig and e to exp, fix atan2

Izaakwltn:
There are some types that need `pi` or `e` but shouldn't need `Complex` or `Polar` in order to satisfy `Elementary` to get them. 

This shouldn't break external code, since Elementary will be a marker class and will still contain them.

@stylewarning 

I've also checked this against `cl-quil/discrete` and it doesn't seem to have broken anything.

### DIFF
--- a/library/big-float/impl-default.lisp
+++ b/library/big-float/impl-default.lisp
@@ -121,7 +121,9 @@
 
   (declare ilog2-abs (Integer -> Integer))
   (define (ilog2-abs x)
-    (- (bit-length (abs x)) 1))
+    (- (bit-length (abs x)) 1)))
+
+(coalton-toplevel
 
   (define-type Big-Float
     (BFValue Dyadic)
@@ -540,10 +542,13 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
   (define (make-recip-results n start f)
     "Reciprocal of `make-result`."
     (recip-result
-     (eval-series start (fn (m) (f (toInteger m))) n)))
+     (eval-series start (fn (m) (f (toInteger m))) n))))
+
+(coalton-toplevel
 
   ;; XXX: We can pre-compute some values here
-  (define (bf-pi _)
+  (declare bf-pi (Unit -> Big-Float))
+  (define (bf-pi)
     "Return the value of pi to the currently set precision."
     ;; Chudnovsky algorithm
     (progn
@@ -675,8 +680,11 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
     (define (asin x)
       (atan (* x (reciprocal-sqrt (- 1 (* x x))))))
     (define (acos x)
-      (- (scale (bf-pi) -1) (asin x))))
+      (- (scale (bf-pi) -1) (asin x)))
+    (define pi (BFconst bf-pi))))
 
+(coalton-toplevel
+  
   (define-instance (Exponentiable Big-Float)
     (define (exp x)
       (let prec = (get-precision))
@@ -760,7 +768,13 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
         ((Tuple x (BFConst f)) (pow x (f)))
         (_ (exp (* y (ln x))))))
     (define (log b x)
-      (/ (ln x) (ln b))))
+      (/ (ln x) (ln b)))
+    (define ee (BFConst bf-ee)))
+
+  (declare bf-ee (Unit -> Big-Float))
+  (define (bf-ee)
+    "Return the value of ee = exp(1) to the currently set precision."
+    (exp 1))
 
   (define-instance (Radical Big-Float)
     (define (sqrt x)
@@ -783,14 +797,7 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
     (define (polar z)
       (Tuple (magnitude z) (phase z))))
 
-  (declare bf-ee (Unit -> Big-Float))
-  (define (bf-ee)
-    "Return the value of ee = exp(1) to the currently set precision."
-    (exp 1))
-
-  (define-instance (Elementary Big-Float)
-    (define pi (BFConst bf-pi))
-    (define ee (BFConst bf-ee)))
+  (define-instance (Elementary Big-Float))
 
   (define (big-float->string x)
     "Returns a Big-Float in scientific notation."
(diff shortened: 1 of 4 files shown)


stylewarning (changes requested):

library/big-float/impl-default.lisp
```diff
@@ -760,8 +761,13 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
...
         (_ (exp (* y (ln x))))))
     (define (log b x)
-      (/ (ln x) (ln b))))
+      (/ (ln x) (ln b)))
+    (define ee (BFConst (exp 1))))
 
+  (define (bf-ee)
+    "Return the value of ee = exp(1) to the currently set precision."
+    (BFConst (exp 1)))
+
```
stylewarning: I think these are still incorrect. I believe it should be:

```
(declare bf-ee (Unit -> Big-Float))
(define (bf-ee)
  (exp 1))

...
(define ee (BFConst bf-ee))
```

Is my understanding correct that this is the right approach?

library/big-float/impl-sbcl.lisp
```diff
@@ -255,6 +255,12 @@
 (coalton-library/math/complex::%define-standard-complex-instances Big-Float)
 
 (coalton-toplevel
+  
+  (define (bf-pi _)
```
stylewarning, line 259: The `_` shouldn't be necessary.

library/big-float/impl-sbcl.lisp
```diff
@@ -274,8 +280,18 @@
         (cl:values (sb-mpfr:acos x))))
     (define (atan x)
       (lisp Big-Float (x)
-        (cl:values (sb-mpfr:atan x)))))
-
+        (cl:values (sb-mpfr:atan x))))
+    (define pi
+      "Return the value of pi to the currently set precision."
+      (bf-pi)))
+  ;; BUG: Pi is calculated just once, so if we change precision,
```
stylewarning, line 287: put this comment above `(define pi ...)`

library/big-float/impl-default.lisp
```diff
@@ -675,7 +675,8 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
     (define (asin x)
       (atan (* x (reciprocal-sqrt (- 1 (* x x))))))
     (define (acos x)
-      (- (scale (bf-pi) -1) (asin x))))
+      (- (scale (bf-pi) -1) (asin x)))
+    (define pi (BFConst (bf-pi))))
```
stylewarning, line 679: The purpose of `BFConst` is to capture a _function_ which computes a constant value. We can't `(define pi (bf-pi))` because then some fixed precision of pi will be computed and stored, and it will never react to an increase in precision. Hence we have another kind of "big-float" which is a [`BFConst` which holds on to a function `Unit -> Big-Float`](https://github.com/coalton-lang/coalton/blob/main/library/big-float/impl-default.lisp#L129) which (re-)computes the constant value at the time it's used.

library/big-float/impl-default.lisp
```diff
@@ -760,37 +761,26 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
         ((Tuple x (BFConst f)) (pow x (f)))
         (_ (exp (* y (ln x))))))
     (define (log b x)
-      (/ (ln x) (ln b))))
+      (/ (ln x) (ln b)))
+    (define ee (BFConst (exp 1))))
```
stylewarning, line 765: See comment above. This should be `(BFConst bf-ee)`.

library/big-float/impl-sbcl.lisp
```diff
@@ -299,7 +315,12 @@
         (True (/ (ln x) (ln n)))))
     (define (ln x)
       (lisp Big-Float (x)
-        (cl:values (sb-mpfr:log x)))))
+        (cl:values (sb-mpfr:log x))))
+    (define ee 
+      "Return the value of ee = exp(1) to the currently set precision."
+      (bf-ee)))
+  ;; BUG: EE is calculated just once, so if we change precision,
```
stylewarning, line 322: put the comment above `(define ee...`)

stylewarning (changes requested):
Does this current rev actually type check? It seems like a type error to call `BFConst` on something that's not `Unit -> Big-Float`. See comments herein.

stylewarning:
I'm concerned about why this program is type checking... `BFConst :: (Unit -> Big-Float) -> Big-Float` but you're providing it `Big-Float` in certain places. Are the tests passing because MPFR is being loaded (which doesn't have this same issue) and not the fallback?

stylewarning:
Sorry for the noise. The tests are failing for Allegro. You can try testing with `:coalton-portable-bigfloat` in your `*features*` before you compile Coalton.

library/big-float/impl-default.lisp
Izaakwltn: The issue I'm running into is that `ee` is a member of Exponentiable, in line with this PR. This causes issues if I define `(bf-ee)` before defining the instance:

```
(define (bf-ee)
  (exp 1)) ;; exp isn't defined yet
(coalton-toplevel
   (define-instance Exponentiable Big-Float
    (define ...
    (define ...
    (define (exp ...
    (define ee
```
or after:
```
(coalton-toplevel
   (define-instance Exponentiable Big-Float
    (define ...
    (define ...
    (define (exp ...
    (define ee (BFConst bf-ee)))) ;; bf-ee isn't defined-yet

(define (bf-ee)
  (exp 1)) 
```

stylewarning: In that case, `(BFConst (fn (_) (exp 1)))`

Izaakwltn:
Tests are all passing now, and default Big-Float `pi` and `ee` are guaranteed to be necessary precision

library/big-float/impl-default.lisp
```diff
@@ -760,9 +768,15 @@ returns the nth SeriesSplit, return the series evaluated to the Nth element."
         ((Tuple x (BFConst f)) (pow x (f)))
         (_ (exp (* y (ln x))))))
     (define (log b x)
-      (/ (ln x) (ln b))))
+      (/ (ln x) (ln b)))
+    (define ee (BFConst bf-ee)))
+
+  (declare bf-ee (Unit -> Big-Float))
```
stylewarning, line 774: nit: this can be deleted i think

library/big-float/impl-sbcl.lisp
stylewarning: this _ is still here

stylewarning (changes requested):
a few little things but I think they ought to be addressed for standard library cleanliness

library/big-float/impl-default.lisp
Izaakwltn: The declare? When I take it out it fails to load

stylewarning (approved):

Issue #1008, opened 2023-10-06
(Complex Single-Float) should have a repr CL::(COMPLEX SINGLE-FLOAT)


stylewarning:
The current implementation is not wrong, but the `repr` is `(COMMON-LISP:OR COMMON-LISP:NUMBER C:COMPLEX)`, which means any sort of (Lisp) compiler optimization is hosed. This is basically the same issue as specialized arrays, but doesn't deserve as much ceremony as specialized arrays, and there are only two specialized types worth caring about: singles and doubles. We should work to make Coalton emit these specialized types.

digikar99:
How easy or difficult would it be to make `specialize` work for types? For example - 

```lisp
(in-package :coalton-user)

(coalton-toplevel
  (repr :native (or cl:number (cl:complex *)))
  (define-type (complx :a)))

(coalton-toplevel
  (repr :native (cl:complex single-float))
  (define-type complx/single-float)
  (specialize complx complx/single-float (complx single-float)))
```

Currently, the above results in an error that `complx` is an unknown function or variable. But if `specialize` also worked for types, then the case of complex as well as specialized-vectors/arrays seems solved (?).

stylewarning:
@digikar99 I think we can have something similar-ish, where at the definition of `complx` we can specify different ways `:a` substitutions can lead to optimized representations. See the draft PR for example details where a proof of concept of the mechanics were implemented, but the syntax is still TBD.

digikar99:
Yup, this looks good! 

On a sidenote, I wonder if coalton should have tagged versions, and/or separate main and dev branches to avoid breaking things for coalton users.

stylewarning:
> Yup, this looks good! 
> 
> 
> 
> On a sidenote, I wonder if coalton should have tagged versions, and/or separate main and dev branches to avoid breaking things for coalton users.

Eventually we'll tag versions when things sort of converge. We know we can't keep breaking things forever.

stylewarning:
@YarinHeffes This can be closed, yes?

PR #1015, opened 2023-10-11
add `(LispArray :t)`, add specialized representations of `LispArray` and `Complex`

stylewarning:
This PR does three things:

- Adds a `LispArray` type to match the semantics of `(simple-array <t> (*))`. I think this is a good compromise between my concerns and @eliaslfox's suggestions. The name `LispArray` was chosen intentionally to emphasize this behavior, and not place itself as a "competitor" to `Vector` or `Seq`.
- Adds specialization of `(Complex :t)` type.
- Adds specialization of `(LispArray :t)` type.

Addresses #1008 in principle.

```
                (declare f ((coalton-library/math/complex:Complex Single-Float) -> Single-Float))
                (define (f z)
                  (coalton-library/math/complex:square-magnitude z))
                
                (define (g z)
                  (coalton-library/math/complex:square-magnitude z)))
firing type specialization: (COALTON-LIBRARY/MATH/COMPLEX:COMPLEX #T20689) => (COMMON-LISP:OR COMMON-LISP:NUMBER COALTON-LIBRARY/MATH/COMPLEX:COMPLEX)
firing type specialization: (COALTON-LIBRARY/MATH/COMPLEX:COMPLEX COALTON:SINGLE-FLOAT) => (COMMON-LISP:COMPLEX COMMON-LISP:SINGLE-FLOAT)
(COMMON-LISP:PROGN
 (COMMON-LISP:DECLAIM
  (SB-EXT:MUFFLE-CONDITIONS SB-KERNEL:REDEFINITION-WARNING))
 (COALTON-IMPL/GLOBAL-LEXICAL:DEFINE-GLOBAL-LEXICAL G
                                                    COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY)
 (COMMON-LISP:DEFUN G (#:DICT1872 Z-3371)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE #:DICT1872 Z-3371)
                        (COMMON-LISP:TYPE
                         COALTON-LIBRARY/MATH/COMPLEX::CLASS/COMPLEX
                         #:DICT1872)
                        (COMMON-LISP:TYPE
                         (COMMON-LISP:OR COMMON-LISP:NUMBER COMPLEX) Z-3371)
                        (COMMON-LISP:VALUES COMMON-LISP:T
                                            COMMON-LISP:&OPTIONAL))
   (COALTON-LIBRARY/MATH/COMPLEX:SQUARE-MAGNITUDE #:DICT1872 Z-3371))
 (COMMON-LISP:SETF G (COALTON-IMPL/RUNTIME/FUNCTION-ENTRY::F2 #'G))
 (COALTON-IMPL/GLOBAL-LEXICAL:DEFINE-GLOBAL-LEXICAL F
                                                    COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY)
 (COMMON-LISP:DEFUN F (Z-3370)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE Z-3370)
                        (COMMON-LISP:TYPE
                         (COMMON-LISP:COMPLEX COMMON-LISP:SINGLE-FLOAT) Z-3370)
                        (COMMON-LISP:VALUES COMMON-LISP:SINGLE-FLOAT
                                            COMMON-LISP:&OPTIONAL))
   (COALTON-LIBRARY/MATH/COMPLEX:SQUARE-MAGNITUDE
    COALTON-LIBRARY/MATH/COMPLEX::|INSTANCE/COMPLEX SINGLE-FLOAT| Z-3370))
 (COMMON-LISP:SETF F (COALTON-IMPL/RUNTIME/FUNCTION-ENTRY::F1 #'F))
 (COMMON-LISP:DECLAIM
  (SB-EXT:UNMUFFLE-CONDITIONS SB-KERNEL:REDEFINITION-WARNING))
 (COMMON-LISP:VALUES))
```

### DIFF
--- a/library/lisparray.lisp
+++ b/library/lisparray.lisp
@@ -1,0 +1,78 @@
+;;;; lisparray.lisp
+;;;;
+;;;; An interface to Common Lisp rank-1 SIMPLE-ARRAYs.
+
+(coalton-library/utils:defstdlib-package #:coalton-library/lisparray
+  (:use #:coalton)
+  (:local-nicknames
+   (#:types #:coalton-library/types))
+  (:export
+   #:LispArray
+   #:make
+   #:make-uninitialized
+   #:length
+   #:aref
+   #:set!))
+
+(in-package #:coalton-library/lisparray)
+
+(named-readtables:in-readtable coalton:coalton)
+
+#+coalton-release
+(cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
+
+(coalton-toplevel
+  ;; The representation of (LispArray :t) is specially dealt with by
+  ;; the compiler in lisp-type.lisp.
+  (define-type (LispArray :t)
+    "A one-dimensional, non-resizable array of elements.
+
+These arrays are represented as possibly specialized `(cl:simple-array <type> (cl:*))` and are meant to be used as a tool either to interface with Lisp code or to implement efficient data structures. One should consult `Vector` or `Seq` for more general sequential data structure needs.
+
+Whether or not the arrays are specialized depends on the underlying Lisp implementation. Consult `cl:upgraded-array-element-type` to determine whether `LispArray` may get specialized.")
+
+  (declare make (types:RuntimeRepr :t => UFix -> :t -> LispArray :t))
+  (define (make n x)
+    "Make a new `LispArray` of length `n` initialized to `x`.
+
+If the type of `x` represents a specialized array "
+    ;; FIXME: how can we get this statically?
+    (let ((type (types:runtime-repr (types:proxy-of x))))
+      (lisp (LispArray :t) (n x type)
+        (cl:make-array n :element-type type :initial-element x))))
+
+  (declare make-uninitialized (types:RuntimeRepr :t => UFix -> LispArray :t))
+  (define (make-uninitialized n)
+    "Make a new LispArray of length `n` that can store elements of type `:t`.
+
+WARNING: The consequences are undefined if an uninitialized element is read before being set.
+"
+    (let p = types:Proxy)
+    (let p_ = (types:proxy-inner p))
+    (let type = (types:runtime-repr p_))
+    (types:as-proxy-of
+     (lisp (LispArray :t) (n type)
+       (cl:make-array n :element-type type))
+     p))
+
+  (declare length (LispArray :t -> UFix))
+  (define (length v)
+    "Return the length of the `LispArray` `v`."
+    (lisp UFix (v)
+      (cl:length v)))
+
+  (declare aref (LispArray :t -> UFix -> :t))
+  (define (aref v i)
+    "Read the `i`th value of the `LispArray` `v`."
+    (lisp :t (v i)
+      (cl:aref v i)))
+
+  (declare set! (LispArray :t -> UFix -> :t -> Unit))
+  (define (set! v i x)
+    "Set the `i`th value of the `LispArray` `v` to `x`."
+    (lisp Unit (v i x)
+      (cl:setf (cl:aref v i) x)
+      Unit)))
+
+#+sb-package-locks
+(sb-ext:lock-package "COALTON-LIBRARY/LISPARRAY")
--- a/library/math/complex.lisp
+++ b/library/math/complex.lisp
@@ -25,11 +25,18 @@
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
 
 (coalton-toplevel
-  (repr :native (cl:or cl:number complex))
+  ;; The representation of (Complex :t) is specially dealt with by the
+  ;; compiler in lisp-type.lisp.
   (define-type (Complex :a)
     "Complex number that may either have a native or constructed representation."
     (%Complex :a :a))
+)
 
+;; Quirk: We had to split the above COALTON-TOPLEVEL from the bottom
+;; one because Allegro needs to know about Complex before it gets used
+;; as a Lisp type in codegen. SBCL and CCL tolerate it fine.
+
+(coalton-toplevel
   (define-class (Num :a => Complex :a)
     (complex (:a -> :a -> (Complex :a)))
     (real-part (Complex :a -> :a))
--- a/src/typechecker/lisp-type.lisp
+++ b/src/typechecker/lisp-type.lisp
@@ -15,6 +15,31 @@
 
 (in-package #:coalton-impl/typechecker/lisp-type)
 
+(defun lisp-type= (x y)
+  (and (subtypep x y)
+       (subtypep y x)))
+
+;;; This is a list of Coalton tycon names that (1) have a finite
+;;; number of specialized representations, and (2) won't yet exist
+;;; when this file is first compiled.
+
+;;; Complex
+(defun complex-tycon-p (tycon)
+  (if (find-package "COALTON-LIBRARY/MATH/COMPLEX")
+      (eq tycon (find-symbol "COMPLEX" "COALTON-LIBRARY/MATH/COMPLEX"))
+      nil))
+
+(defvar *specialized-complex-part-types-considered*
+  '(cl:single-float
+    cl:double-float))
+
+;;; Simple Arrays
+(defun lisparray-tycon-p (tycon)
+  (if (find-package "COALTON-LIBRARY/LISPARRAY")
+      (eq tycon (find-symbol "LISPARRAY" "COALTON-LIBRARY/LISPARRAY"))
+      nil))
+
+
 ;;;
 ;;; Lisp types for coalton types
 ;;;
@@ -30,7 +55,6 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
     't)
 
   (:method ((ty tycon) env)
-    (declare (ignore))
     (let*
         ((tcon-name (tycon-name ty))
          (type-entry (lookup-type env tcon-name :no-error t)))
@@ -48,6 +72,40 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
       ;; If we are a function, emit a function type
       ((function-type-p ty)
        'function-entry)
+
+      ((typep (tapp-from ty) 'tycon)
+       (let ((from (tycon-name (tapp-from ty)))
+             (to   (tapp-to ty)))
+         (cond
+           ;; First we deal with specialized parametric types.
+           ;;
+           ;; (Complex :t)
+           ((complex-tycon-p from)
+            (let ((lisp-to (lisp-type to env)))
+              (cond
+                ((member lisp-to
+                         *specialized-complex-part-types-considered*
+                         :test #'lisp-type=)
+                 `(cl:complex ,lisp-to))
+                (t
+                 ;; We can use FROM directly since it'll be the name
+                 ;; of the Complex tycon, which matches the Lisp type
+                 ;; exactly.
+                 `(cl:or cl:number ,from)))))
+
+           ;; (LispArray :t)
+           ((lisparray-tycon-p from)
+            ;; Subtle observation: We don't want to recursively call
+            ;; LISP-TYPE on a TYVAR, because that will resolve to
+            ;; `cl:t`, which isn't the correct type of
+            ;; arbitrarily-typed arrays.
+            (if (typep to 'tyvar)
+                `(cl:simple-array cl:* (cl:*))
+                `(cl:simple-array ,(lisp-type to env) (cl:*))))
+           
+           ;; Otherwise we fall back.
+           (t
+            (lisp-type (tapp-from ty) env)))))
 
       ;; Otherwise, emit the applied type
       (t
--- a/coalton.asd
+++ b/coalton.asd
@@ -168,6 +168,7 @@
                (:file "optional")
                (:file "result")
                (:file "tuple")
+               (:file "lisparray")
                (:file "list")
                (:file "vector")
                (:file "char")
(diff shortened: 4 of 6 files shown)


stylewarning:
I thought maybe something like

```
(repr :specialize (Complex Single-Float) :as (cl:complex cl:single-float)
      :specialize (Complex Double-Float) :as (cl:complex cl:double-float)
      :specialize (Complex :a) :as (cl:or cl:number Complex))
```

as syntax for specific parameter specializations. Maybe could be more flexible by binding lambda functions or something, but I'm not sure it needs to be that ostentatious.

library/math/complex.lisp
```diff
@@ -24,8 +26,21 @@
...
+           (cl:lambda (ty)
+             ;; (cl:format cl:t "testing ~S~%" ty)
+             (cl:typecase ty
+               (ty:tycon
+                (cl:case (ty:tycon-name ty)
+                  (coalton:single-float (cl:values '(cl:complex cl:single-float) cl:t))
+                  (coalton:double-float (cl:values '(cl:complex cl:double-float) cl:t))
+                  (cl:otherwise (cl:values '(cl:or cl:number Complex) cl:t))))
+               (cl:t
+                (cl:values '(cl:or cl:number Complex) cl:t))))))
```
digikar99, line 41: It should be `(cl:or cl:number cl:complex)` instead of `(cl:or cl:number Complex)` (?)

eliaslfox:
I would prefer to hard code these in `lisp-type` instead of adding an extension point. Common Lisp has a fairly small number of cases where this is necessary .

library/math/complex.lisp
eliaslfox: It's [this complex](https://github.com/coalton-lang/coalton/blob/main/library/math/complex.lisp#L29)

library/math/complex.lisp
```diff
@@ -24,8 +26,21 @@
...
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
 
+;;; Tell Coalton that we should expect Complex to have a special type representation.
+(cl:eval-when (:compile-toplevel :load-toplevel :execute)
+  (cl:setf (cl:gethash 'Complex coalton-impl/typechecker/lisp-type:*special-type-representations*)
+           (cl:lambda (ty)
+             ;; (cl:format cl:t "testing ~S~%" ty)
+             (cl:typecase ty
+               (ty:tycon
+                (cl:case (ty:tycon-name ty)
```
eliaslfox, line 36: This should look at the lisp type of it's argument so that newtypes are handled correctly.

digikar99:
> Common Lisp has a fairly small number of cases where this is necessary .

Within the Common Lisp builtins, the number of cases is small. But the extensibility looks important in the wider lisp ecosystem, although you can achieve the same things with a typeclass (?).

For example, one use case I imagine is defining a parametric coalton type with runtime representations given by [magicl](https://github.com/quil-lang/magicl)'s tensor types. This way, a parametric coalton type `(tensor :t)` can have runtime representation given by `magicl:tensor/single-float`, `magicl:tensor/double-float`, etc. Now, instead of using the parametric type, one can also use a coalton typeclass. So, the extensibility isn't a must-have.

eliaslfox:
In that case you should use a type class. The type isn't really parametric because it can't be instantiated with arbitrary types.

stylewarning:
> I would prefer to hard code these in `lisp-type` instead of adding an extension point. Common Lisp has a fairly small number of cases where this is necessary .

I had a circularity issue: LISP-TYPE doesn't know about COMPLEX yet. Do you have any suggestions?

(I also started to sketch a language feature, but I thought it was too cumbersome for just a couple specialized parametric types, like complexes and simple-arrays.)

library/math/complex.lisp
digikar99: Oh, okay! `Complex` denotes the complex structure-class that gets defined with the coalton type.

eliaslfox:
Use find-symbol. We do it in a couple places already.

stylewarning:
The `LispArray` work will only be meaningfully completed when there's `RandomAccess` available.

library/lisparray.lisp
```diff
@@ -0,0 +1,30 @@
...
+    "A one-dimensional, non-resizable array of elements.
+
+These arrays are represented as possibly specialized `(cl:simple-array <t> (cl:*))` and are meant to be used as a flexible tool to implement efficient data structures. One should consult `Vector` or `Seq` for more general sequential data structure needs.
+
+Whether or not the arrays are specialized depends on the underlying Lisp implementation. Consult `cl:upgraded-array-element-type` to determine whether `LispArray` may get specialized.")
+
+)                                       ; COALTON-TOPLEVEL
+
+#+sb-package-locks
+(sb-ext:lock-package "COALTON-LIBRARY/MATH/COMPLEX")
```
stylewarning, line 30: oops, will fix

library/lisparray.lisp
```diff
@@ -0,0 +1,82 @@
+;;;; lisparray.lisp
+;;;;
+;;;; An interface to Common Lisp rank-1 SIMPLE-ARRAYs.
+
+(coalton-library/utils:defstdlib-package #:coalton-library/lisparray
+  (:use #:coalton)
+  (:local-nicknames (#:ram #:coalton-library/randomaccess)
```
stylewarning, line 7: I forgot to remove this.

src/typechecker/lisp-type.lisp
```diff
@@ -15,6 +15,42 @@
...
+(defvar *specialized-array-element-types-considered*
+  `(
+    ;; Float types
+    cl:single-float cl:double-float
+    ;; Complex float types
+    (cl:complex cl:single-float) (cl:complex cl:double-float)
+    ;; Integer types
+    cl:fixnum
+    (cl:and cl:fixnum cl:unsigned-byte)
+    ,@(loop :for byte-type :in '(cl:signed-byte cl:unsigned-byte)
```
colescott, line 48: I feel like explicit enumeration would be clearer and more grep-able

library/lisparray.lisp
```diff
@@ -0,0 +1,81 @@
...
+  ;; The representation of (LispArray :t) is specially dealt with by
+  ;; the compiler in lisp-type.lisp.
+  (define-type (LispArray :t)
+    "A one-dimensional, non-resizable array of elements.
+
+These arrays are represented as possibly specialized `(cl:simple-array <type> (cl:*))` and are meant to be used as a tool either to interface with Lisp code or to implement efficient data structures. One should consult `Vector` or `Seq` for more general sequential data structure needs.
+
+Whether or not the arrays are specialized depends on the underlying Lisp implementation. Consult `cl:upgraded-array-element-type` to determine whether `LispArray` may get specialized.")
+
+)                                       ; COALTON-TOPLEVEL
```
colescott, line 33: Are these toplevels split intentionally? If so, then I believe a comment is warranted

src/typechecker/lisp-type.lisp
```diff
@@ -49,6 +84,40 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
        'function-entry)
 
+      ((typep (tapp-from ty) 'tycon)
+       (let ((from (tycon-name (tapp-from ty)))
+             (to   (tapp-to ty)))
+         (cond
+           ;; First we deal with specialized parametric types.
+           ;;
+           ;; (Complex :t)
+           ((and (not (null (complex-tycon)))
```
colescott, line 94: I would either move the `find-symbol` business into here or have the function be something like `complex-type-p` which does this full check as a single predicate.

library/lisparray.lisp
```diff
@@ -0,0 +1,81 @@
...
+
+(coalton-toplevel
+  ;; The representation of (LispArray :t) is specially dealt with by
+  ;; the compiler in lisp-type.lisp.
+  (define-type (LispArray :t)
+    "A one-dimensional, non-resizable array of elements.
+
+These arrays are represented as possibly specialized `(cl:simple-array <type> (cl:*))` and are meant to be used as a tool either to interface with Lisp code or to implement efficient data structures. One should consult `Vector` or `Seq` for more general sequential data structure needs.
+
+Whether or not the arrays are specialized depends on the underlying Lisp implementation. Consult `cl:upgraded-array-element-type` to determine whether `LispArray` may get specialized.")
```
colescott, line 31: Would it be possible/useful to add a `specialized?` function which checks this within Coalton?

colescott (commented):
Super cool! Just a few small nits

library/lisparray.lisp
stylewarning: no, but it would be nice if we discussed the style of thus stuff

library/lisparray.lisp
stylewarning: What would this check, the array object itself? A type?

library/math/complex.lisp
```diff
@@ -25,11 +25,18 @@
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
 
 (coalton-toplevel
-  (repr :native (cl:or cl:number complex))
+  ;; The representation of (Complex :t) is specially dealt with by the
+  ;; compiler in lisp-type.lisp.
```
digikar99, line 29: Probably change `lisp-type.lisp` to `src/typechecker/lisp-type.lisp`.

digikar99:
1. If it's `aref` for array-ref, then `(setf aref)` or `aset` feels more consistent than `set!`. Not a hard requirement, so depends on what others find appropriate.

2. Are the symbols from `coalton-library/types` package supposed to be accessible from `coalton-user`? If not, and perhaps even then, is it possible to `make-unitialized` depend on its return type? So that the following is legal code

```lisp
(coalton-toplevel
  (define (make-u8-array size)
    (the (LispArray U8) (make-uninitialized size))))
```

instead of requiring the user to write the slightly esoteric

```lisp
(coalton-toplevel
  (define (make-u8-array size)
    (make-uninitialized size (the (proxy u8) proxy))))
```

3. This is probably not an issue with this PR, but Common Lisp has this quirk that the imaginary part can be float 0.0 but never the rational 0. I'm missing how Coalton is handling this.

library/lisparray.lisp
```diff
@@ -0,0 +1,78 @@
...
+    (lisp :t (v i)
+      (cl:aref v i)))
+
+  (declare set! (LispArray :t -> UFix -> :t -> Unit))
+  (define (set! v i x)
+    "Set the `i`th value of the `LispArray` `v` to `x`."
+    (lisp Unit (v i x)
+      (cl:setf (cl:aref v i) x)
+      Unit))
+  )
```
eliaslfox, line 75: formatting

library/lisparray.lisp
```diff
@@ -0,0 +1,78 @@
+;;;; lisparray.lisp
+;;;;
+;;;; An interface to Common Lisp rank-1 SIMPLE-ARRAYs.
+
+(coalton-library/utils:defstdlib-package #:coalton-library/lisparray
+  (:use #:coalton)
+  (:local-nicknames (#:ty #:coalton-library/types))
```
eliaslfox, line 7: import as `types`. Also newline after `:local-nicknames`.

library/lisparray.lisp
```diff
@@ -0,0 +1,78 @@
...
+    "Make a new `LispArray` of length `n` initialized to `x`.
+
+If the type of `x` represents a specialized array "
+    ;; FIXME: how can we get this statically?
+    (let ((type (ty:runtime-repr (ty:proxy-of x))))
+      (lisp (LispArray :t) (n x type)
+        (cl:make-array n :element-type type :initial-element x))))
+
+  (declare make-uninitialized (ty:RuntimeRepr :t => UFix -> ty:Proxy :t -> LispArray :t))
+  (define (make-uninitialized n p)
```
eliaslfox, line 44: You don't need to take a proxy here. [This trick](https://github.com/coalton-lang/coalton/blob/6fdd091375eb52ae6584b20b0165f62bf0dc6678/library/vector.lisp#L73-L76) let's you reference the functions return type.

src/typechecker/lisp-type.lisp
```diff
@@ -49,6 +93,41 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
+           ;; First we deal with specialized parametric types.
+           ;;
+           ;; (Complex :t)
+           ((complex-tycon-p from)
+            (let ((lisp-to (lisp-type to env)))
+              (cond
+                ((member lisp-to
+                         *specialized-complex-part-types-considered*
+                         :test #'lisp-type=)
+                 `(cl:complex ,lisp-to))
```
eliaslfox, line 109: the `cl:` is unnecessary

src/typechecker/lisp-type.lisp
```diff
@@ -49,6 +93,41 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
+                (t
+                 ;; We can use FROM directly since it'll be the name
+                 ;; of the Complex tycon, which matches the Lisp type
+                 ;; exactly.
+                 `(cl:or cl:number ,from)))))
+
+           ;; (LispArray :t)
+           ((lisparray-tycon-p from)
```
eliaslfox, line 117: This seems unnecessary, SBCL will handle the fallback for us.

library/lisparray.lisp
stylewarning: great trick

src/typechecker/lisp-type.lisp
stylewarning: i put it there since there's a coalton/whatever:complex, and this file deals with both lisp types and coalton tycons

eliaslfox (approved):

Issue #1013, opened 2023-10-11
Big-Float constants `ee` and `pi` don't change to match precision. 


Izaakwltn:
These variables are locked in with the initial big-float precision, and don't adjust as precision changes.

Initial precision is 53 bits so they're basically locked in as Double-Floats.

stylewarning:
note: this is only an issue with the MPFR version of the library. The portable version handles this fine.

Izaakwltn:
I've been thinking about this bug, and thought of a couple quick but potentially janky approaches for fixing it.

- Every time precision is set, redefine `Trigonometric Big-Float` and `Exponentiable Big-Float`, which will automatically update the definition of `pi` and `ee` respectively. This could be kept in separate functions like `reset-bf-constants` or `set-precision-and-constants`. Bubble-gum and paperclip approach. 
    
  - Note: Is it possible to redefine *one* member of a type class or does every redefinition have to be applied as a whole?
- Convert `Big-Float` to `MPFR-Float` and make a new `Big-Float` type in the style of `impl-default.lisp` that uses `MPFR-Float`s instead of `Dyadic`s. This is probably the nuclear option. 
- Simplifying the possibility above, give `Big-Float` two constructors that contain either an `MPFR-Float` object or a `(Unit -> MPFR-Float)` function:
```
(define-type Big-Float
  (BFValue MPFR-Float)
  (BFConst (Unit -> MPFR-Float))
```

stylewarning:
Ignoring potential language changes, the last suggestion would be the right approach.

PR #1024, opened 2023-10-18
Adding Big-Float constructors for variable prec constants

Izaakwltn:
This addresses issue #1013 

I've renamed the existing `sb-mpfr` compatible `Big-Float` type to `MPFR-Float` type, and added a new `Big-Float` type with two constructors: `(BFValue MPFR-Float)` and `(BFConst (:a -> MPFR-Float))`. 

This allows for constants to readjust to current precision.

### DIFF
--- a/library/big-float/README.md
+++ b/library/big-float/README.md
@@ -8,6 +8,12 @@ By default SBCL will use the `sb-mpfr` library.
 This depends on the MPFR library being installed, and visible within your
 dynamic linker path. Otherwise, you will get an error while compiling.
 
+When using the `sb-mpfr` library, two different float types are implemented:
+- `MPFR-Float`, a Coalton wrapper on the typical `mpfr-float`, and 
+- `Big-Float`, a Coalton type with two constructors:
+  - `(BFValue MPFR-Float)` and
+  - `(BFConst (:a -> MPFR-Float))` which allows constants to have variable precision.
+
 Before coalton is compiled set the environment
 variable`COALTON_PORTABLE_BIGFLOAT=1` or add the feature
 `:coalton-portable-bigfloat` to proceed anyways.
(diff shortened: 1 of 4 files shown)


stylewarning (changes requested):
This seems a little repetitive. How about something like

```
(define (mpfr-of bf)
  (match bf
    ((BFValue v) v)
    ((BFConst f) (f))))
```

With this, something like `==` would be:

```
(define (== a b)
  (== (mpfr-of a) (mpfr-of b)))
```

without all of the case analysis. Unary functions are easier:

```
(define (lift f x)
  (nest BFValue f mpfr-of x))
```

so you have

```
(define (sin x) (lift sin x))
```

Note that this approach does lead to some nonintuitive behavior. Some values are "privileged" while others are not. So something like

```
(define pi-1 (* 4 (atan 1))
(define pi-2  pi)

(set precision)

(== pi pi-2) ==> false
```

happens. I'm not sure if this is ok or not.

eliaslfox:
I don't agree with this solution. It adds additional overhead to all bigfloat computations. Pi and E should be made methods so that they alone incur the overhead.

Izaakwltn:
Would that be as simple as changing the class definitions from (:a) to (unit -> :a) and making sure it works for each type? 

If so I ca th

stylewarning:
the only "problem" with this, which may be minor, is that idk any programming language where something that looks like a constant is a function.

    (* (* 4/3 (pi)) (^ r 3))

is just non-intuitive, since nobody would imagine pi to be a function or method, and there's no visual indication as such.

FWIW, I don't think big float speed will be dominated by the cost of a few extra pattern matches. With that said, there's definitely a more fundamental problem here.

Izaakwltn:
Another option:

Have both `pi` & `ee` values and `c-pi` & `c-ee` compute methods.

This would allow `big-float`s to use `c-pi` & `c-ee` to recompute $\pi$ each time without disturbing the existing syntax. In addition, there may be future types that would benefit from having method constants.

Izaakwltn:
I updated Big-Float according to the initial review

PR #1025, opened 2023-10-18
check if unsigned numbers underflow

Jobhdez:
@Izaakwltn  Hopefully this is not an issue that I made another pr. I had git issues with the other one.

This solves #921.

This is the interaction I just had in this new pr:

```
(coalton-toplevel 
		(declare add-8bit (U8 -> U8 -> U8))
		(define (add-8bit x y)
		  (+ x y))
		(declare sub-8bit (U8 -> U8 -> U8))
		(define (sub-8bit x y)
		  (- x y))
		(declare mul-8bit (U8 -> U8 -> U8))
		(define (mul-8bit x y)
		  (* x y))
		(declare fint (Integer -> U8))
		(define (fint x)
		  (fromInt x)))
; No value
COALTON-USER> (coalton (add-8bit 2 3))
5
COALTON-USER> (coalton (add-8bit 3 -4))
; Evaluation aborted on #<SIMPLE-ERROR "Unsigned value underflowed 8 bits." {1007939573}>.
COALTON-USER> (coalton (sub-8bit 3 2))
1
COALTON-USER> (coalton (sub-8bit 3 5))
; Evaluation aborted on #<SIMPLE-ERROR "Unsigned value underflowed 8 bits." {100835E453}>.
COALTON-USER> (coalton (mul-8bit 3 4))
12
COALTON-USER> (coalton (mul-8bit 3 -4))
; Evaluation aborted on #<SIMPLE-ERROR "Unsigned value underflowed 8 bits." {1008C8C2B3}>.
COALTON-USER> (coalton (fint 4))
4
COALTON-USER> (coalton (fint -2))
; Evaluation aborted on #<SIMPLE-ERROR "Unsigned value underflowed 8 bits." {10017D9693}>.
```

Thanks

### DIFF
--- a/library/math/num.lisp
+++ b/library/math/num.lisp
@@ -191,7 +191,18 @@
 
        (define (- a b)
          (lisp ,type (a b)
-           (cl:values (cl:mod (cl:- a b) ,(cl:expt 2 bits)))))
+	   (cl:if (cl:< a b)
+		  (cl:cond
+		    ((cl:= ,bits 8)
+		     255)
+		    ((cl:= ,bits 16)
+		     65535)
+		    ((cl:= ,bits 32)
+		     4294967295)
+		    ((cl:= ,bits 64)
+		     18446744073709551615)
+		    (cl:t 4611686018427387903))
+		  (cl:values (cl:mod (cl:- a b) ,(cl:expt 2 bits))))))
 
        (define (* a b)
          (lisp ,type (a b)
@@ -199,7 +210,18 @@
 
        (define (fromInt x)
          (lisp ,type (x)
-           (cl:values (cl:mod x ,(cl:expt 2 bits))))))))
+	   (cl:if (cl:minusp x)
+		  (cl:cond
+		    ((cl:= ,bits 8)
+		     255)
+		    ((cl:= ,bits 16)
+		     65535)
+		    ((cl:= ,bits 32)
+		     4294967295)
+		    ((cl:= ,bits 64)
+		     18446744073709551615)
+		    (cl:t 4611686018427387903))
+		  (cl:values (cl:mod x ,(cl:expt 2 bits)))))))))
 
 (coalton-toplevel
   (define-num-checked Integer cl:identity)


Izaakwltn:
I tested it and this provides correct underflow errors.

I think it would make sense to use `cl:error` instead of `cl:cerror`, and remove the `continue-format-string` since there is no wrapping operation defined. Currently, if you select ` 0: [CONTINUE] Continue, wrapping around.` in the debugger, you get another error because it's still expecting a number and isn't getting one:
```
The value
  COMMON-LISP:NIL
is not of type
  (UNSIGNED-BYTE 16)
   [Condition of type COMMON-LISP:TYPE-ERROR]
```

Otherwise, this seems solid.

Izaakwltn:
Looks good to me! I have no gripes. 

I think there should be a future issue to have unsigned numbers error on overflow too, but that can be a separate thing.

I'm going to approve it but I'll wait to see if anyone else wants to review it before merging.

Izaakwltn (approved):

library/math/num.lisp
```diff
@@ -161,6 +161,35 @@
 ;;; Num instances for integers
 ;;;
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
```
stylewarning, line 164: Issues with tabs. Please convert to the appropriate number of spaces.

library/math/num.lisp
```diff
@@ -161,6 +161,35 @@
...
+  (cl:defmacro define-unsigned-num-underflow (type bits)
+    "Define a `Num' instance for Type which signals an error on underflow."
+    `(define-instance (Num ,type)
+       (define (+ a b)
+         (lisp ,type (a b)
+	   (cl:if (cl:and (cl:< b 0) (cl:< a (cl:- 0 b)))
+		  (cl:error ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
```
stylewarning, line 171: I think there should be a CERROR allowing wrapping behavior.

library/math/num.lisp
```diff
@@ -161,6 +161,35 @@
 ;;; Num instances for integers
 ;;;
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:defmacro define-unsigned-num-underflow (type bits)
+    "Define a `Num' instance for Type which signals an error on underflow."
+    `(define-instance (Num ,type)
+       (define (+ a b)
+         (lisp ,type (a b)
+	   (cl:if (cl:and (cl:< b 0) (cl:< a (cl:- 0 b)))
```
stylewarning, line 170: (< x 0) ==> (minusp x)

(- 0 x) ==> (- x)

library/math/num.lisp
```diff
@@ -161,6 +161,35 @@
...
+		  (cl:error ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  (cl:+ a b))))
+       
+       (define (- a b)
+         (lisp ,type (a b)
+	   (cl:if (cl:and (cl:>= b 0) (cl:< a (cl:+ 0 b)))
```
stylewarning, line 176: b is guaranteed >= 0 for unsigned numbers

library/math/num.lisp
stylewarning, line 170: b can never be negative for unsigned numbers

library/math/num.lisp
```diff
@@ -161,6 +161,35 @@
...
+		  (cl:error ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		  (cl:- a b))))
+       
+       (define (* a b)
+	 (lisp ,type (a b)
+	   (cl:if (cl:or (cl:and (cl:and (cl:> b 0) (cl:< a 0)) (cl:< a (cl:/ 0 b)))
```
stylewarning, line 182: unsigned multiplication can't underflow

stylewarning:
(Incidentally this means we have no tests that actually look for correct overflow / underflow behavior.)

Jobhdez:
@stylewarning thank you for this.

I will try to address your review as soon as I can. Thanks. 

For the wrapper, I just have to make it so when the unsigned values underflow, the user can continue, if they want, with the highest number possible of the given unsigned type right?

so if two addition of U8 integers underflow, the wrapper needs to allow the user continue from the error with the wrapped number which in this case would be 255? correct? ust trying to make sure I am thinking the right thing. thanks

library/math/num.lisp
Jobhdez: what do you mean by this? are you saying the code that i wrote doesnt detect the underflow or that in general unsigned multiplication doesnt underflow?

wouldnt this be an underflow: `2 * -3`?

Izaakwltn: -3 isn't an unsigned number

library/math/num.lisp
```diff
@@ -161,6 +161,50 @@
...
+       (define (+ a b)
+	 (lisp ,type (a b)
+	   (cl:cond ((cl:and (cl:minusp b) (cl:< a (cl:- b)))
+		     (cl:cerror "Continue, wrapping around." ,(cl:format cl:nil "Unsigned value underflowed ~D bits." bits))
+		     (,underflow-handler ,bits))
+		    (cl:+ a b))))
+	 
+	 (define (- a b)
+	   (lisp ,type (a b)
+	     (cl:cond ((cl:< a (cl:+ 0 b))
```
Izaakwltn, line 192: what is the point of (cl:+ 0 b)? doesn't it just evaluate to b?

Jobhdez: I can simplify this. Thanks

library/math/num.lisp
```diff
@@ -161,6 +161,50 @@
...
+		       (,underflow-handler ,bits))
+		      (cl:- a b))))
+	 
+	 (define (* a b)
+	   (lisp ,type (a b)
+	     (cl:* a b)))
+
+	 (define (fromInt x)
+	   (lisp ,type (x)
+	     (cl:cond ((cl:< x 0)
```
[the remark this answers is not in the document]
Jobhdez: Will make this change as well; thanks

Jobhdez:
@Izaakwltn addressed your review

library/math/num.lisp
```diff
@@ -161,6 +161,47 @@
 ;;; Num instances for integers
 ;;;
 
+(cl:defmacro %define-unsigned-underflow-handler (name bits)
```
stylewarning, line 164: i think this can be deleted

library/math/num.lisp
```diff
@@ -161,6 +161,47 @@
...
+  `(cl:defun ,name (bits)
+     (cl:declare (type (unsigned-byte ,bits) bits))
+     (cl:ecase ,bits
+       (8 (cl:the (cl:unsigned-byte ,bits) 255))
+       (16 (cl:the (cl:unsigned-byte ,bits) 65535))
+       (32 (cl:the (cl:unsigned-byte ,bits) 4294967295))
+       (64 (cl:the (cl:unsigned-byte ,bits) 18446744073709551615)))))
+
+
+(%define-unsigned-underflow-handler %unsigned-8-bit-underflow-handler 8)
```
stylewarning, line 174: these can be deleted

library/math/num.lisp
```diff
@@ -214,7 +255,12 @@
   (define-num-wrapping U16 16)
   (define-num-wrapping U32 32)
   (define-num-wrapping U64 64)
-  (define-num-wrapping UFix #.+unsigned-fixnum-bits+))
+  (define-num-wrapping UFix #.+unsigned-fixnum-bits+)
+  
+  (define-unsigned-num-underflow U8 %unsigned-8-bit-underflow-handler 8)
```
stylewarning, line 260: these are re-defining `Num` for types that already have it. Above, `define-num-wrapping` is already defining `-` and the like for each of the listed types.

stylewarning (changes requested):
Thanks for your patience in review.

This PR needs to modify `define-num-wrapping`'s definition of `-` and `fromInt` only. Adding another `define` macro is not needed (and is actually redundant), and the other "handler" functionality is not all that useful outside of the scope of this change, so it can be eliminated.

Jobhdez:
> Thanks for your patience in review.
> 
> This PR needs to modify `define-num-wrapping`'s definition of `-` and `fromInt` only. Adding another `define` macro is not needed (and is actually redundant), and the other "handler" functionality is not all that useful outside of the scope of this change, so it can be eliminated.

mybad for the late reply. But your comment means that I only pretty much have take what I wrote in my function and move to the `define-num-wrapping` function? thanks

stylewarning:
Yes, essentially.

Jobhdez:
> Yes, essentially.

Ok I’ll get it done ASAP.

Izaakwltn:
At least one test failing due to the `ecase`, specifically in test `iter-min-max`: 

```
While evaluating the form starting at line 3, column 0
  of #P"/__w/coalton/coalton/run-tests.lisp":
Unhandled SB-KERNEL:CASE-FAILURE in thread #<SB-THREAD:THREAD "main thread" RUNNING
                                              {1001090073}>:
  62 fell through COMMON-LISP:ECASE expression. Wanted one of (8 16 32 64).
``` 

Your `ecase` can get matches for bits that aren't 8, 16, 32, or 64. 

I would recommend loading the coalton test suite and either running all tests locally or just running `iter-min-max` to debug.

Izaakwltn:
I think `+unsigned-fixnum-bits+` is sbcl specific, so the tests are passing in SBCL but still not in the other implementations:

Allegro:
https://github.com/coalton-lang/coalton/actions/runs/8151586516/job/22279665406?pr=1025#step:5:507

CCL:
https://github.com/coalton-lang/coalton/actions/runs/8151586516/job/22279665807?pr=1025#step:6:434

You might be able to do an `#+sbcl` exception for that approach and find another way to handle the other implementations, but if you find a uniform solution for all three types that would be optimal.

Jobhdez:
@Izaakwltn @stylewarning hey guys, my apologies for the delay on this issue. Im going to try izaak's suggestions today or tomorrow and figure out the cause. i do plan to finish this.

stylewarning:
Due to the age of this PR, I'm going to be closing it. Please feel free to open another PR when you believe you've solved the issue. If the issue needs more clarification, please ask in the issue.

Issue #1021, opened 2023-10-17
`for` loop evaluation in a `coalton` macro body typechecks when `coalton-codegen` does not.


macrologist:
That is, this works:

``` lisp
(coalton 
 (let ((v (vector:new))) 
   (for x in "hello" 
        (vector:push! (fn () (traceobject "x" x)) v))
   (for x in v (x))))

```

But this fails to typecheck at codegen

```lisp
(coalton:coalton-codegen 
  (define (thunk)
    (let ((v (vector:new))) 
      (for x in "hello" 
           (vector:push! (fn () (traceobject "x" x)) v))
      (for x in v (x)))))
```

macrologist:
This wasn't always the case. When I refactored `for` in terms of `while-let`, taking advantage of `node-bind`, I failed to think `node-bind`'s behavior through.

Notably, `while-let` is unaffected.

PR #1026, opened 2023-10-19
Fix: codegen appends Unit expr onto loop bodies.

macrologist:
Addresses #1021

### DIFF
--- a/src/codegen/translate-expression.lisp
+++ b/src/codegen/translate-expression.lisp
@@ -560,7 +560,7 @@ Returns a `node'.")
      :type tc:*unit-type*
      :label (tc:node-while-label expr)
      :expr (translate-expression (tc:node-while-expr expr) ctx env)
-     :body (translate-expression (tc:node-while-body expr) ctx env)))
+     :body (append-unit (translate-expression (tc:node-while-body expr) ctx env))))
 
   (:method ((expr tc:node-while-let) ctx env)
     (declare (type pred-context ctx)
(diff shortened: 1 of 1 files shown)


stylewarning (approved):
I object to the name "force-unit" but I don't immediately have a better alternative.

macrologist:
> I object to the name "force-unit" but I don't immediately have a better alternative.

If @eliaslfox doesn't object to the approach, I can rename it something like 

`(append-unit-expr <node>)` or something

eliaslfox (approved):
I like `append-unit`.

eliaslfox:
@macrologist did you figure out why the behavior was different in `coalton-codegen`?

macrologist:
@eliaslfox 
> @macrologist did you figure out why the behavior was different in `coalton-codegen`?

Sort of. The specialization of `typecheck-node` on `node-bind` unifies the `node-bind`'s expression type with it's body's  type.  Looking at it again I see that typecheck-node returns a type, and that recursive invocations rely on that return type: I did not realize this earlier. 

I think that the loop nodes should all return unit from typecheck-node but aren't. They're returning the type of the body.

Perhaps this PR should be reverted in favor altering typecheck-node instead. Or perhaps both -- though both would be a little redundant. My preference would be to drop these commits and alter typecheck-node.

macrologist:
(a bit more verbiage) From the perspective of correctness, afaict, the choice is fairly arbitrary. From the perspective of performance, killing these commits in favor of altering typecheck-node seems better. But maybe I'm wrong about correctness?

eliaslfox:
> killing these commits in favor of altering typecheck-node seems better

This seems like a better solution.

PR #1029, opened 2023-10-20
Add INLINE to define inline functions and methods

digikar99:
This PR is a sketch to add inlinable functions and methods to coalton. List of tasks to complete before merge:

- [x] Basic function inlining
- [x] Basic method inlining
- [x] Parsing errors
- [x] Basic tests
- [x] Inline/Non-inline recursive and mutually recursive definitions: we don't want a stackoverflow 
- [ ] Confirm slot naming conventions: is `inline-p` okay, or should they be renamed to `inline`? Although, I hope it doesn't create confusion with `set-method-inline` and friends.
- [ ] Any suggestions about whether some code changes are good or bad?
- [ ] (Optional) Inline cases when the number of arguments differ from the number of parameters
- [ ] (Optional) Any related trivial features
- [ ] Ready for merge

---

The intended use is as follows:

```lisp
(coalton-toplevel
  (declare foo (single-float -> single-float))
  (inline) ; <--- THIS ATTRIBUTE
  (define (foo x)
    (lisp single-float (x) (cl:+ x x))))

(coalton-toplevel
  (define (foo-caller x)
    (foo x)))
```

Without `foo` being `inline`, the disassembly of `foo-caller` would have included a call to `foo` as follows:

```
COALTON-USER> (disassemble 'foo-caller)

; disassembly for FOO-CALLER
; Size: 53 bytes. Origin: #x540AAB0D                          ; FOO-CALLER
; 0D:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; 11:       488945F8         MOV [RBP-8], RAX
; 15:       4883EC10         SUB RSP, 16
; 19:       488BD6           MOV RDX, RSI
; 1C:       B902000000       MOV ECX, 2
; 21:       48892C24         MOV [RSP], RBP
; 25:       488BEC           MOV RBP, RSP
; 28:       E8D54F33FC       CALL #x503DFB02                  ; #<FDEFN FOO>
; 2D:       480F42E3         CMOVB RSP, RBX
; 31:       488B75F0         MOV RSI, [RBP-16]
; 35:       80FA19           CMP DL, 25
; 38:       7503             JNE L0
; 3A:       C9               LEAVE
; 3B:       F8               CLC
; 3C:       C3               RET
; 3D: L0:   CC55             INT3 85                          ; OBJECT-NOT-SINGLE-FLOAT-ERROR
; 3F:       08               BYTE #X08                        ; RDX(d)
; 40:       CC10             INT3 16                          ; Invalid argument count trap
```

But now that `foo` can be `inline`d, the disassembly of `foo-caller` is: 

```
COALTON-USER> (disassemble 'foo-caller)

; disassembly for FOO-CALLER
; Size: 31 bytes. Origin: #x540AAE7F                          ; FOO-CALLER
; 7F:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; 83:       488945F8         MOV [RBP-8], RAX
; 87:       0F28D1           MOVAPS XMM2, XMM1
; 8A:       F30F58D1         ADDSS XMM2, XMM1
; 8E:       660F7ED2         MOVD EDX, XMM2
; 92:       48C1E220         SHL RDX, 32
; 96:       80CA19           OR DL, 25
; 99:       C9               LEAVE
; 9A:       F8               CLC
; 9B:       C3               RET
; 9C:       CC10             INT3 16                          ; Invalid argument count trap
```

This not only extends to regular functions, but also to class method instances. 

```lisp
(coalton-toplevel
  (define-class (bar :a)
    (add (:a -> :a -> :a))))

(coalton-toplevel
    (define-instance (bar single-float)
    (inline) ; <--- THIS ATTRIBUTE
    (define (add x y)
      (lisp single-float (x) (cl:+ x x)))))

(coalton-toplevel
  (declare add-caller (single-float -> single-float -> single-float))
  (define (add-caller x y)
    (add x y)))
```

Effectively, this brings us inlinable static dispatch with zero runtime overhead for calling simple functions.

---

~~In an [earlier PR](https://github.com/coalton-lang/coalton/pull/987), it was suggested that a `coalton:inline` declaration may be added to define inline functions and methods. I felt a slight inclination towards a `coalton:define-inline` form instead of a `coalton:inline` declaration. Nonetheless, if a declaration seems better, I will attempt to modify the PR accordingly.~~ I am adding it as an attribute for the moment instead of any separate toplevel form. If someone has a preference for a toplevel form, they can always [write a macro](https://github.com/coalton-lang/coalton/pull/1029#issuecomment-1787749479). The case for converting it into a CL-like declaration for more [granular control](https://github.com/coalton-lang/coalton/pull/1029#issuecomment-1787749479) remains.

This is still a sketch since I haven't looked over every case carefully. I'd be also be delighted if this facility could be achieved by more minimal changes.

### DIFF
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -133,21 +133,28 @@
                   (node-seq-nodes node))))
 
 (defmethod propagate-constants ((node node-lisp) env &optional constant-bindings)
-  (let ((new-lisp-vars nil))
-    (make-node-let
-     :type (node-type node)
-     :bindings (loop :for (lisp-var . coalton-var) :in (node-lisp-vars node)
-                     :for constant-value := (constant-var-value coalton-var constant-bindings :no-error t)
-                     :if constant-value
-                       :collect (let ((new-coalton-var (gentemp (symbol-name coalton-var))))
-                                  (push (cons lisp-var new-coalton-var) new-lisp-vars)
-                                  (cons new-coalton-var constant-value))
-                     :else
-                       :do (push (cons lisp-var coalton-var) new-lisp-vars))
-     :subexpr (make-node-lisp
-               :type (node-type node)
-               :vars new-lisp-vars
-               :form (node-lisp-form node)))))
+  (let* ((new-lisp-vars nil)
+         (let-bindings
+           (loop :for (lisp-var . coalton-var) :in (node-lisp-vars node)
+                 :for constant-value := (constant-var-value coalton-var constant-bindings :no-error t)
+                 :if constant-value
+                   :collect (let ((new-coalton-var (gentemp (symbol-name coalton-var))))
+                              (push (cons lisp-var new-coalton-var) new-lisp-vars)
+                              (cons new-coalton-var constant-value))
+                 :else
+                   :do (push (cons lisp-var coalton-var) new-lisp-vars))))
+    (if let-bindings
+        (make-node-let
+         :type (node-type node)
+         :bindings let-bindings
+         :subexpr (make-node-lisp
+                   :type (node-type node)
+                   :vars new-lisp-vars
+                   :form (node-lisp-form node)))
+        (make-node-lisp
+         :type (node-type node)
+         :vars new-lisp-vars
+         :form (node-lisp-form node)))))
 
 (defmethod propagate-constants ((node node-match) env &optional constant-bindings)
   (make-node-match
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -1,0 +1,157 @@
+(defpackage #:coalton-impl/codegen/constant-propagation
+  (:use
+   #:cl
+   #:coalton-impl/codegen/ast)
+  (:local-nicknames
+   (#:tc #:coalton-impl/typechecker))
+  (:export
+   #:propagate-constants))
+
+(in-package #:coalton-impl/codegen/constant-propagation)
+
+(defun constant-node-p (env node constant-bindings)
+  (declare (type node node))
+  ;; FIXME: We need more nodes classified as constants
+  (cond ((node-literal-p node)
+         t)
+        ((and (node-variable-p node)
+              (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
+         t)
+        ((node-variable-p node)
+         (let* ((name (node-variable-value node))
+                (variable-value-node
+                  (or (cdr (assoc name  constant-bindings))
+                      (tc:lookup-function env name :no-error t))))
+           (when variable-value-node
+             (constant-node-p env variable-value-node constant-bindings))))
+        (t
+         nil)))
+
+(defun constant-var-value (var constant-bindings)
+  (cdr (assoc var constant-bindings)))
+
+(defun constant-node-value (env node constant-bindings)
+  (declare (type node node))
+  ;; FIXME: We need more nodes classified as constants
+  (cond ((node-literal-p node)
+         node)
+        ((and (node-variable-p node)
+              (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
+         node)
+        ((node-variable-p node)
+         (let ((variable-value-node
+                 (or (constant-var-value (node-variable-value node) constant-bindings)
+                     (let ((fn (tc:lookup-function env (node-variable-value node) :no-error t)))
+                       (if (and fn (tc:function-env-entry-inline-p fn))
+                           (tc:lookup-code env (tc:function-env-entry-name fn) :no-error t)
+                           nil)))))
+           (if variable-value-node
+               (constant-node-value env variable-value-node constant-bindings)
+               nil)))
+        (t
+         nil)))
+
+(defun propagate-constants-with-bindings (env node constant-bindings)
+  "Returns two values: the first is the propagted value, the second indicates success if non-NIl"
+  ;; (terpri)
+  ;; (print (list node bindings))
+  (cond
+    ((node-literal-p node)
+     node)
+    ((node-variable-p node)
+     ;; FIXME: Can or Should we use an existing structure to handle bindings
+     (or (constant-node-value env node constant-bindings)
+         node))
+    ((node-let-p node)
+     (let ((node-bindings (node-let-bindings node))
+           (new-constant-bindings nil)
+           (nonconstant-bindings nil))
+       (loop :for (var . value) :in node-bindings
+             :for propagated-value-node := (propagate-constants-with-bindings env value constant-bindings)
+             :do (if (constant-node-p env propagated-value-node constant-bindings)
+                     (push (cons var propagated-value-node)
+                           new-constant-bindings)
+                     (push (cons var propagated-value-node)
+                           nonconstant-bindings)))
+       (let ((constant-bindings (append new-constant-bindings constant-bindings)))
+         (if nonconstant-bindings
+             (make-node-let
+              :type (node-type node)
+              :bindings nonconstant-bindings
+              :subexpr (propagate-constants-with-bindings
+                        env
+                        (node-let-subexpr node)
+                        constant-bindings))
+             (propagate-constants-with-bindings
+              env
+              (node-let-subexpr node)
+              constant-bindings)))))
+    ((node-abstraction-p node)
+     (make-node-abstraction
+      ;; FIXME: A more specific type?
+      :type (node-type node)
+      :vars (node-abstraction-vars node)
+      :subexpr (propagate-constants-with-bindings
+                env
+                (node-abstraction-subexpr node)
+                constant-bindings)))
+    ((node-application-p node)
+     (let ((constant-propagated-rands
+             (mapcar (lambda (rand)
+                       (propagate-constants-with-bindings
+                        env
+                        rand
+                        constant-bindings))
+                     (node-application-rands node))))
+       (make-node-application
+        ;; FIXME: A more specific type
+        :type (node-type node)
+        :rator (node-application-rator node)
+        :rands constant-propagated-rands)))
+    ((node-direct-application-p node)
+     (let ((constant-propagated-rands
+             (mapcar (lambda (rand)
+                       (propagate-constants-with-bindings
+                        env
+                        rand
+                        constant-bindings))
+                     (node-direct-application-rands node))))
+       (make-node-direct-application
+        :type (node-type node)
+        :rator-type (tc:make-function-type*
+                     (mapcar #'node-type constant-propagated-rands)
+                     (tc:function-return-type (node-type node)))
+        :rator (node-direct-application-rator node)
+        :rands constant-propagated-rands)))
+    ((node-seq-p node)
+     (make-node-seq
+      :type (node-type node)
+      :nodes (mapcar (lambda (node)
+                       (propagate-constants-with-bindings env node constant-bindings))
+                     (node-seq-nodes node))))
+    ((node-lisp-p node)
+     ;; FIXME: This looks too convoluted.
+     (let ((new-lisp-vars nil))
+       (make-node-let
+        :type (node-type node)
+        :bindings (loop :for (lisp-var . coalton-var) :in (node-lisp-vars node)
+                        :for constant-value := (constant-var-value coalton-var constant-bindings)
+                        :if constant-value
+                          :collect (let ((new-coalton-var (gentemp (symbol-name coalton-var))))
+                                     (push (cons lisp-var new-coalton-var) new-lisp-vars)
+                                     (cons new-coalton-var constant-value))
+                        :else
+                          :do (push (cons lisp-var coalton-var) new-lisp-vars))
+        :subexpr (make-node-lisp
+                  :type (node-type node)
+                  :vars new-lisp-vars
+                  :form (node-lisp-form node)))))
+    (t
+     ;; TODO: Propagate constants in other nodes
+     (error "Don't know how to propagate constants in ~S" node))))
+
+(defun propagate-constants (node env)
+  (declare (type node node))
+  ;; FIXME: Can we use TRAVERSE?
+  ;; The trouble is: TRAVERSE expects a symbol-list, while we want a binding list
+  (propagate-constants-with-bindings env node ()))
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -9,8 +9,11 @@
 
 (in-package #:coalton-impl/codegen/constant-propagation)
 
-(defun constant-var-value (var constant-bindings)
-  (cdr (assoc var constant-bindings)))
+(defun constant-var-value (var constant-bindings &key no-error)
+  (the (or node null)
+       (or (cdr (assoc var constant-bindings))
+           (unless no-error
+             (error "Expected VAR ~S to be a constant but is not" var)))))
 
 (defun constant-node-p (env node constant-bindings)
   (declare (type node node))
@@ -21,11 +24,11 @@
               (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
          t)
         ((node-variable-p node)
-         (constant-var-value (node-variable-value node) constant-bindings))
+         (constant-var-value (node-variable-value node) constant-bindings :no-error t))
         (t
          nil)))
 
-(defun constant-node-value (env node constant-bindings)
+(defun constant-node-value (env node constant-bindings &key no-error)
   (declare (type node node))
   ;; FIXME: We need more nodes classified as constants
   (cond ((node-literal-p node)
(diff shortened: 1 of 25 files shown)


stylewarning:
I haven't reviewed but I just wanted to leave a few pre-comments.

1. Amazing job diving in and making such a broad change. That's a lot of work, and I definitely want to see this PR go through.

2. I suspect we will want a few declarations. One thing we want to do that CL doesn't do is heuristic inlining. I don't know what design we'd choose but things like small lambdas that don't escape or unexported functions might be heuristically inlined by default. (A fully fledged heuristic inliner isn't necessary for this PR, but probably the groundwork for making it happen is.)

Excited to see this. Inlining will be such a huge performance lever in Coalton that could transform it from being "inadequate" to being "faster than Lisp by default".

digikar99:
Thank you for the pre-review! I'm still thinking what the right thing to do would be. But yes, I definitely want to see inlining in coalton. 

One main concern I have with this approach is that it depends on the implementation to perform the inlining - see the following changes in src/codegen/program.lisp :

```lisp
         :if (node-abstraction-p node)
           :append (list
                    (if (node-abstraction-inline-p node)
                        `(declaim (inline ,name))
                        `(declaim (notinline ,name)))
                    (compile-function name node env)
                    `(setf
                      ,name
```

This means that suppose `coalton:inline` and `coalton:notinline` were made similar to `cl:inline` and `cl:notinline` like declarations. In that case, granular inlining as in `coalton::(let (...) (declare (not/inline ...)) ...)` will need to emit `(cl:declare (cl:not/inline ...))` declarations. I want to hear others' thoughts on this. 

As I understand, avoiding this would require some more work, so that inlining can be performed within coalton compilation without depending on the implementation.

stylewarning:
We should definitely attempt to inline ourselves (at the AST level) so that further type and monomorphizer/specializer optimizations. Maybe in specific last-mile optimizations we would use the Lisp inliner. (I'm happy Lisp has inlining, but it's a very sharp and specific tool.)

eliaslfox:
Hey, thanks for working on this.

Wether a function is marked inline should be tracked in the [`function-environment`](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/environment.lisp#L623-L625) instead storing a flag on each ast node. The function environment is set for functions [here](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/define.lisp#L230-L236). It looks like this isn't being set for method definitions, but it can be set [in this loop](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/define-instance.lisp#L159).


I would prefer to use attributes instead of adding a new toplevel form:

```
;; toplevel functions
(inline)
(define (f x) ...)

;; methods
(define-instance (C T)
  (inline)
  (define (m x) ...))
```

The inliner should be written as a separate optimization pass instead of being merged with the `inline-method` pass. The purpose of the method inlining pass is to directly call methods when the type class instance being used is known at the call site.

Lastly, the inliner should generate fresh variable names instead of reusing the names of the functions's parameter names. This way we avoid issues with overlapping variable name. You can generate variable names with `gentemp`, and rewrite the variable references in an ast node with [ast substutions](https://github.com/coalton-lang/coalton/blob/main/src/codegen/ast-subsitutions.lisp).

digikar99:
Thanks for the extensive feedback! I will rework this over the next week or two and push the commits for further review.

Yes, now, I'm in favour of attributes over a new toplevel form. If a user wants, they can write a macro which expands into the attribute+form:

```lisp
(defmacro define-inline (name &body body)
  `(progn
     (inline)
     (define ,name ,@body)))
```

However, in the longer run, if we wish to provide more granular SBCL-like control over inlining, we might need something akin to declarations.

```lisp
(declaim (inline foo))
(defun foo (a)
  (+ a a))
(declaim (notinline foo))

(defun foo-inline-caller (b)
  (declare (inline foo))
  (foo b))

(defun foo-mixed-caller (b)
  (declare (inline foo))
  (+ (foo b)
     (locally (declare (notinline foo)) ; <------ Are coalton attributes suitable for this?
       (foo b))))
```

digikar99:
> Wether a function is marked inline should be tracked in the [function-environment](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/environment.lisp#L623-L625) instead storing a flag on each ast node. The function environment is set for functions [here](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/define.lisp#L230-L236). It looks like this isn't being set for method definitions, but it can be set [in this loop](https://github.com/coalton-lang/coalton/blob/main/src/typechecker/define-instance.lisp#L159).

It looks like creating a `function-inline-environment` might be a better idea, both with respect to minimal code changes as well as the granular control over inlining as outlined above.

Without that, it seems we will need to change `function-env-entry` structure to include an `inline-p` slot in addition to the existing `name` and `arity`. Doing this requires changing all the places where the `function-env-entry` objects are being created to set the `:inline-p` slot. One such example is this [update-function-env function](https://github.com/coalton-lang/coalton/blob/main/src/codegen/transformations.lisp#L234-L236) in codegen/transformations.lisp. However, the value of `inline-p` itself requires more information. This information could either be obtained by adding a parameter `inline-bindings` to `update-function-env` itself, but this seems a bit unclean. Another way is to obtain this information from [split-binding-definitions](https://github.com/coalton-lang/coalton/blob/main/src/codegen/transformations.lisp#L213-L220), but this requires information about inlining to be present in the AST itself, which I understand you prefer to be avoided. 

Let me know if I should proceed by adding a `function-inline-environment` or if some other way is preferable.

eliaslfox:
In every place `set-function` is called except in `update-function-env` whether the function should be inline is known at the call site. `update-function-env` can instead use a different env updater to set the arity without changing inline-p:

```lisp
(define-env-updater set-function-arity (env symbol arity)
  (declare (type environment env)
           (type symbol symbol)
           (type fixnum arity))
  (let ((prev (lookup-function env symbol :no-error t)))
    (update-environment
     env
     :function-environment (immutable-map-set
                            (environment-function-environment env)
                            symbol
                            (make-function-env-entry
                             :name symbol
                             :arity arity
                             :inline-p (and prev (function-env-entry-inline-p prev)))
                            #'make-function-environment))))
```

digikar99:
Sorry, I've been busy. It's been great to see lots of activity on coalton recently!

I might be free to dive into this in detail in another month. Is there any suggestion to break the changes required for this into multiple parts? That way, each part will be easier to review and merge.

src/codegen/optimizer.lisp
```diff
@@ -292,6 +299,113 @@
...
+                         (t
+                          (make-node-application
+                           :type (node-type node)
+                           :rator (make-node-variable
+                                   :type (tc:make-function-type*
+                                          (mapcar #'node-type rands)
+                                          (node-type node))
+                                   :value name)
+                           :rands rands)))))))
+           ;; FIXME: What kind of frontend code evokes DIRECT-APPLICATION
```
eliaslfox, line 355: [This pass](https://github.com/coalton-lang/coalton/blob/main/src/codegen/optimizer.lisp#L107) turns regular calls into direct calls when it's known that regular function calls can be used.

src/codegen/optimizer.lisp
```diff
@@ -292,6 +299,113 @@
...
+           ;; as opposed to APPLICATION?
+           (inline-direct-application (node &rest rest)
+             (declare (ignore rest)
+                      (dynamic-extent rest))
+             (let ((rator (node-application-rator node))
+                   (rands (node-application-rands node)))
+               (when (node-variable-p rator)
+                 (let* ((name (node-variable-value rator))
+                        (code (tc:lookup-code env name :no-error t))
+                        ;; FIXME: We need to lookup inline-p in the environment rather than the AST
```
eliaslfox, line 365: I think this FIXME is done

src/parser/toplevel.lisp
```diff
@@ -1981,6 +2023,17 @@ consume all attributes")))
...
    :source (cst:source form)))
 
+(defun parse-inline (form file)
+  (declare (type cst:cst form)
+           (type coalton-file file)
+           (values attribute-inline))
+
+  (assert (cst:consp form))
+
+  ;; TODO: Parsing errors
```
eliaslfox, line 2033: TODO

src/typechecker/define-class.lisp
```diff
@@ -358,7 +360,9 @@
                      :if (not (zerop method-arity))
                        :do (setf env (tc:set-function env method-name (tc:make-function-env-entry
                                                                        :name method-name
-                                                                       :arity method-arity)))
+                                                                       :arity method-arity
+                                                                       ;; FIXME: Always declare methods to be inline?
```
eliaslfox, line 364: I don't think we should do this

src/typechecker/define-class.lisp
```diff
@@ -334,7 +334,9 @@
            :if (not (zerop class-arity))
              :do (setf env (tc:set-function env codegen-sym (tc:make-function-env-entry
                                                              :name codegen-sym
-                                                             :arity class-arity)))
+                                                             :arity class-arity
+                                                             ;; FIXME: Under what circumstances would we want fundep to be inline?
```
eliaslfox, line 338: Class constructor functions can't be inlined by the Coalton inliner because they don't appear in the code env.

src/typechecker/define-type.lisp
```diff
@@ -532,7 +529,9 @@
                                    :vars nil
                                    :var-names nil
                                    :body (list (util:runtime-quote (type-definition-runtime-type type)))))
-                :source source))
+                :source source
+                ;; FIXME: Is NIL correct?
```
eliaslfox, line 533: yes

src/typechecker/expression.lisp
```diff
@@ -233,8 +234,9 @@
 (defstruct (node-abstraction
             (:include node)
             (:copier nil))
-  (params  (util:required 'params) :type pattern-list :read-only t)
-  (body    (util:required 'body)   :type node-body    :read-only t))
+  (params   (util:required 'vars)     :type pattern-list :read-only t)
+  (body     (util:required 'body)     :type node-body    :read-only t)
+  (inline-p (util:required 'inline-p) :type boolean      :read-only t))
```
eliaslfox, line 239: I think this is unnecessary because it's tracked in the function environment.

eliaslfox (commented):
Hey! Don't worry about breaking this up, it's pretty reviewable.

Sorry I dropped the ball on reviewing this.

This needs some tests, but other than that and a couple nits this looks good to merge.

kartik-s:
Hi, great PR! I wanted to jump in with some thoughts about laying the groundwork to implement heuristic inlining as mentioned by @stylewarning.

In addition to this `(INLINE)` attribute, which tells the optimizer to inline all applications of the given function/method globally, what about having an attribute that behaves similar to the `(MONOMORPHIZE)` attribute, i.e. it tells the optimizer to:

1. Traverse the call graph rooted by the annotated function/method, collecting candidate applications.
2. For each candidate application, check if the application matches some heuristic function of the operator, operands, and potentially the context of the application (e.g. the function/method containing the application). If it does, inline it.
3. Iterate to fixed point, or until some stopping condition is detected by the (stateful) heuristic function.

This could be implemented by mostly copying the existing code for the monomorphizer, tweaking `CANDIDATE-SELECTION`   and `REWRITE-CALLSITES` appropriately (e.g. `REWRITE-CALLSITES` would contain the AST splicing/substitution implemented in this PR). We could also factor out this "select + rewrite candidate applications" and then (re)implement the monomorphizer and this inliner using the common logic.

stylewarning:
@kartik-s I think these are great ideas, but I think for this PR, we should rename `inline` to `always-inline`, and then come up with a smarter inlining plan (in the spirit of your comment) as a separate consideration.

@digikar99 Would it be possible to get this MR rebased, if it's not too much trouble?

digikar99:
Sorry, I missed the last review by eliaslfox! I will be putting this on priority for the upcoming week or two. 

I think heuristic inlining and a monomorphize-like inline should rather be in a separate PR. Even implementing basic inline-ing seems to require a fair amount of changes.

The currently missing "feature" for basic inlining is method inlining, that will allow inlining the `add` method of the `single-float` instance below:

```lisp
(coalton-toplevel
  (define-class (bar :a)
    (add (:a -> :a -> :a))))

(coalton-toplevel
    (define-instance (bar single-float)
    (inline) ; <--- THIS ATTRIBUTE
    (define (add x y)
      (lisp single-float (x) (cl:+ x x)))))

(coalton-toplevel
  (declare add-caller (single-float -> single-float -> single-float))
  (define (add-caller x)
    (add x)))
```

Earlier I had mistakenly made changes in src/typechecker/define-class.lisp to achieve the same. But it seems the appropriate place for changes might be src/typechecker/define-instance.lisp. May be no changes are required to define-instance.lisp either, and instead, setting the INLINE-P-TABLE in ENTRY-POINT would do the job.

---

I particularly need a review or suggestion on the ENTRY-POINT in src/entry.lisp. I've added an INLINE-P-TABLE which is populated from the contents of the parsed program. The INLINE-P slot in FUNCTION-ENV-ENTRY would be directly filled from INLINE-P-TABLE. This approach works for normal functions defined using toplevel DEFINE.

However, this approach runs into an issue while dealing with instance-methods. The inline attribute is present in the parsed object, but the parsed object does not contain the codegen-sym that should be marked as inline in INLINE-P-TABLE. The TY-INSTANCES and TOPLEVEL-INSTANCES contain codegen-sym, but they do not contain information about INLINE-P - or atleast my effort would be to avoid polluting them with the attributes. My current issue is: what is a good way to go from method objects of (parser:program-instances program) to the codegen-syms corresponding to those methods?

Or, is creating and filling the INLINE-P-TABLE in ENTRY-POINT a bad idea?

digikar99:
Basic method inlining is now in place too! That leaves

1. Error handling, 
2. Slot naming conventions: is `inline-p` okay, or should they be renamed to `inline`? Although, I hope it doesn't create confusion with `set-method-inline` and friends.
3. Any other suggestions about some code change is a good or bad change.
4. Further optimization: do we want to inline the case when the number of arguments differ from the number of parameters?
5. Related trivial features (eg: inline attribute inside `define-class`)
6. Anything else I might be overlooking.

Once those are in place, we can move the focus to heuristic inlining and monomorphize-like "inline everything", may be in a different PR.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,182 @@
...
+   #:coalton-impl/codegen/ast)
+  (:local-nicknames
+   (#:tc #:coalton-impl/typechecker))
+  (:export
+   #:propagate-constants))
+
+(in-package #:coalton-impl/codegen/constant-propagation)
+
+(defun constant-var-value (var constant-bindings)
+  (cdr (assoc var constant-bindings)))
```
amorphedstar, line 13: when checking that value exists later you need the whole cell, not just its `cdr`, which might still exist but be nil.
you can do this similar to `gethash`:
```suggestion
  (let ((x (assoc var constant-bindings)))
    (if (null x)
        (values nil nil) ; not found
        (values x t))))
```
(and then change usage of this function accordingly)

stylewarning: if only there were a type system (:

digikar99:
I think this is ready for further review now - I need opinions and suggestions to resolve the FIXMEs and the remaining tasks mentioned in the [top comment](https://github.com/coalton-lang/coalton/pull/1029#issue-1954366828).

amorphedstar:
The main inlining pass is now merged in [https://github.com/coalton-lang/coalton/pull/1193](https://github.com/coalton-lang/coalton/pull/1193), so for this PR, now it suffices to just keep the changes for handling syntax `(inline)` with the table and `inline-p` stuff. In `heuristic-inline-applications` in `optimizer.lisp`, we will just need the `inline-p` lookup and then `(funcall heuristic code)` can be replaced with `(or inline-p (funcall heuristic code))`. (Also, the current `inline-applications` will need to be updated to call `heuristic-inline-applications` with heuristic `(constantly nil)` when `settings:*coalton-heuristic-inlining*` is `nil`, instead of just skipping inlining entirely.)

Also, with the new `action`s in `traverse.lisp`, you should be able to update constant propagation to use `traverse`, passing the list the same way args are passed to `*traverse*` in `heuristic-inline-applications` and `make-binding-list-actions`.

digikar99:
Oops, I again missed an update here for quite a while! I have decreased my repository watch-level to a more appropriate level now!

Given the state of this PR and the updates that have taken place in coalton to enable inlining, I think I will issue two separate PRs for

1. constant-propagation
2. inlining syntax and connecting syntax with the inline and constant-propagation pass

stylewarning:
> Oops, I again missed an update here for quite a while! I have decreased my repository watch-level to a more appropriate level now!
> 
> 
> 
> Given the state of this PR and the updates that have taken place in coalton to enable inlining, I think I will issue two separate PRs for
> 
> 
> 
> 1. constant-propagation
> 
> 2. inlining syntax and connecting syntax with the inline and constant-propagation pass

Great ideas!

stylewarning:
Closed in favor of the series of PRs that followed.

PR #1033, opened 2023-12-04
integral.lisp uses nil instead of cl:nil

herbertjones:
macro if taking nil as not cl:nil and including signed checks in unsigned types

Example:

```lisp
(%define-integral-native U8 nil)
```
expands to
```lisp
(specialize isqrt u8-isqrt (u8 -> u8))
  (declare u8-isqrt (u8 -> u8))
  (define (u8-isqrt a)
    (if (< a 0) (error "Can't take ISQRT of a negative number.")
        (lisp u8
            (a)
          (common-lisp:isqrt a))))
```

and 

```lisp
(%define-integral-native U8 cl:nil)
```
expands to
```lisp
(specialize isqrt u8-isqrt (u8 -> u8))
  (declare u8-isqrt (u8 -> u8))
  (define (u8-isqrt a)
    (lisp u8
        (a)
      (common-lisp:isqrt a)))
```


eliaslfox (approved):
Thanks for the fix

PR #1034, opened 2023-12-09
Fix get-line-from-index function to handle UTF-8 characters

tojoqk:
When Coalton displays errors or warnings, there's a potential for an "end of file" error to occur if the file contains multi-byte characters.

The end-of-file error is caused by return incorrect line counts from `get-line-from-index`. Each character was counted as one byte, causing inaccuracies when processing multi-byte characters.

To resolve this issue, I have implemented a fix that accurately calculates line numbers by considering UTF-8 encoding.

For instance: 
```
(defpackage #:greeting
  (:use #:coalton
        #:coalton-prelude))

(in-package #:greeting)

(coalton-toplevel
  (define-type Greeting
    Hello
    Thanks
    Bye)

  (define (greet-ja g)
    (match g
      ((Hello) "こんにちは")
      ((Thanks) "ありがとう"))))
```

Error message:
```
1 compiler notes:

greeting.lisp:1:1:
  read-error: 
    
      end of file on #<SB-SYS:FD-STREAM for "file /tmp/greeting.lisp" {100874F373}>
    
      (in form starting at line: 9, column: 0, position: 144)

Compilation failed.
```

### DIFF
--- a/src/error.lisp
+++ b/src/error.lisp
@@ -513,11 +513,19 @@ Returns (VALUES LINE-NUM LINE-START-INDEX)"
   (file-position file 0)
   (let ((line 1)
         (line-start-index 0))
-    (loop :for i :to index
-          :for char := (read-char file)
+    (loop :for char := (read-char file)
+          :for code := (char-code char)
+          ;; It is assumed that code is utf-8 code point.
+          :for length := (cond ((<= code #x7f) 1)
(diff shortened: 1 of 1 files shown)


stylewarning (approved):

stylewarning:
Awesome!

PR #1038, opened 2024-01-04
Add Queue data structure to library

colescott:


library/queue.lisp
```diff
@@ -0,0 +1,258 @@
...
+    "Peek at the first item of `q`."
+    (if (empty? q)
+        None
+        (Some (peek-unsafe q))))
+
+  (declare peek-unsafe (Queue :a -> :a))
+  (define (peek-unsafe q)
+    "Peek at the first item of `q` without checking if the queue is empty."
+    (lisp :a (q)
+      (cl:car (queue-internal-elements q))))
```
stylewarning, line 156: this could be a type error, which i guess is par for the course for UNSAFE

stylewarning (approved):
seems ok to me

library/queue.lisp
```diff
@@ -0,0 +1,258 @@
...
+   #:items!))
+
+(in-package #:coalton-library/queue)
+
+(named-readtables:in-readtable coalton:coalton)
+
+#+coalton-release
+(cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
+
+(cl:declaim (cl:inline make-queue-internal))
```
eliaslfox, line 35: constructors are inline by default

library/queue.lisp
```diff
@@ -0,0 +1,258 @@
...
+    "Remove and return the first item of `q`."
+    (if (empty? q)
+        None
+        (Some (pop-unsafe! q))))
+
+  (declare pop-unsafe! (Queue :a -> :a))
+  (define (pop-unsafe! q)
+    "Remove and return the first item of `q` without checking if the queue is empty."
+    (lisp :a (q)
+      (cl:decf (queue-internal-length q))
```
eliaslfox, line 142: Reverse the order of these operations

eliaslfox (approved):

library/queue.lisp
colescott: It seems better for it to error rather than return invalid data. Will fix

eliaslfox:
Add this to the doc generator and add tests

PR #1031, opened 2023-11-17
Added type declarations to `computable-reals` functions

Izaakwltn:
This adds type declarations to non-instance `computable-reals` functions.

PR #1041, opened 2024-01-09
Computable-reals type ambiguity fix

Izaakwltn:
I added type declarations to avoid having to specify creal types every time.

### DIFF
--- a/library/computable-reals/computable-reals.lisp
+++ b/library/computable-reals/computable-reals.lisp
@@ -33,7 +33,7 @@
 
 (coalton-toplevel
   
- (repr :native cr:creal)
+ (repr :native (cl:or cr:creal cl:rational))
  (define-type Creal)
 
  (define (set-comparison-threshold! k)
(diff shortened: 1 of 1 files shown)


Izaakwltn:
This is actually a better version of PR #1031, which I didn't realize was there. I'll delete 1031.

library/computable-reals/computable-reals.lisp
```diff
@@ -33,23 +33,23 @@
 
 (coalton-toplevel
   
- (repr :native cr:creal)
- (define-type Creal)
+  (repr :native (cl:or cr:creal cl:rational))
```
macrologist, line 36: Hey It's a nit but can you remove these leading spaces? They appear in a few spots throughout this commit.

macrologist (commented):
I think I may have done something just like this in code I was writing a while back that used Creals.

macrologist (approved):

PR #1052, opened 2024-01-29
Add Practical Common Lisp port to examples

Jason94:
This PR ports some of the "practical" chapters from _Practical Common Lisp_ into Coalton. It includes the first two, and I'm working on chapter 23 (the spam filter) which could be a follow-up PR. I think that is a great source to show how to write practical, application style code in Coalton.

To that end, two goals of the style employed are:

- Highlight usage that is likely to show up in writing real applications, such as interop with CL and handling mutable state.
- Avoid some of the more complex areas of the type system, like type classes, monads, and _do_ notation.

I've written some fun side-code in Coalton before, but never a full program, so there might be places where the code could be improved or made more "Coaltony". I had a few places with particular questions for experienced Coalton writers, so I called them out in some comments.

### DIFF
--- a/examples/practical-coalton/15_pathnames.lisp
+++ b/examples/practical-coalton/15_pathnames.lisp
@@ -4,7 +4,8 @@
    #:coalton
    #:coalton-prelude)
   (:local-nicknames
-    (#:pth #:com.gigamonkeys.pathnames))
+    (#:pth #:com.gigamonkeys.pathnames)
+    (#:str #:coalton-library/string))
   (:export
    :list-directory
    :file-exists-p
@@ -14,7 +15,9 @@
    :pathname-as-file
    :walk-directory
    :directory-p
-   :file-p))
+   :file-p
+   :IncludeDirs
+   :ExcludeDirs))
 (in-package :practical-coalton.pathnames)
 
 (named-readtables:in-readtable coalton:coalton)
@@ -28,19 +31,9 @@
 ;;
 ;; First, we start by defining a few Coalton types that provide
 ;; a type-safe interface to the Common Lisp pathname functions.
-;; For example, in several of the Common Lisp functions a wildcard
-;; pathname is an invalid argument. Instead of checking for wildcard
-;; paths and erroring if found, we can use the type system to ensure
-;; that we never pass a wildcard path to a function that doesn't
-;; accept it.
 ;;
 ;; Second, we define a corresponding Coalton function for each
 ;; function exported by the Common Lisp pathname package.
-
-;; TODO: I'm open to any suggestions to make this code less verbose.
-;; If you count the comments at the top of the file and the package
-;; definition (which the original doesn't have), this is almost
-;; as long as the Common Lisp implementation this is wrapping!
 
 (cl:defmacro wrap-lisp-call (type function value)
   "Wrap a call to a Lisp function that returns a value of type TYPE
@@ -54,10 +47,6 @@ A nil return value is converted to None."
 
 (coalton-toplevel
 
-  (define-class (Path :a))
-
-  (define-instance (Path String))
-
   (repr :native cl:pathname)
   (define-type Pathname)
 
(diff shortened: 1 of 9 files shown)


examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+;; For example, in several of the Common Lisp functions a wildcard
+;; pathname is an invalid argument. Instead of checking for wildcard
+;; paths and erroring if found, we can use the type system to ensure
+;; that we never pass a wildcard path to a function that doesn't
+;; accept it.
+;;
+;; Second, we define a corresponding Coalton function for each
+;; function exported by the Common Lisp pathname package.
+
+;; TODO: I'm open to any suggestions to make this code less verbose.
```
Jason94, line 48: I'm open to any suggestions to make this code less verbose.

If you count the comments at the top of the file and the package definition (which the original doesn't have), this is almost as long as the Common Lisp implementation this is wrapping!

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+  (define-instance (Path Pathname))
+
+  (repr :native cl:pathname)
+  (define-type WildPathname)
+
+  (define-instance (Path WildPathname)))
+
+(coalton-toplevel
+
+  ;; TODO: It's a little unforunate that the Either type is called
```
Jason94, line 86: It's a little unfortunate for this use-case that the Either type is called Result in Coalton, because it makes a usage to represent one of two possible values a little confusing. 

Is there a more idiomatic way to represent this?

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+
+(coalton-toplevel
+  
+  ;; Create an enum type to make the :directories argument to walk-directory
+  ;; more discoverable from the type signature.
+  (define-type WalkDirectoryOptions
+    IncludeDirs
+    ExcludeDirs)
+  
+  ;; TODO: Is there a better way to be able to check if two WalkDirectoryOptions
```
Jason94, line 154: Is there a better way to check for equality between elements of an "enum-like" type such as `WalkDirectoryOptions`?

Jason94:
> Also, in response to your intro, I think you can pretty easily avoid monads and `do` notation, but I don't think you should shy away from type classes, as they're the most idiomatic way of writing coalton.

Yeah, that's a good point. I even used a typeclass in the ch 15 example, so I don't know why I thought that as I was writing the intro!

stylewarning:
Awesome! I'll be sure to check this out.

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+A nil return value is converted to None."
+  `(lisp (Optional ,type) (,value)
+     (cl:let ((result (,function ,value)))
+       (cl:if result
+         (coalton (Some (lisp ,type () result)))
+         (coalton None)))))
+
+(coalton-toplevel
+
+  (define-class (Path :a))
```
stylewarning, line 65: I would define a new type for paths:

```
(repr :native cl:pathname)
(define-type Pathname)

(define-type Path
  (PathString String)
  (LispPathname Pathname)
  ...)
```

I don't know if a type classes is needed. (That's only needed if we anticipate needing to define new kinds of paths in library code.)

examples/practical-coalton/15_pathnames.lisp
stylewarning: In this case, with a `Path` type, you get this "for free" without Either/Result/etc.

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+  
+  ;; Create an enum type to make the :directories argument to walk-directory
+  ;; more discoverable from the type signature.
+  (define-type WalkDirectoryOptions
+    IncludeDirs
+    ExcludeDirs)
+  
+  ;; TODO: Is there a better way to be able to check if two WalkDirectoryOptions
+  ;; are equal? This seems inefficient at runtime and verbose.
+  (define-instance (Eq WalkDirectoryOptions)
```
stylewarning, line 156: Your way is currently the simplest way without resorting to a hack.

The hack would be to define `WalkDirectoryOptions` as an `(repr :enum)` (guarantees the members to be symbols), and define

```
(define (== x y) (lisp Boolean (x y) (cl:eq x y)))
```

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,177 @@
...
+  (define (file-p pth)
+    "Return a pathname for the file named by PTH if it is an existing file.
+Otherwise return None."
+    (wrap-lisp-call :a pth:file-p pth)))
+
+(coalton-toplevel
+  
+  ;; Create an enum type to make the :directories argument to walk-directory
+  ;; more discoverable from the type signature.
+  (define-type WalkDirectoryOptions
```
stylewarning, line 150: See note below about `(repr :enum)`

examples/practical-coalton/1_simple_database.lisp
```diff
@@ -0,0 +1,250 @@
...
+    (Database (cel:new Nil)))
+
+  (declare add-record (Database -> CD -> Database))
+  (define (add-record db cd)
+    (cel:push! (cd-data db) cd)
+    db)
+
+  (declare dump-cd (CD -> Unit))
+  (define (dump-cd cd)
+    (traceobject "title" (.title cd))
```
stylewarning, line 74: We should really add some default printing/IO functions....

examples/practical-coalton/1_simple_database.lisp
```diff
@@ -0,0 +1,250 @@
...
+This is useful for converting keywords to Coalton struct accessors in the
+query DSL, like:
+
+(where :rating 8), in which :rating will eventually be transformed into
+(.rating cd) in the generated Coalton code.
+
+Example:
+
+(keyword-to-struct-accessor :title) => .TITLE"
+  (cl:intern (cl:concatenate 'cl:string (cl:string '#:.) (cl:string keyword))))
```
stylewarning, line 102: (concatenate 'string "." (string keyword)) works

stylewarning (changes requested):
Please make a .asd file that captures the dependencies and the like. It looks like some systems are missing (e.g., `quickload "Chapter-15"`)

macrologist:
I'm willing to close this and open up a discussion on the issues raised here.

stylewarning:
> I'm willing to close this and open up a discussion on the issues raised here.

Wrong PR I think. :)

examples/practical-coalton/15_pathnames.lisp
Jason94: One of the complexities in this library is there are two kinds of Lisp pathnames: wildcard pathnames and concrete pathnames. Some of the functions, like `list-directory` can only take a concrete pathname or a string. In the current implementation I'm using `(Path :a) => :a -> ...` to represent functions that can take any of the three, and `Result String Pathname` to represent functions that can only take a string or concrete pathname.

Is there a better way to encode that in the type signature?

stylewarning: Ah I see what you're saying. I personally think the function ought to just error. I don't think it's worth the additional complexity just to avoid a single instance.

Another is to have more layers of types and classes to express these constraints, but I don't think that's a good idea.

Jason94: That makes sense. Thanks for the comments! I'm going to be out of town through the weekend, but I'll try to get a revision up next week.

macrologist:
> > I'm willing to close this and open up a discussion on the issues raised here.
> 
> Wrong PR I think. :)

oops!  sorry

examples/practical-coalton/1_simple_database.lisp
Jason94: I tried that and it didn't work. I think it shouldn't work because these are the CL versions of those functions?

examples/practical-coalton/15_pathnames.lisp
```diff
@@ -0,0 +1,163 @@
...
+
+  (declare pathname->string (Pathname -> String))
+  (define (pathname->string pathname)
+    (lisp String (pathname)
+      (cl:write-to-string pathname)))
+  
+  (repr :native cl:pathname)
+  (define-type WildPathname)
+
+  ;; TODO: Is this safe? I want to be able to pass any of these
```
Jason94, line 61: Is this kosher? It saves a lot of typing, but I'm not sure if `(repr :native cl:t)` is dangerous to use for some reason.

stylewarning:
Hey @Jason94, I'm going to close this because of inactivity, but also because in its current state it's unlikely to fit as a good self-contained example for the purposes of this repo. If you decide to make your own repo, we can link to it. We're also happy to code review as you'd like on our Discord!

PR #1058, opened 2024-02-12
Improve testing coverage for error printer

jbouwman:
This PR adds a unit test that exercises source offset-aware error printing logic.

:sb-cover reports about 72% expression coverage in src/error.lisp


eliaslfox (approved):

colescott (approved):
looks like a good snapshot test for when we hack on the printer

Issue #961, opened 2023-06-20
Coalton-Lisp interop for structs


stylewarning:
I may have missed it, but I think we need to document the interop guidelines for Coalton-Lisp interop for structs, since the assessors are no longer "normal" named functions AFAIK.

PR #1065, opened 2024-03-04
Added documentation/recipe for Coalton struct lisp interop

Izaakwltn:
Addresses https://github.com/coalton-lang/coalton/issues/961
This adds documentation for how to use Coalton structs from Lisp, with two main recipes.


stylewarning:
My only protest is that this doesn't really tell me how to interop with structs, but rather that you have to grab data out yourself.

Is it worth being explicit about our promises or lack thereof about struct objects?

stylewarning:
Ping: This needs improved diction, capitalization, etc. and the above feedback needs to be addressed.

PR #1066, opened 2024-03-07
Generic error printer and stream access

jbouwman:
Move error printing from src/error.lisp to src/source-error.lisp, and make it independent of Coalton.

- Don't require a `file` argument when constructing errors; instead, bind a source file, or string, at compiler entry.
- Provide access to file information and contents via generics, to provide a common interface to string, file and displaced-file sources.
- Merge `coalton-error` structure into condition classes.
- Break up `display-coalton-error` into smaller functions.


jbouwman:
suspended until there is interest in working on error ergonomics

PR #1079, opened 2024-03-27
Drop unused exports and entry point

jbouwman:
- typechecker/entry.lisp is unused
- accessors in typechecker/define.lisp have moved to toplevel.lisp


eliaslfox:
`nreverse` is used to process definitions loosely in the order they appear in the file. This matters for type errors.

jbouwman:
> `nreverse` is used to process definitions loosely in the order they appear in the file. This matters for type errors.

ok, restored

stylewarning:
> > `nreverse` is used to process definitions loosely in the order they appear in the file. This matters for type errors.
> 
> 
> 
> ok, restored

don't you mean: "ok, reversed"

PR #1082, opened 2024-04-03
Deduplicate fundep environment entries

jbouwman:
Prevent addition of duplicate fundep entries by moving the no-match case inside the update block.

(A change to fundep environment type signature was proposed and rolled back.)

### DIFF
--- a/src/typechecker/environment.lisp
+++ b/src/typechecker/environment.lisp
@@ -1369,17 +1369,17 @@
                          (fundep-entry-from s)
                          from-tys
                          (fundep-entry-to s)
-                         to-tys))))))
+                         to-tys)))))
 
-              ;; Insert a new relation if there wasn't a match
-              (setf env
-                    (insert-fundep-entry%
-                     env
-                     (ty-class-name class)
-                     i
-                     (make-fundep-entry
-                      :from from-tys
-                      :to to-tys)))))
+                ;; Insert a new relation if there wasn't a match
+                (setf env
+                      (insert-fundep-entry%
+                       env
+                       (ty-class-name class)
+                       i
+                       (make-fundep-entry
+                        :from from-tys
+                        :to to-tys))))))
 
   env)
 


eliaslfox:
Why?

eliaslfox:
I don't think this is correct, and I'm confused as to why the tests are passing.

The fundep environment was previously:

`ImmutableMap<class, ImmutableMap<fundep-idx, ImmutableList<pred>>>`

This pr changes it to:

`ImmutableMap<class, List<pred>>`

Knowing which relations are associated w/ which fundeps is be required. See the following example:

```common-lisp
(in-package #:coalton-user)

(coalton-toplevel
  (define-class (C :a :b :c (:a -> :b) (:b -> :c)))
  
  (define-instance (C Integer String Double-Float)))

(in-package #:coalton-impl/typechecker/environment)

(lookup-fundep-environment coalton-impl/entry::*global-environment* 'coalton-user::C)

#S(IMMUTABLE-LISTMAP
   :DATA #{|
            (0
             #[
               #S(FUNDEP-ENTRY :FROM (COALTON:INTEGER) :TO (COALTON:STRING))
               #S(FUNDEP-ENTRY :FROM (COALTON:INTEGER) :TO (COALTON:STRING)) ])
            (1
             #[
               #S(FUNDEP-ENTRY
                  :FROM (COALTON:STRING)
                  :TO (COALTON:DOUBLE-FLOAT))
               #S(FUNDEP-ENTRY
                  :FROM (COALTON:STRING)
                  :TO (COALTON:DOUBLE-FLOAT)) ]) |}/#[ ])
```

Note how `Integer -> String` is attached to fundep #1 and `String -> Double-Float` is attached to fundep #2. I'm not sure about the duplicate entries, that looks like a bug lol.

jbouwman:
I've reverted changes to fundep environment: this changeset only prevents duplicate entries.

eliaslfox (approved):

PR #1083, opened 2024-04-05
Adds split functions for sequence types including lists, vectors, seqs and strings

Izaakwltn:
This adds splitting by single-character delimiters

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -241,11 +241,11 @@
   (define (nth-cdr n l)
     "Returns the nth-cdr of a list."
     (cond ((null? l)
-	   Nil)
-	  ((arith:zero? n)
-	   l)
-	  (True
-	   (nth-cdr (arith:1- n) (cdr l)))))
+           Nil)
+          ((arith:zero? n)
+           l)
+          (True
+           (nth-cdr (arith:1- n) (cdr l)))))
   
   (declare elemIndex (Eq :a => :a -> List :a -> Optional UFix))
   (define (elemIndex x xs)
@@ -604,13 +604,6 @@
            (any f xs)))
       ((Nil) False)))
 
-  (declare split (Char -> String -> (List String)))
-  (define (split c str)
-    (lisp (List String) (c str)
-      (cl:let ((split-chars (cl:list c)))
-        (cl:declare (cl:dynamic-extent split-chars))
-        (uiop:split-string str :separator split-chars))))
-
   (declare perms (List :a -> (List (List :a))))
   (define (perms l)
     "Produce all permutations of the list L."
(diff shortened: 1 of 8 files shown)


library/string.lisp
```diff
@@ -134,6 +136,32 @@ does not have that suffix."
   (define (chars str)
     "Returns an iterator over the characters in `str`."
     (iter:into-iter str))
+  
+  (declare delim-split (Char -> String -> (List String)))
```
stylewarning, line 140: can we just call it split?

stylewarning, line 140: it might be desirable to return an iterator

stylewarning (commented):
in the future it might be nice for this to work on any sequence type with any Eq element type

library/string.lisp
Izaakwltn: I was thinking about that- I'm not sure whether the other split is useful or even desirable.

Izaakwltn:
A few updates on the most recent commit:

- Added the `splittable` class which has a method `split` [`coalton-library/split`]
- Added `splittable` instances for `List` `Vector` and `Seq` containing elements with `Eq` instances.
- Renamed the existing `string:split` to `string:bisect`
- The new `string:split` calls `split:split` on a list of chars.

stylewarning:
I'm not 100% convinced this is a class we should include. My main reason is that it seems a little contrived: we wanted something generic so we forced it to be. I think the true generic aspect of something that splittable isn't that it's splittable, but rather that elements have a natural ordering (i.e., they can be indexed) and each element can be accessed in finite time.

My recommendation for the direction of this PR is to create a function `split` in each relevant module (no need to give distinguishing names), forgo the type class, and we can ponder in the future some nice classes to capture such sequential properties.

Izaakwltn:
Now this merely defines a function split for string, seq, list, and vector.

Izaakwltn:
No urgency, but ready for review whenever the next PR/Issue cleanup happens.

5/10:
- rebased onto main so it's still compatible
- cleaned up formatting

library/seq.lisp
```diff
@@ -207,6 +209,30 @@ a new `Seq` instance."
                    (rebalance-branches
                     (fold <> (butlast lsubts) (make-list nsubts (butfirst rsubts))))))))))))
 
+  (declare split ((Eq :a) (types:RuntimeRepr :a) => :a -> (Seq :a) -> (iter:Iterator (Seq :a))))
```
eliaslfox, line 212: This implementation needs to be streaming. The current version won't work with infinite iterators.

library/vector.lisp
```diff
@@ -247,6 +248,18 @@
...
 
+  (declare list->vec (List :a -> Vector :a))
+  (define (list->vec x)
+    (into x))
+
+  (declare vec->list (Vector :a -> List :a))
+  (define (vec->list x)
+    (into x))
+
+  (declare split ((Eq :a) => :a -> (Vector :a) -> (iter:Iterator (Vector :a))))
```
eliaslfox, line 259: Write an implementation that doesn't require converting to lists and back.

library/vector.lisp
```diff
@@ -247,6 +248,18 @@
      iter)
     Unit)
 
+  (declare list->vec (List :a -> Vector :a))
```
eliaslfox, line 251: don't define this

library/vector.lisp
```diff
@@ -247,6 +248,18 @@
...
+  (define (list->vec x)
+    (into x))
+
+  (declare vec->list (Vector :a -> List :a))
```
eliaslfox, line 255: don't define this

library/seq.lisp
Izaakwltn: I'm confused about this- 

I don't believe`Seq` can be infinite, so I'm not sure how a `seq` could or should generate an infinite/stream `Iterator`.

Izaakwltn: Or do you mean simply that I should also define split for `Iterator` and ensure infinite compatibility?

eliaslfox:
Add iter split:

```
(iter:for-each log
  (iter:split 3
  (map
    (fn (x)
      (log x)
      x)
    (iter:into-iter (1 2 3 2 3 1)))))


1
2
3
(1 2)
2
3
(2)
1
(1)
```

library/list.lisp
```diff
@@ -764,5 +780,7 @@ This function is equivalent to all size-N elements of `(COMBS L)`."
   (define-instance (Default (List :a))
     (define (default) Nil)))
 
+
```
eliaslfox, line 783: ?

eliaslfox:
Read [this](https://wiki.haskell.org/Bracket_pattern) and implement it for Coalton w/ `unwind-protect`.

stylewarning (changes requested):
some humble requests

Izaakwltn:
I'm going to close this for now

PR #1086, opened 2024-04-08
Make references to operators and operands consistent

jbouwman:
Replace s/rator/operator/g ; s/rands/operands/g for accessibility and consistency.

(Despite feeling a twinge of being a killjoy, because the naming is mildly amusing.)

### DIFF
--- a/src/codegen/ast-subsitutions.lisp
+++ b/src/codegen/ast-subsitutions.lisp
@@ -55,23 +55,23 @@
              (values node))
     (make-node-application
      :type (node-type node)
-     :rator (apply-ast-substitution subs (node-application-rator node))
-     :rands (apply-ast-substitution subs (node-application-rands node))))
+     :operator (apply-ast-substitution subs (node-application-operator node))
+     :operands (apply-ast-substitution subs (node-application-operands node))))
 
   (:method (subs (node node-direct-application))
     (declare (type ast-substitution-list subs)
              (values node))
 
-    (when (find (node-direct-application-rator node) subs :key #'ast-substitution-from)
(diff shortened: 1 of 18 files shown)


stylewarning:
*rator* and *rand* was terminology stolen from the Scheme world. William Byrd describes it at [40:27 in his presentation "The Most Beautiful Program Ever Written"](https://youtu.be/OyfBQmvr2Hc?t=40m27s). See also examples [1,2,3].

With that said, I can accept if this bit of idiom isn't clear and we prefer their wordier alternatives.

***

Refs.

1. https://www.cs.oberlin.edu/~rms/classes/cs275/labs/lab7/lab72.html
2. https://www.rose-hulman.edu/class/cs/csse304/201910/DailyNotes/32-ClassNotes.pdf
3. https://gerbil.scheme.org/tutorials/languages.html#custom-references-and-procedure-application

stylewarning:
I forgot to add the truly original reference where I learned of it first: *Essentials of Programming Languages* by Friedman and Wand. [Page 43](https://doc.lagout.org/science/0_Computer%20Science/1_Principles%20of%20Programming%20Languages/Essentials%20of%20Programming%20Languages%203e%20%28MIT%2C%202008%29.pdf) for an interface to procedure applications: `app-exp->rator` and `app-exp->rand`.

eliaslfox:
I don't think this is necessary.

jbouwman:
Thank you for the explanation; in light of the background it's preferable to document than to change. 

Good case of [Chesterton fencing](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)?

PR #1087, opened 2024-04-09
Generic environment

jbouwman:
Define a functional environment class for the type checker to use. Instead of reifying structures and collections for each partition of the environment (type, value, class, instance, fundep, specialization, name, code, etc.), use fset to build a two-level purely functional map.

A primary motivation is to simplify calculation of environment updates during compilation. Instead of wrapping each environment mutator in a macro that emits a copy of the envrionment-altering invocation, low-level environment set and unset operations can be calculated directly, by unwinding chained environment records back to the original environment record.

This also supports an arrangement where environment updates, generated code, AST info and other debugging artifacts are captured during compilation, rather then emitted all at once, as multiple return values from entry-point.

Generic access also simplifies utilities for dumping environment contents, and may count towards better support for incremental and partially-consistent definitions.

Scattered direct calls to functions in fset are removed.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -58,6 +58,7 @@
                 :serial t
                 :components ((:file "function-entry")
                              (:file "package")))
+               (:file "environment")
                (:module "typechecker"
                 :serial t
                 :components ((:file "kinds")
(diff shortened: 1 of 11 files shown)


jbouwman:
Closing, to revisit after file work

PR #1091, opened 2024-04-11
Package import, import-from and export statements

jbouwman:
Extend the `package` form to support full package definitions.

Supported syntax:

``` coalton
(package <package-name>
  "documentation string"
  (import <other-package-name>
          (<other-package-name2 as <other2>)
          ...)
  (import-from <other-package-name>
          <symbol1>
          <symbol2>
          ...)
  (export <symbol3>
          <symbol4>
          ...))
```

Example:

``` coalton
(package my-package
  (import coalton-library/list)
  (export sum-even-numbers))
```

## Import

`import` allows you to specify which packages a package depends on:

``` coalton
(package my-package
  (import coalton-library/list)
  ...)
```

You can use nicknames for imported packages using the `as` keyword:

``` coalton
(package my-package
  (import (coalton-library/list as list))
  ...)
```

## Import-from

The `import-from` form allows you to specify which symbols from an imported package should be made available within your package:

``` coalton
(package my-package
  (import-from coalton-library/list
    filter)
  ...)
```

## Export

``` coalton
(package my-package
  ...
  (export sum-even-numbers))
```


tests/toplevel-tests.lisp
```diff
@@ -0,0 +1,66 @@
+(in-package #:coalton-tests)
+
+(deftest test-parse-package ()
```
Izaakwltn, line 3: I think before this goes through at least one example in `small-coalton-programs` should be retrofitted with this package form, or maybe `thih` since it's already incorporated into CI.

src/parser/toplevel.lisp
```diff
@@ -573,32 +666,85 @@ consume all attributes"))))
...
+        (error 'parse-error
+               :err (coalton-error
+                     :span (cons (- (file-position stream) 2)
+                                 (- (file-position stream) 1))
+                     :file file
+                     :message "Unexpected EOF"
+                     :primary-note "missing package form")))
+      (parse-package form file))))
+
+(defun lisp-package (toplevel-package file)
```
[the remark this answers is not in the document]
jbouwman: added documentation to both main functions

docs/package.md
```diff
@@ -0,0 +1,90 @@
...
+(package my-package
+  (import coalton-library/list as list)
+  (export sum-even-numbers))
+```
+
+The `import-from` form allows you to specify which symbols from an
+imported package should be made available within your package:
+
+``` coalton
+(package my-package
```
Izaakwltn, line 56:
```
(O package, my-package
  (export
    our 
    fearful 
    trip 
    is 
    done))

File: src/parser/toplevel.lisp [hunk 4]
```diff
@@ -549,11 +535,118 @@ consume all attributes"))))
...
+         (body (cst:rest clause))
+         (clause-type (cst:raw head)))
+    (cond ((string= clause-type "IMPORT")
+           (do-forms (lambda (form)
+                       (parse-import-statement toplevel-package form file))
+             body))
+          ((string= clause-type "EXPORT")
+           (add-export toplevel-package (map-forms (lambda (form)
+                                                     (parse-symbol form file "Malformed export statement"))
+                                                   body)))
```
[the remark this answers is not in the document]
jbouwman: Agree on nicknames, they're already common in Coalton codebase, for example, https://github.com/coalton-lang/coalton/blob/66c8db92383c6622562ebab8a846bebdcd951e46/examples/thih/src/thih.lisp

Propose to cover both cases with `(import pkg-x)` meaning import all symbols from pkg-x, and `(import (pkg-x as x))` meaning make all exported symbols in pkg-x available under prefix `x`.
```(import package-x)

File: src/parser/toplevel.lisp [hunk 5]
```diff
@@ -573,32 +666,85 @@ consume all attributes"))))
...
+                                 (- (file-position stream) 1))
+                     :file file
+                     :message "Unexpected EOF"
+                     :primary-note "missing package form")))
+      (parse-package form file))))
+
+(defun lisp-package (toplevel-package file)
+  (with-slots (name import export) toplevel-package
+    (let ((package (uiop:ensure-package name)))
+      (use-package (find-package "COALTON") package)
```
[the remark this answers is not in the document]
jbouwman: clarified the guarantees, and created doc/package.md

I'm inclined not to support 'midway' packages for now, since they are relatively uncommon, can be supported by keeping those source files as lisp+coalton-toplevel, and they may place new requirements on parsing and code generation.

tests/toplevel-tests.lisp
jbouwman: added a version of thih.coal and will feed it to the parser in a test case (upcoming)

docs/package.md
jbouwman: can also allow parser to accept "exhort" for "export"

eliaslfox:
I would prefer to merge asdf support for .coal files before this. Without that it's difficult to tell if the compiler is generating valid fasls that can survive restarts. 

The [initial Coalton prototype](https://github.com/stylewarning/deprecated-coalton-prototype) had a problem where restarts would clear the type database. Solving this required building the current environment system. I think similar care will be needed to ensure that packages are re-created when cached fasls are loaded.

jbouwman:
> I would prefer to merge asdf support for .coal files before this. Without that it's difficult to tell if the compiler is generating valid fasls that can survive restarts.

#1111 contains support for coal->fasl compilation: there is an open and blocking question there over the necessity to include a new generic to supplement and possibly supersede the use of make-load-form.

src/parser/toplevel.lisp
```diff
@@ -549,11 +545,127 @@ consume all attributes"))))
...
+             (pushnew (list (package-symbol-name source-nick)
+                            (package-symbol-name source-package))
+                      import-as))))
+        (t
+         (with-slots (import) toplevel-package
+           (pushnew (package-symbol-name (parse-symbol form file "Malformed import statement"))
+                    import)))))
+
+(defun parse-package-clause (toplevel-package clause file)
+  "Parses a coalton package clause of the form of ({operation} {symbol}* ...)"
```
jbouwman, line 623: weird and obsolete docstring

docs/package.md
```diff
@@ -0,0 +1,91 @@
+# The Coalton `(package ...)` Form
+
+In Coalton, the `package` form is used to define a package by
+specifying which dependencies to import, and which definitions to
```
Izaakwltn, line 4: Technically it's which symbols to export?

docs/package.md
```diff
@@ -0,0 +1,91 @@
...
+export.
+
+The form also establishes the current package: any definitions that
+follow `(package <name>)` will be contained within that package.
```
Izaakwltn, line 8: maybe something like 

"... This replaces the explicit call of `(in-package <name>)` used in Lisp programs."

docs/package.md
```diff
@@ -0,0 +1,91 @@
...
+  (import-from <other-package-name>
+          <symbol1>
+          <symbol2>
+          ...)
+  (export <symbol3>
+          <symbol4>
+          ...))
+```
+
+An example:
```
Izaakwltn, line 29: Maybe an example with all bells and whistles accounted for:

````
``` coalton
(package my-package
  (import coalton-library/list
          (coalton-library/vector as vec))
  (import-from coalton-library/char
          upcase
          downcase)
  (export sum-even-numbers))
```
````

File: docs/package.md [hunk 10]
```diff
@@ -0,0 +1,91 @@
...
+  "Package documentation string (optional)"
+  (import-from coalton-library/list
+    filter)
+  (export sum-even-numbers))
+```
+
+## Exporting Definitions
+
+You can export definitions from your package using the `export`
+form. This allows other packages to use to these definitions.
```
Izaakwltn, line 66: too many `to`s

docs/package.md
```diff
@@ -0,0 +1,91 @@
...
+For example, if you have a function `sum-even-numbers` that you
+want to make available for use by other packages:
+
+``` coalton
+(package my-package
+  (import coalton-library/list)
+  (export sum-even-numbers))
+```
+
+In this case, the `sum-even-numbers` function will be exported from
```
Izaakwltn, line 77: maybe symbol instead of function

docs/package.md
```diff
@@ -0,0 +1,91 @@
...
+your package and can be used by other packages that import your
+package.
+
+## Lisp Interoperability
+
+Packages defined by `package` are fundamentally structurally compatible
+with Lisp packages, as the import, import-from, and export forms
+evaluate unambiguously to Lisp defpackage counterparts `use`,
```
Izaakwltn, line 85:
```
`defpackage`
```

Izaakwltn:
```
`:use`, `:local-nicknames`, and `:export`.
```

docs/package.md
```diff
@@ -0,0 +1,91 @@
...
+`local-nicknames`, and `export`.
+
+- Any given Coalton package can be imported into any Lisp package.
+- Any given Lisp package be imported into any Coalton package.
+- The name of a coalton-defined package be the same whether importing into Coalton or Lisp.
+- Coalton packages are compatible with utilities such as uiop:define-package's `:mix-reexport` clause.
```
Izaakwltn, line 91:
```
`uiop:define-package`'s
```

examples/small-coalton-programs/src/oldies.coal
```diff
@@ -0,0 +1,13 @@
+(package small-coalton-programs/oldies
+  "The venerable fact and fib functions"
```
Izaakwltn, line 2: It seems to be a fib about `fib` (counter to the fact of `fact`).

Potential "serious" package name- `recursive-basics`.

examples/small-coalton-programs/src/oldies.coal
```diff
@@ -0,0 +1,13 @@
...
+  (import coalton-prelude)
+  (export fact))
+
+
+(declare fact (Integer -> Integer))
+(define (fact n)
+  "Compute the factorial of a given integer"
+  (if (== n 0)
```
Izaakwltn, line 10:
```
(import-from `coalton-library/math` 
            zerop)
```
```
(if (zero? n)
```

examples/small-coalton-programs/src/oldies.coal
```diff
@@ -0,0 +1,13 @@
...
+    1
+    (* n
+      (fact (- n 1)))))
```
Izaakwltn, line 13: `(1- n)`

Izaakwltn (commented):
Looks good to me, just a few nits about docs and the test-case/example

eliaslfox:
I'm trying this out locally and I can't get asdf to compile a file. 

test.asd
```
(asdf:defsystem #:test
  :serial t
  :pathname "src"
  :depends-on (#:coalton #:coalton-asdf)
  :components ((:coalton-file "test.coal")))
```

src/test.coal
```
(package test)

(define x 5)
```

```
CL-USER> (asdf:load-system :test)
T
CL-USER> (in-package :test)
The name "TEST" does not designate any package.
```

Izaakwltn:
> I'm trying this out locally and I can't get asdf to compile a file.
> 
> test.asd
> 
> ```
> (asdf:defsystem #:test
>   :serial t
>   :pathname "src"
>   :depends-on (#:coalton #:coalton-asdf)
>   :components ((:coalton-file "test.coal")))
> ```
> 
> src/test.coal
> 
> ```
> (package test)
> 
> (define x 5)
> ```
> 
> ```
> CL-USER> (asdf:load-system :test)
> T
> CL-USER> (in-package :test)
> The name "TEST" does not designate any package.
> ```

I think the system needs `:system-depends-on (#:coalton-asdf)`

examples/small-coalton-programs/src/fact-fib.coal
```diff
@@ -0,0 +1,28 @@
...
+
+(declare fact (Integer -> Integer))
+(define (fact n)
+  "Compute the factorial of a given integer"
+  (if (zero? n)
+    1
+    (* n
+       (fact (- n 1)))))
+
+(declare fib (Integer -> Integer))
```
Izaakwltn, line 20: Slightly more coalton-y:

```
  (declare fib (Integer -> Integer))
  (define (fib n)
    "Compute the nth Fibonacci number."
    (match n
      (0 0)
      (1 1)
      (_ (+ (fib (1- n))
             (fib (- n 2))))))
```

examples/small-coalton-programs/src/fact-fib.coal
```diff
@@ -0,0 +1,28 @@
...
+  (import
+    coalton-prelude)
+  (import-from
+    coalton-library/math
+    zero?)
+  (export
+    fact
+    fib))
+
+(declare fact (Integer -> Integer))
```
Izaakwltn, line 12: Also slightly more coalton-y:

```
  (declare fact (Integer -> Integer))
  (define (fact n)
    "Compute the factorial of a given integer."
    (match n
      (0 1)
      (_ (* n
           (fact (1- n))))))
```

examples/small-coalton-programs/src/fact-fib.coal
jbouwman: done

src/codegen/codegen-expression.lisp
```diff
@@ -136,7 +136,7 @@
     (let ((match-expr (codegen-expression (node-while-let-expr expr) current-function env))
           (body-expr (codegen-expression (node-while-let-body expr) current-function env))
           (label (node-while-let-label expr))
-          (match-var (gensym "MATCH")))
+          (match-var (gentemp "MATCH")))
```
jbouwman, line 139: Gensyms are not interned so can't be used as bindings in code that will survive a prin1/read cycle; however gentemp will litter the codegen package. Define a temporary package for these?

eliaslfox:
> I think the system needs :defsystem-depends-on (#:coalton-asdf)

Thanks! I added defsystem-depends-on and got a different error.

```
Error while trying to load definition for system meh from
pathname /Users/ejlawsonfox/lisp/meh/meh.asd:
   COMPILE-FILE-ERROR while
   compiling #<CL-SOURCE-FILE "coalton-asdf" "coalton-asdf">
```

jbouwman:
>  READ error during COMPILE-FILE: Symbol "COMPILE-COALTON" not found in the COALTON-IMPL/ENTRY package
> ```

I wonder if it's looking at a different coalton system, that symbol seems to be present and exported.

eliaslfox:
> I wonder if it's looking at a different coalton system, that symbol seems to be present and exported.

Good catch. My lisp environment was screwed up.

The system is compiling now but it doesn't look like the compiled code is being run. I get compile errors, but I still can't in-package.

Izaakwltn:
> The system is compiling now but it doesn't look like the compiled code is being run. I get compile errors, but I still can't in-package.

I just tried and it's working for me

eliaslfox:
That works for me. I was doing something else wrong :/

I think I ran into an actual issue. The fasl files produced don't appear to recreate their packages.

```
(asdf:load-system :asd-test)
;; restart inferior lisp
(asdf:load-system :asd-test)
```

I got the following error:
```
The loader tried loading the symbol named X into the package named TEST, but the package did not get defined, and does not exist.
```

src/codegen/codegen-expression.lisp
eliaslfox: This *should* be fine. `gentemp` will by default intern into `*package*` which should be bound to the package being compiled into.

stylewarning:
(Side comment: Can we test that ASDF corner cases work or fail appropriately? Like bad extension, missing file, etc. In a separate PR maybe.)

jbouwman:
> The loader tried loading the symbol named X into the package named TEST, but the package did not get defined, and 

Aha, it wasn't emitting the defpackage form: I've restored that -- https://github.com/coalton-lang/coalton/pull/1091/files#diff-dbe076ada6d143bc146af69913f741b08735798471bd7bc03e0ca4a2dd7d089fR240 -- followed your steps, and it now appears to work.

jbouwman:
> (Side comment: Can we test that ASDF corner cases work or fail appropriately? Like bad extension, missing file, etc. In a separate PR maybe.)

https://github.com/coalton-lang/coalton/pull/1091/files#diff-dbe076ada6d143bc146af69913f741b08735798471bd7bc03e0ca4a2dd7d089fR240 might be a good place to add those

Also, touching each error case for invalid package forms.

coalton-asdf.asd
```diff
@@ -0,0 +1,23 @@
+;;; Coalton ASDF extensions
```
eliaslfox, line 1: Is there a reason this needs to be a separate asdf system?

src/entry.lisp
```diff
@@ -230,10 +233,13 @@
...
       (let* ((file (error:make-coalton-file :stream input
                                             :name name))
-             (program (entry-point (parser:read-program input file :mode :file))))
+             (program (parser:read-program input file :mode :file))
+             (program-text (entry-point program)))
         (print-form output (make-prologue))
         (print-form output (make-environment-updater updates))
-        (print-form output program)))))
+        (print-form output (parser:make-defpackage (parser:program-package program)))
+        (print-form output '(in-package :cl))
```
eliaslfox, line 241: why?

src/parser/toplevel.lisp
```diff
@@ -11,6 +11,15 @@
   (:shadowing-import-from
    #:coalton-impl/parser/base
    #:parse-error)
+  (:import-from
```
eliaslfox, line 14: Use trivial-package-local-nicknames

src/parser/toplevel.lisp
```diff
@@ -554,56 +544,214 @@ consume all attributes"))))
 
       (parse-expression form file))))
 
+;;; Packages
```
eliaslfox, line 547: Please rewrite this so that it matches the conventions used throughout the rest of the parser. The custom iterator implementation seems unnecessary.

src/parser/toplevel.lisp
```diff
@@ -554,56 +544,214 @@ consume all attributes"))))
...
+  (with-slots (name import import-as import-from export) toplevel-package
+    (let ((package (uiop:ensure-package name
+                                        :use (cons "COALTON" import)
+                                        :import-from import-from
+                                        :export export)))
+      (loop :for (local-nickname actual-package) :in import-as
+            :do (add-package-local-nickname actual-package package))
+      package)))
+
+(defun program-lisp-package (program)
```
eliaslfox, line 737: Don't do this. Store the correct package value on the program struct.

src/entry.lisp
```diff
@@ -267,3 +272,16 @@
            (t
             (with-output-to-string (lisp-stream)
               (compile-to-lisp coal-file-name coal-stream lisp-stream)))))))))
+
+(defun compile (stream)
```
eliaslfox, line 276: When is this used?

src/parser/toplevel.lisp
```diff
@@ -554,56 +544,214 @@ consume all attributes"))))
...
+                          :span (cons (- (file-position stream) 2)
+                                      (- (file-position stream) 1))
+                          :file file
+                          :message "Unexpected EOF"
+                          :primary-note "missing package form")))
+           (parse-package form file))
+      (do-symbols (symbol)
+        (unintern symbol)))))
+
+(defun lisp-package (toplevel-package)
```
eliaslfox, line 724: This function and `make-defpackage` will have subtly different behavior. They need to share a single implementation.

eliaslfox (dismissed):

src/entry.lisp
```diff
@@ -217,56 +220,56 @@
...
+
+(defun compile (stream name &key (load t) (output-file nil))
+   "Compile Coalton code in STREAM, returning the pathname of the generated .fasl file. If OUTPUT-FILE is nil, the built-in compiler default output location will be used."
+   (uiop:with-temporary-file (:stream lisp-stream
+                              :pathname lisp-file
+                              :type "lisp"
+                              :direction ':output)
+     (compile-to-lisp stream name lisp-stream)
+     :close-stream
+     (cond (output-file
```
jbouwman, line 269: (not (null output-file))

tests/toplevel-tests.lisp
jbouwman: actually this is now 'fact-fib.coal'

tests/entry-tests.lisp
```diff
@@ -1,49 +1,34 @@
...
-   #:coalton-prelude
-   #:coalton-library/math)
-  (:local-nicknames (#:list #:coalton-library/list))
-  (:export
-   #:fact))
+  "Test that the Coalton compiler compiles a test file into something that looks like Lisp source."
+  (with-open-file (input (compile-test-file) :direction ':input)
+    (let ((source (entry:codegen input "test")))
+      (with-input-from-string (stream source)
+        (let ((form-types (loop :for form := (read stream nil nil) :while form :collect (first form))))
```
jbouwman, line 11: bring out as a helper, this looks unclear

coalton-asdf.asd
jbouwman: Yes, there is a cyclic load error, because secondary systems are presumed dependencies of the primary system, which prevents the system from accepting coalton/compiler as a dependency.

```
Unhandled LOAD-SYSTEM-DEFINITION-ERROR in thread #<SB-THREAD:THREAD tid=259 "main thread" RUNNING {7005000003}>: Error while trying to load definition for system coalton from pathname /Users/jlbouwman/git/coalton/coalton.asd:
Circular dependency:
((#<DEFINE-OP > . #<SYSTEM "coalton">)
 (#<LOAD-OP > . #<SYSTEM "coalton/asdf">)
 (#<LOAD-OP > . #<CL-SOURCE-FILE "coalton/asdf" "coalton-asdf">)
 (#<PREPARE-OP > . #<CL-SOURCE-FILE "coalton/asdf" "coalton-asdf">)
 (#<PREPARE-OP > . #<SYSTEM "coalton/asdf">)
 (#<LOAD-OP > . #<SYSTEM "coalton/compiler">)
 (#<LOAD-OP > . #<CL-SOURCE-FILE "coalton/compiler" "package">)
 (#<PREPARE-OP > . #<CL-SOURCE-FILE "coalton/compiler" "package">)
 (#<PREPARE-OP > . #<SYSTEM "coalton/compiler">))
```

src/parser/toplevel.lisp
```diff
@@ -554,56 +536,213 @@ consume all attributes"))))
...
+       (:use "COALTON" ,@import)
+       ,@(mapcar (lambda (form)
+                   `(:import-form ,@form))
+                 import-from)
+       (:local-nicknames ,@import-as)
+       (:export ,@export))))
+
+(defun lisp-package (toplevel-package)
+  "Create a Lisp package by evlauating the defpackage representation of a TOPLEVEL-PACKAGE form."
+  (eval (make-defpackage toplevel-package)))
```
macrologist, line 737: If constructing the package "by hand" using `make-package` etc is too onerous, this is an acceptable alternative. 

I'd probably use `make-package` though, as it allows for a greater degree of control and doesn't call up the whole evaluator.

macrologist (approved):
Looks good to me. The interface `package` has everything reasonable we'd want while remaining minimal and thus open to change.

src/parser/toplevel.lisp
jbouwman: That was the first thing I explored, and the main thing with that was, there are a number of special mutation functions (e.g., those for package local nicknames) that have varying forms in the supported Lisp implementations. There was also a remark here about potentially diverging behavior, so the eval is a cheap and effective way of getting consistency between coalton-parse and lisp-load time package definitions, and retaining CL package redefinition semantics (for now).

src/parser/toplevel.lisp
```diff
@@ -554,56 +536,213 @@ consume all attributes"))))
...
+                          :message "Unexpected EOF"
+                          :primary-note "missing package form")))
+           (parse-package form file))
+      (do-symbols (symbol)
+        (unintern symbol)))))
+
+(defun make-defpackage (toplevel-package)
+  "Generate a Lisp defpackage form from a Caolton toplevel-package structure."
+  (with-slots (name import import-from import-as export) toplevel-package
+    `(defpackage ,name
```
eliaslfox, line 727: uiop:define-package

src/parser/toplevel.lisp
```diff
@@ -554,56 +536,213 @@ consume all attributes"))))
...
+       (:use "COALTON" ,@import)
+       ,@(mapcar (lambda (form)
+                   `(:import-from ,@form))
+                 import-from)
+       (:local-nicknames ,@import-as)
```
eliaslfox, line 732: trivial-package-local-nicknames

jbouwman:
Outcome of steering meeting:

- [ ] replace defclass with defstruct
- [ ] prove that parse doesn't accept garbage input, and produce bad errors

jbouwman:
@stylewarning @eliaslfox guidance on this changeset?

tests/parser-test-files/package.txt
stylewarning, line 1: NICE ERRORS!

tests/entry-tests.lisp
```diff
@@ -1,49 +1,39 @@
...
-   #:coalton
-   #:coalton-prelude
-   #:coalton-library/math)
-  (:local-nicknames (#:list #:coalton-library/list))
-  (:export
-   #:fact))
+(deftest test-compile-to-lisp ()
+  "Test that the Coalton compiler compiles a test file into something that looks like Lisp source."
+  (with-open-file (input (compile-test-file) :direction ':input)
+    (let ((source-form-types (mapcar #'first
```
stylewarning, line 15: nit: prefer the newline break before the `(mapcar`

src/parser/toplevel.lisp
```diff
@@ -1857,7 +1925,7 @@ consume all attributes")))
                  :message "Unexpected form"
                  :primary-note "expected an identifier")))
 
-  (when (string= "_" (symbol-name (cst:raw form)))
+  (when (string-equal "_" (cst:raw form))
```
stylewarning, line 1928: surely `string=` is OK here

src/parser/toplevel.lisp
```diff
@@ -554,56 +547,135 @@ consume all attributes"))))
...
+  "Generate a Lisp defpackage form from a Caolton toplevel-package structure."
+  `(defpackage ,(toplevel-package-name package)
+     (:use "COALTON" ,@(toplevel-package-import package))
+     ,@(mapcar (lambda (form)
+                 `(:import-from ,@form))
+               (toplevel-package-import-from package))
+     (:local-nicknames ,@(toplevel-package-import-as package))
+     (:export ,@(toplevel-package-export package))))
+
+(defun lisp-package (package)
```
stylewarning, line 668: Can we use `make-package` instead of `eval`?

src/parser/toplevel.lisp
```diff
@@ -481,6 +473,7 @@
                      :classes nil
                      :instances nil
                      :specializations nil))
+           (*package* (program-lisp-package program))
 
            (attributes (make-array 0 :adjustable t :fill-pointer t)))
```
stylewarning, line 478: cool, i forgot that `fill-pointer` can be `t`

src/parser/toplevel.lisp
```diff
@@ -448,31 +457,14 @@
...
 
-         ;; Initial package to read (package) forms into
-         (*package* *package*))
+         ;; In mode :file, the value of package is a toplevel-package structure
+         (package nil))
 
-    ;; Parse package form
-    (when (eq :file mode)
-      (setf *package* (find-package "COALTON-IMPL/PARSER/READ"))
+    (when (eql mode :file)
```
stylewarning, line 463: do i even have to say it?

(also use `eq`)

src/parser/toplevel.lisp
```diff
@@ -432,14 +446,9 @@
...
-;; Empty package for reading (package) forms
-;;
-(defpackage #:coalton-impl/parser/read
-  (:use))
-
-(defun read-program (stream file &key (mode (error "you must supply this bozo!")))
-  "Read a PROGRAM from the SE:FILE."
+(defun read-program (stream file &key mode)
+  "Read a PROGRAM from FILE (an instance of source-error:file).
+If MODE is :file, a package form is required."
```
stylewarning, line 451: Document all acceptable `mode`s here and their meanings.

stylewarning (changes requested):
I reviewed everything except the cursor code; I'll follow up with that.

Everything looks pretty good.

src/parser/toplevel.lisp
jbouwman: Only reason not to is `make-package` doesn't support package-local nicknames. There may be a good opportunity to revisit package invariants when specifying how redefinitions work during interactive development. I'm thinking in particular of how clunky changes to package exports are, when working with common lisp.

src/parser/toplevel.lisp
jbouwman: done

tests/toplevel-tests.lisp
```diff
@@ -0,0 +1,68 @@
...
+            second
+            third))")))
+
+(deftest test-lisp-package ()
+  "Lisp packages can be constructed from parsed Coalton package forms."
+  (flet ((del-pkg (package-designator)
+           (when (find-package package-designator)
+             (delete-package package-designator)))
+         (ext-syms (p)
+           (let (l)
```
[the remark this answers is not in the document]
jbouwman: done

tests/utilities.lisp
```diff
@@ -95,3 +95,27 @@ Returns (values SOURCE-PATHNAME COMPILED-PATHNAME)."
...
+(defun run-suite (pathname)
+  (let ((file (test-file pathname)))
+    (loop :for (line description program expected-error)
+            :in (coalton-tests/loader:load-suite file)
+          :for generated-error := (collect-compiler-error program)
+          :do (cond ((null generated-error)
+                     (is nil "program should have failed to compile: ~A" description))
+                    (t
+                     (check-string= (format nil "program text.~%~
+input file: ~A~%line number: ~A~%test case: ~A~%expected error (A) and generated error (B)"
```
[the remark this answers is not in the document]
jbouwman: done

src/parser/toplevel.lisp
jbouwman: done

tests/loader.lisp
```diff
@@ -0,0 +1,112 @@
...
+          (loader-error state) nil)
+    state))
+
+(defun start-case (state)
+  (with-next-state state
+    (setf (loader-start state) (loader-line state))
+    state))
+
+(defun line-start-p (line char)
+  (and (< 0 (length line))
```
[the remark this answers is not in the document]
jbouwman: done

tests/loader.lisp
```diff
@@ -0,0 +1,112 @@
...
+           (ecase (loader-state state)
+             (:program (next-state state :error))))
+          (t
+           (collect-line state line)))))
+
+(defun load-suite (pathname)
+  (with-open-file (stream pathname :direction ':input :element-type 'character)
+    (loop :for state := (make-loader)
+            :then (process-line state (read-line stream nil nil))
+          :when (eq :done (loader-state state))
```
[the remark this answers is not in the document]
jbouwman: done

tests/toplevel-tests.lisp
```diff
@@ -0,0 +1,68 @@
...
+  (with-input-from-string (stream string)
+    (parser:with-reader-context stream
+      (let ((form (parser:maybe-read-form stream parser::*coalton-eclector-client*)))
+        (coalton-impl/parser/toplevel::parse-package
+         (coalton-impl/parser/cursor:make-cursor form))))))
+       
+
+(deftest test-parse-package ()
+  "Coalton toplevel package forms are successfully parsed."
+  (is (parse-package
```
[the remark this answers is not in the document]
jbouwman: done

src/parser/toplevel.lisp
jbouwman: using eval form for now - there will be a subsequent MR with missing package forms ('shadow') that would be a good place to make semantics of (package) final

stylewarning (approved):

PR #1097, opened 2024-04-26
tc-env-lookup-value: suggest type constructors

### LINKED ISSUE #1095: Using Type name instead of constructor returns opaque `unknown variable` error
This is a pretty minor nit, but when you call the type name instead of the constructor:

```
(defpackage #:unknown-variable-type-error
  (:use #:coalton
        #:coalton-prelude))

(in-package #:unknown-variable-type-error)

(coalton-toplevel

  (define-type DataType
    (Data1 Integer)
    (Data2 String))

  (define (return-data int)
    (DataType int)))
```
It returns the error: 
```
(DATATYPE INT)))
 ^^^^^^^^ unknown variable
```
Which seems vague, and also, technically it's not an _unknown_ variable.

jbouwman:
On variable lookup miss, existing suggestion code has no effect because it looks at the two places just checked. Instead check the environment to see if the var names a type, meaning, maybe the user was looking for a type constructor.

Closes #1095

### DIFF
--- a/src/typechecker/tc-env.lisp
+++ b/src/typechecker/tc-env.lisp
@@ -30,10 +30,10 @@
 (defstruct (tc-env
             (:predicate nil))
 
-  ;; The main copiler env
+  ;; The main compiler env
   (env      (util:required 'env)          :type tc:environment :read-only t)
 
-  ;; Hash table mappinig variables bound in the current translation unit to types
+  ;; Hash table mapping variables bound in the current translation unit to types
   (ty-table (make-hash-table :test #'eq)  :type hash-table     :read-only t))
 
 (defun tc-env-add-variable (env name)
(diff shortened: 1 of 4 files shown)


src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,58 @@
...
+  "If value lookup failed, generate suggestions for what to do, if anything."
+  (let ((suggestions nil))
+    (flet ((suggest (fmt-string &rest fmt-args)
+             (push (apply #'format nil fmt-string fmt-args) suggestions)))
+
+      ;; If the symbol names a type, user may have intended to use a type constructor
+      (a:when-let ((type (tc:lookup-type (tc-env-env env) var-name :no-error t)))
+        (suggest "Did you mean a constructor of type ~A?" (tc:type-entry-name type)))
+
+      ;; ...
```
stylewarning, line 69: can you write a little note here?

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,58 @@
...
+      )
+    (nreverse suggestions)))
+
 (defun tc-env-lookup-value (env var file)
-  "Lookup a variable named VAR in ENV."
+  "Lookup a value named VAR in ENV."
```
stylewarning, line 74: "Lookup the value of the variable named VAR in ENV." maybe?

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,58 @@
...
+                     (lookup-var env var-name))))
+    (unless scheme
+      ;; Variable is unbound: create an error
+      (error 'tc-error
+             :err (coalton-error
+                   :span (parser:node-source var)
+                   :file file
+                   :message "Unknown variable"
+                   :primary-note "unknown variable"
+                   :help-notes (mapcar
```
stylewarning, line 91: there's a mapcar here and a loop-collect below. can we just make this a loop-collect as well?

stylewarning (commented):
love this kind of usability enhancement. just some bitty things

stylewarning (approved):
lgtm, would like @eliaslfox to peek too

src/typechecker/tc-env.lisp
```diff
@@ -4,6 +4,7 @@
    #:coalton-impl/typechecker/base
    #:coalton-impl/typechecker/parse-type)
   (:local-nicknames
+   (#:a #:alexandria)
```
eliaslfox, line 7: Don't PLN alexandria.

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
 
   (tc:qualified-ty-type (tc:fresh-inst (setf (gethash name (tc-env-ty-table env)) (tc:to-scheme (tc:make-variable))))))
 
+(defun lookup-var-type (env var-name)
+  "Look up a type binding of VAR-NAME in type checker environment ENV."
+  (gethash var-name (tc-env-ty-table env)))
+
+(defun lookup-var (env var-name)
```
eliaslfox, line 55: Don't define this function.

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
 
   (tc:qualified-ty-type (tc:fresh-inst (setf (gethash name (tc-env-ty-table env)) (tc:to-scheme (tc:make-variable))))))
 
+(defun lookup-var-type (env var-name)
```
eliaslfox, line 51: Don't define this function.

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
...
+  (tc:lookup-value-type (tc-env-env env) var-name :no-error t))
+
+(defun tc-env-suggest-value (env var-name)
+  "If value lookup failed, generate suggestions for what to do, if anything."
+  (let ((suggestions nil))
+    (flet ((suggest (fmt-string &rest fmt-args)
+             (push (apply #'format nil fmt-string fmt-args) suggestions)))
+
+      ;; If the symbol names a type, user may have intended to use a type constructor
+      (a:when-let ((type (tc:lookup-type (tc-env-env env) var-name :no-error t)))
```
eliaslfox, line 66: Don't use `when-let`

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
...
+  (gethash var-name (tc-env-ty-table env)))
+
+(defun lookup-var (env var-name)
+  "Look up a variable binding of VAR-NAME in type checker environment ENV."
+  (tc:lookup-value-type (tc-env-env env) var-name :no-error t))
+
+(defun tc-env-suggest-value (env var-name)
+  "If value lookup failed, generate suggestions for what to do, if anything."
+  (let ((suggestions nil))
+    (flet ((suggest (fmt-string &rest fmt-args)
```
eliaslfox, line 62: Just write this inline.

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
...
 
+(defun lookup-var-type (env var-name)
+  "Look up a type binding of VAR-NAME in type checker environment ENV."
+  (gethash var-name (tc-env-ty-table env)))
+
+(defun lookup-var (env var-name)
+  "Look up a variable binding of VAR-NAME in type checker environment ENV."
+  (tc:lookup-value-type (tc-env-env env) var-name :no-error t))
+
+(defun tc-env-suggest-value (env var-name)
```
eliaslfox, line 59: Add type annotations to this function.

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
...
+
+      ;; If the symbol names a type, user may have intended to use a type constructor
+      (a:when-let ((type (tc:lookup-type (tc-env-env env) var-name :no-error t)))
+        (suggest "Did you mean a constructor of type ~A?" (tc:type-entry-name type)))
+
+      ;; There is more we can do here to return useful suggestions,
+      ;; including checking other places in the environment and
+      ;; catching simple spelling errors.
+
+      )
```
eliaslfox, line 73: Paren goes on line 67

src/typechecker/tc-env.lisp
```diff
@@ -47,57 +48,59 @@
...
+    (nreverse suggestions)))
+
 (defun tc-env-lookup-value (env var file)
-  "Lookup a variable named VAR in ENV."
+  "Lookup the value of a variable named VAR in ENV."
```
eliaslfox, line 77: the type of

eliaslfox (changes requested):

src/typechecker/tc-env.lisp
stylewarning: i'll accept the PLN

    (#:the-library-of-alexandria #:alexandria)

however. send tweet

jbouwman: As a compromise, it could be referred to via the Mongolian vowel separator https://github.com/jbouwman/coalton/pull/2

eliaslfox (approved):

PR #1101, opened 2024-05-06
Adds error-type to Tryinto to allow for non-string errors

Izaakwltn:
This adds error-types to `Tryinto` so that it can return non-string errors.

### DIFF
--- a/library/classes.lisp
+++ b/library/classes.lisp
@@ -246,7 +246,7 @@
   ;;
 
   (define-class (Into :a :b)
-    "INTO imples *every* element of `:a` can be represented by an element of `:b`. This conversion might not be bijective (i.e., there may be elements in `:b` that don't correspond to any in `:a`)."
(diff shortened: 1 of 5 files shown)


library/classes.lisp
```diff
@@ -255,11 +255,9 @@
   (define-instance (Into :a :a)
     (define (into x) x))
 
-  (define-class (TryInto :a :b)
-    "TRY-INTO implies some elements of `:a` can be represented exactly by an element of :b, but sometimes not. If not, an error string is returned."
-    ;; Ideally we'd have an associated-type here instead of locking in
-    ;; on String.
-    (tryInto (:a -> Result String :b)))
+  (define-class (TryInto :a :b :c)
```
eliaslfox, line 258: `TryInto :a :b :c (:a :b -> :c)` see [functional dependencies](https://wiki.haskell.org/Functional_dependencies).

Izaakwltn: I just pushed what I believe is more compliant.

eliaslfox (approved):

PR #1102, opened 2024-05-06
Adds `Signalable` class, for signalling errors

Izaakwltn:
Currently the function `error` only takes arguments of type `String`. 

This PR allows you to define error behavior for any type without first converting each object to a string first. 

This also allows for error behavior to be specified within one instance definition rather than making scattered formatting functions or repetitive boilerplate lisp format strings. 

**Note** This doesn't break any existing `error` uses, since there is an instance defined for `String`. 

Eventually `warn` might also be useful but I think it's worth thinking a little more about.

Here are some examples:

```
(defpackage #:signalable-demo
  (:use #:coalton
        #:coalton-prelude)
  (:local-nicknames (#:math #:coalton-library/math)
                    (#:result #:coalton-library/result)))

(in-package #:signalable-demo)

(coalton-toplevel

  (define-type InvalidIntError
    (OddError Integer)
    (EvenError Integer))

  (define-instance (Signalable InvalidIntError)
    (define (error x)
      (match x
        ((OddError i)
         (error (lisp String (i)
                  (cl:format cl:nil "Invalid Integer: ~a is Odd." i))))
        ((EvenError i)
         (error (lisp String (i)
                  (cl:format cl:nil "Invalid Integer: ~a is Even." i))))))))

;; Some working use-cases:
(coalton-toplevel

  (define (only-if-even x)
    (if (math:zero? (mod x 2))
        x
        (error (OddError x))))

  (define (evn? x)
    (if (math:zero? (mod x 2))
        (Ok x)
        (Err (OddError x))))
  
  (define (error-if-odd x)
    (result:unwrap-or-error (evn? x))))
```

### DIFF
--- a/library/classes.lisp
+++ b/library/classes.lisp
@@ -3,13 +3,14 @@
    #:coalton)
   (:local-nicknames
    (#:types #:coalton-library/types))
-
   (:export
    #:Addressable #:eq?)
   (:export
    #:Eq #:==
    #:Num #:+ #:- #:* #:fromInt)
   (:export
+   #:Signalable
+   #:error
    #:Tuple
    #:Optional #:Some #:None
    #:Result #:Ok #:Err
@@ -44,6 +45,25 @@
 
 #+coalton-release
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
+
+;;;
+;;; Signaling errors and warnings
+;;;
+
+(coalton-toplevel
+ 
+ ;;
+ ;; Signalling errors on supported types
+ ;;
+ (define-class (Signalable :a)
+   "Signals errors or warnings by calling their respective lisp conditions."
+   (error (:a -> :b)))
+
+  (define-instance (Signalable String)
+    (define (error str)
+      (lisp :a (str)
+        (cl:error str)))))
+
 
 (coalton-toplevel
 
@@ -264,11 +284,6 @@
   ;;
   ;; Unwrappable for fallible unboxing
   ;;
-
-  (declare error (String -> :a))
-  (define (error str)
-    "Signal an error by calling `CL:ERROR`."
-    (lisp :a (str) (cl:error str)))
 
   (define-class (Unwrappable :container)
     "Containers which can be unwrapped to get access to their contents.
--- a/library/result.lisp
+++ b/library/result.lisp
@@ -11,7 +11,8 @@
    #:ok?
    #:err?
    #:map-err
-   #:flatten))
+   #:flatten
+   #:ok-or-error))
 
 (in-package #:coalton-library/result)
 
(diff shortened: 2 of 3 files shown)


library/classes.lisp
```diff
@@ -3,13 +3,16 @@
...
   (:local-nicknames
    (#:types #:coalton-library/types))
-
   (:export
    #:Addressable #:eq?)
   (:export
    #:Eq #:==
    #:Num #:+ #:- #:* #:fromInt)
   (:export
+   #:coalton-warning
```
eliaslfox, line 12: unused

library/classes.lisp
```diff
@@ -3,13 +3,16 @@
...
+   #:Signalable
+   #:error
+   #:warn
```
eliaslfox, line 15: unused

library/result.lisp
```diff
@@ -52,6 +54,14 @@
       ((Ok x) x)
       ((Err x) x)))
 
+  (declare unwrap-or-error (Signalable :a => (Result :a :b) -> :b))
```
eliaslfox, line 57: I think this should be defined on the `Unwrappable` class.

Izaakwltn: I'll keep thinking about this, but my initial thought is that this needs to be just for `Result` because other `Unwrappable` instances (currently only `Optional`) don't necessarily carry an error/fail type to be declared `Signalable`. 

Using `unwrap` on an `Optional` already provides a default error string, which is all that an ideal `unwrap-or-error` would provide anyways. 

In order to avoid confusion it could be renamed something like `ok-or-error`.

Izaakwltn: I fixed the package definitions and changed to `ok-or-error`.

library/classes.lisp
```diff
@@ -3,13 +3,15 @@
    #:coalton)
   (:local-nicknames
    (#:types #:coalton-library/types))
-
+
```
eliaslfox, line 6: extra space

library/classes.lisp
```diff
@@ -45,6 +47,22 @@
 #+coalton-release
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
 
+;;;
+;;; Signaling errors and warnings
+;;;
+
+(coalton-toplevel
```
eliaslfox, line 54: this can go in the below coalton-toplevel

library/classes.lisp
```diff
@@ -319,7 +332,8 @@ Typical `fail` continuations are:
                     (fn () default)
                     container))
 
-  (declare as-optional ((Unwrappable :container) => (:container :elt) -> (Optional :elt)))
+  (declare as-optional ((Unwrappable :container) => (:container :elt) -> (Optional :elt)
+                        ))
```
eliaslfox, line 336: formatting

eliaslfox (approved):

PR #1104, opened 2024-05-08
Initial support for compilation of .coalton files

jbouwman:
Define a backend protocol to support multiple output schemes. Instead of collecting and returning code and environment from the compiler entry point, and consing an environment update log, provide functions for incremental output (emit, emit-env), and return only the functionally-updated environment, which caller can keep or discard.

Deduplicate set-code environment updates, which reduces FASL size a bit, about 10% on ARM, and a little more on AMD64.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -98,7 +98,8 @@
                              (:file "package")))
                (:module "codegen"
                 :serial t
-                :components ((:file "pattern")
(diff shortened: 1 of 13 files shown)


jbouwman:
Portions to reappear in a new PR after #1091 is complete

PR #1107, opened 2024-05-13
Uses iterator operations to fix (Into Vector List) stack overflow

### LINKED ISSUE #1106: `Into (Vector :a) (List :a)` triggers stack overflows at large vector sizes
```
(coalton-toplevel

  ;; stack overflows
  (define (numvec->numlist)
    (let v = (vec:new))
    (iter:for-each! (fn (x)
                      (vec:push! x v)
                      Unit)
                    (iter:up-to 10000))
    (The (List Integer) (into v)))

  ;; potential alternative, does not overflow
  (define (numvec->numlist2)
    (let v = (vec:new))
    (iter:for-each! (fn (x)
                      (vec:push! x v)
                      Unit)
                    (iter:up-to 10000))
    (The (List Integer) (iter:collect! (iter:into-iter v)))))
```

Izaakwltn:
Fixes #1106 

I'm not sure this is the most eloquent solution but it's quick and works.

### DIFF
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -318,12 +318,7 @@
 
   (define-instance (Into (Vector :a) (List :a))
     (define (into v)
-      (let ((inner
-              (fn (v index)
-                (if (>= index (length v))
-                    Nil
-                    (Cons (index-unsafe index v) (inner v (+ 1 index)))))))
-        (inner v 0))))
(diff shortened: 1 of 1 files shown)


Izaakwltn:
Maybe instead it should be a `coerce` action

PR #1109, opened 2024-05-17
ASDF component type for .coalton

jbouwman:
Support for .coalton files.

Behavior identical to .lisp, for now, but provides a subclass of asdf:source-file that can call a Coalton compiler entry point.


coalton-asdf.lisp
```diff
@@ -0,0 +1,28 @@
...
+;;       :components ((:coalton-file "types")
+;;                    (:coalton-file "main")))
+;;
+;; This class is defined in the ASDF package, so that the keyword :coalton-file works.
+;;
+;; see: https://github.com/fare/asdf/blob/master/doc/best_practices.md#using-asdf-extensions
+
+(in-package :asdf)
+
+(defclass coalton-file (cl-source-file)
```
stylewarning, line 24: is it kosher to define new things in ASDF? I think we should make a new package.

coalton-asdf.lisp
```diff
@@ -0,0 +1,28 @@
...
+  ((type :initform "coalton")))
+
+(defmethod source-file-type ((c coalton-file) (s module))
+  "coalton")
```
stylewarning, line 28: any objection to using `coal` :D ?

stylewarning (changes requested):

coalton-asdf.lisp
jbouwman: asdf documentation suggests that this is a common practice, in: https://github.com/fare/asdf/blob/master/doc/best_practices.md#using-asdf-extensions 

> If you need an ASDF extension, the recommended way is to use :defsystem-depends-on. The extension will define new classes of operations or components, new functions, etc. If it defines them in the ASDF package, you can refer to them using a keyword of the same name, e.g. :cffi-wrapper-file for one of the component classes defined by system cffi-grovel, :static-program-op for one of the operation classes defined by cffi-toolchain, or :f2cl-system for the system class defined by system f2cl.

stylewarning: absolutely goofy but that's what they recommend so that's what they recommend

coalton-asdf.lisp
stylewarning: i like ghibli resonance

jbouwman: Changed to use .coal.

Parser test files had been using .coalton -- renamed those ,and took the opportunity to place good and bad inputs in separate directories.

stylewarning (approved):

PR #1110, opened 2024-05-21
Simplified coalton-file asdf presence

Izaakwltn:
This simplifies the recent `coalton-file` asdf feature by including it in `coalton.asd`.

I don't think there is really any reason to keep this feature quarantined. This change ensures that coalton users will be able to load `.coal` files without having to load an additional dependency, and this also keeps the repo a little sparser. 

Much of the standard library would likely depend on this eventually, so it would need to be a dependency anyways.


Izaakwltn:
Actually there is an issue with this:

If you load an external asd that depends on coalton and has `.coal` types, it will error from knowing :coalton-file before loading the #:coalton dependency that would fix that. 

Loading coalton and then loading the dependent system works though.

Izaakwltn:
> Actually there is an issue with this:
> 
> If you load an external asd that depends on coalton and has `.coal` types, it will error from not knowing `:coalton-file` before loading the #:coalton dependency that would fix that.
> 
> Loading coalton and then loading the dependent system works though.

Nevermind, I just saw the note in coalton-asdf.lisp: `:defsystem-depends-on` has to be added to any system with `.coal` files. 

So, actually, this PR would change the working model to:
```
(asdf:defsystem #:my-coalton-project
  :defsystem-depends-on (#:coalton)
  :serial t
  :components ((:coalton-file "my-coalton-project")))
```
which would currently be:
```
(asdf:defsystem #:my-coalton-project
  :defsystem-depends-on (#:coalton-asdf)
  :depends-on (#:coalton)
  :serial t
  :components ((:coalton-file "my-coalton-project")))
```

eliaslfox (approved):

stylewarning:
I thought ASDF files weren't supposed to have "code" in them.

eliaslfox:
We do it already [here](https://github.com/coalton-lang/coalton/blob/main/coalton.asd#L204). If this was enough code to actually call the compiler, then I agree it should go in another file.

stylewarning:
> We do it already [here](https://github.com/coalton-lang/coalton/blob/main/coalton.asd#L204). If this was enough code to actually call the compiler, then I agree it should go in another file.

I'm not opposed to it, I just wasn't sure if anybody knew the recommendation off-hand. Trying to imagine weird lisp compilers like to-C compilers and how they work with stuff like this.

Izaakwltn:
It will probably need to go in a separate file once an actual compilation method is worked out...

PR #1111, opened 2024-05-30
direct compilation of .coal files

jbouwman:
Compiler entry points for .coal -> .lisp and .coal -> .fasl

### DIFF
--- a/src/entry.lisp
+++ b/src/entry.lisp
@@ -13,7 +13,8 @@
    #:*global-environment*
    #:entry-point                        ; FUNCTION
    #:expression-entry-point             ; FUNCTION
-   #:file-entry-point                   ; FUNCTION
+   #:compile-coalton                    ; FUNCTION
+   #:compile-coalton-toplevel           ; FUNCTION
    ))
 
 (in-package #:coalton-impl/entry)
@@ -29,9 +30,7 @@
 
          (file (parser:program-file program))
 
-         (env *global-environment*)
-
-         (tc:*env-update-log* nil))
+         (env *global-environment*))
 
     (multiple-value-bind (type-definitions instances env)
         (tc:toplevel-define-type (parser:program-types program)
@@ -85,28 +84,8 @@
 
                 (analysis:analyze-translation-unit translation-unit env file)
 
-                (multiple-value-bind (program env)
-                    (codegen:compile-translation-unit translation-unit monomorphize-table env)
+                (codegen:compile-translation-unit translation-unit monomorphize-table env)))))))))
 
-                  (values
-                   (if settings:*coalton-skip-update*
-                       program
-                       `(progn
-                          (eval-when (:load-toplevel)
-                            (unless (eq (settings:coalton-release-p) ,(settings:coalton-release-p))
-                              ,(if (settings:coalton-release-p)
-                                   `(error "~A was compiled in release mode but loaded in development."
-                                           ,(or *compile-file-pathname* *load-truename*))
-                                   `(error "~A was compiled in development mode but loaded in release."
-                                           ,(or *compile-file-pathname* *load-truename*)))))
-
-                          (let ((coalton-impl/typechecker/environment::env *global-environment*))
-                            ,@(loop :for elem :in (reverse tc:*env-update-log*)
-                                    :collect elem)
-                            (setf *global-environment* coalton-impl/typechecker/environment::env))
-
-                          ,program))
-                   env))))))))))
 
 (defun expression-entry-point (node file)
   (declare (type parser:node node)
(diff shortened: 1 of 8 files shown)


src/compiler.lisp
```diff
@@ -0,0 +1,660 @@
...
+
+;;; The code below defines a format for environment updates that can
+;;; be written as Common Lisp source. When Coalton code is embedded in
+;;; Lisp files using the coalton-toplevel macro, the arguments to the
+;;; environment mutators in the typechecker/environment package appear
+;;; in the macroexpansion as externalizable structures, in that the
+;;; make-load-form generic is defined for possible values, using
+;;; make-load-form-saving-slots.
+;;;
+;;; Producing Lisp source from Coalton requires a source-representable
```
eliaslfox, line 28: Isn't make-load-form source representable? You can write `#S(...)` in a regular lisp file.

jbouwman: Is there a better way? A simple example, taking a kfun instance,

```
(defvar s (tc:make-kfun :from tc:+kstar+ :to tc:+kstar+))
```

is there a way to produce a (reasonably platform independent) source representation from 's'? make-load-form returns an allocator and initializer:

```
(multiple-value-list (make-load-form-saving-slots s))
  ==>
((SB-KERNEL::ALLOCATE-STRUCT 'COALTON-IMPL/TYPECHECKER/KINDS:KFUN)
 (PROGN
  (SETF (SB-KERNEL:%INSTANCE-REF * → * 1) '*)
  (SETF (SB-KERNEL:%INSTANCE-REF * → * 2) '*)))
```

Could you elaborate a little on how you're suggesting to use #s syntax?

eliaslfox: Why does it need to be platform independent?

jbouwman: It doesn't have to be.

eliaslfox: In that case can we continue using make-load-form instead of defining bespoke serialization logic?

src/entry.lisp
```diff
@@ -13,7 +13,8 @@
    #:*global-environment*
    #:entry-point                        ; FUNCTION
    #:expression-entry-point             ; FUNCTION
-   #:file-entry-point                   ; FUNCTION
+   #:compile-coalton                    ; FUNCTION
```
stylewarning, line 16: nit: maybe consider renaming these functions?

tests/entry-tests.lisp
```diff
@@ -0,0 +1,50 @@
+(in-package #:coalton-tests)
+
+(defun compile-test-file ()
+  (merge-pathnames "examples/small-coalton-programs/src/classical.coal"
+                   (asdf:system-source-directory "coalton")))
+
+(deftest test-compile-to-lisp ()
+  "Test that the Coalton compiler compiles a test file into Lisp source."
+  (let ((source (entry:compile-coalton (compile-test-file) :format :source)))
```
stylewarning, line 9: nit: quote keyword

src/reader.lisp
```diff
@@ -16,6 +16,36 @@
   "Is the Coalton reader allowed to parse the current input?
 Used to forbid reading while inside quasiquoted forms.")
 
+(defmacro with-coalton-file ((file stream) &body body)
+  "Stream and error bookkeeping helper for coalton macros that binds FILE to a coalton-file instance."
+  `(let ((opened-streams nil))
```
stylewarning, line 21: some little bits of gensymming and the like

src/reader.lisp
```diff
@@ -33,142 +63,24 @@ Used to forbid reading while inside quasiquoted forms.")
...
-                            (error:render-coalton-warning c))))
-                     (multiple-value-bind (program env)
-                         (entry:entry-point (parser:read-program stream file :mode :toplevel-macro))
-                       (setf entry:*global-environment* env)
-                       program)))
-              ;; Clean up any opened file streams
-              (dolist (s opened-streams)
-                (close s)))))
+          (with-coalton-file (file stream)
+            (entry:compile-coalton-toplevel (parser:read-program stream file :mode :toplevel-macro))))
```
stylewarning, line 67: nit: quote kw

stylewarning (changes requested):
just a couple nits

src/entry.lisp
jbouwman: I agree these names are not quite right. I think there will be better focus when the calling patterns from swank and emacs have been run through. Ok to leave as is (for the moment)?

stylewarning: yes

src/reader.lisp
```diff
@@ -16,6 +16,39 @@
   "Is the Coalton reader allowed to parse the current input?
 Used to forbid reading while inside quasiquoted forms.")
 
+(defmacro with-coalton-file ((file stream) &body body)
+  "Ensure that in case of error, error text is rendered before stream is closed; stream is then closed; also that FILE is bound to an instance of coalton-impl/error:coalton-file, which is expected the parser functions."
```
stylewarning, line 20: Can we write this in a more direct style? Something like

> Given a stream STREAM, bind a COALTON-IMPL/ERROR:COALTON-FILE instance to the symbol FILE and execute BODY. In the event of a Coalton compiler error or warning, print it.

src/reader.lisp
```diff
@@ -16,6 +16,39 @@
...
+        (pathname (gensym))
+        (filename (gensym))
+        (file-input-stream (gensym)))
+    `(let ((,opened-streams nil))
+       (unwind-protect
+            (let* ((,pathname (or *compile-file-truename* *load-truename*))
+                   (,filename (if ,pathname (namestring ,pathname) "<unknown>"))
+                   (,file-input-stream
+                     (cond
+                       ((or #+sbcl (sb-int:form-tracking-stream-p ,stream)
```
stylewarning, line 31: This deserves a comment, even though it was from old code.

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+    (with-environment-updates updates
+      (let* ((file (error:make-coalton-file :stream input
+                                            :name name))
+             (program (entry-point (parser:read-program input file :mode :file))))
+        (print-form output (make-prologue))
+        (print-form output (make-environment-updater updates))
+        (print-form output program)))))
+
+(defun compile-coalton (coal-file &key (output-file nil) (format :default))
+  "Compile a Coalton file. The FORMAT keyword argument controls how the output is generated:)
```
stylewarning, line 239: Delete paren at the end

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+
+:default Generate Lisp source, and compile immediately to .fasl. If no output file is specified, compiled Lisp code will be written to a temporary file, and that path will be returned.
```
stylewarning, line 241: Indent and capitalize the keywords.

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+                                    :suffix "lisp"
+                                    :direction :output
+                                    :keep t)
+           (compile-to-lisp coal-file-name coal-stream lisp-stream)
+           :close-stream
+           (compile-file lisp-file :output-file output-file)))
+        (:source
+         (cond (output-file
+                (with-open-file (lisp-stream output-file
+                                             :direction :output
```
stylewarning, line 262: nit: quote keywords

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+                               :element-type 'character)
+    (let ((coal-file-name (etypecase coal-file
+                            (pathname (pathname-name coal-file))
+                            (string coal-file))))
+      (ecase format
+        (:default
+         (uiop:with-temporary-file (:stream lisp-stream
+                                    :pathname lisp-file
+                                    :suffix "lisp"
+                                    :direction :output
```
stylewarning, line 254: nit: quote keyword

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+                                    :keep t)
+           (compile-to-lisp coal-file-name coal-stream lisp-stream)
+           :close-stream
+           (compile-file lisp-file :output-file output-file)))
+        (:source
+         (cond (output-file
```
stylewarning, line 260: nit: newline after cond

src/entry.lisp
```diff
@@ -194,23 +173,96 @@
...
+     (unless (eq (settings:coalton-release-p) ,(settings:coalton-release-p))
+       ,(if (settings:coalton-release-p)
+            `(error "~A was compiled in release mode but loaded in development."
+                    ,(or *compile-file-pathname* *load-truename*))
+            `(error "~A was compiled in development mode but loaded in release."
+                    ,(or *compile-file-pathname* *load-truename*))))))
+
+(defun print-form (stream form)
+  "Print a FORM to a STREAM, separated by 2 lines."
+  (let ((*package* (find-package :cl)))
```
stylewarning, line 221: `*package* is going to get reset to `cl-user` with `with-standard-io-syntax`. I would instead do:

```
(with-standard-io-syntax
  (let ((*package* (find-package "KEYWORD")))
    (prin1 ...) ...))
```

stylewarning (changes requested):

eliaslfox:
This pr should include asdf integration for Coalton files. That way it'll be much easier to check if fasls can be reloaded after restarting sbcl.

stylewarning:
> This pr should include asdf integration for Coalton files. That way it'll be much easier to check if fasls can be reloaded after restarting sbcl.

We definitely need the proper ASDF integration, but I am going to petition that we prioritize that in a follow-up PR, just because there are a couple dependent PRs waiting for this one.

src/reader.lisp
```diff
@@ -33,142 +70,24 @@ Used to forbid reading while inside quasiquoted forms.")
...
-                       (multiple-value-bind (program env)
-                           (entry:entry-point (parser:read-program stream file :mode :toplevel-macro))
-                         (declare (ignore env))
-                         `',program))))
-              ;; Clean up any opened file streams
-              (dolist (s opened-streams)
-                (close s)))))
+          (with-coalton-file (file stream)
+            (let ((settings:*emit-type-annotations* nil))
+              `',(entry:entry-point (parser:read-program stream file :mode :toplevel-macro)))))
```
stylewarning, line 79: nit: quote

src/reader.lisp
```diff
@@ -33,142 +70,24 @@ Used to forbid reading while inside quasiquoted forms.")
...
-                           (entry:entry-point (parser:read-program stream file :mode :toplevel-macro))
-                         (declare (ignore program env))
-                         nil))))
-              ;; Clean up any opened file streams
-              (dolist (s opened-streams)
-                (close s)))))
+          (with-coalton-file (file stream)
+            (let ((settings:*emit-type-annotations* nil)
+                  (settings:*coalton-dump-ast* t))
+              (entry:entry-point (parser:read-program stream file :mode :toplevel-macro))
```
stylewarning, line 85: nit: quote

stylewarning (approved):

Issue #1073, opened 2024-03-22
In repl `coalton:Coalton` macro: Single-Float automatically converts to Double-Float


Izaakwltn:
When using `coalton:coalton` in the repl, single floats are automatically converted to Double-Floats.

This behavior doesn't happen with `coalton-codegen` or `coalton-toplevel`.

Most likely a reader error.

```
COALTON-USER> (coalton 1.0d0)
1.0d0
COALTON-USER> (coalton 1.0)
1.0d0
COALTON-USER> (coalton 1.0f0)
1.0d0
COALTON-USER> (coalton-toplevel (define x 1.0f0)) ;; this works correctly
; No value
COALTON-USER> (type-of 'x)
SINGLE-FLOAT
COALTON-USER> cl:*read-default-float-format*
COMMON-LISP:SINGLE-FLOAT
```

Here is the behavior after changing `cl:*read-default-float-format*` to other floating-point types:
```
;; setting to long-float
COALTON-USER> (cl:setq  cl:*read-default-float-format* 'cl:long-float)
COMMON-LISP:LONG-FLOAT
COALTON-USER> (coalton:coalton 1.0f0)
1.0f0
COALTON-USER> (coalton:coalton 1.0d0)
1.0
COALTON-USER> (coalton:coalton 1.0l0)
1.0
;; setting to double-float
COALTON-USER> (cl:setq  cl:*read-default-float-format* 'cl:double-float)
COMMON-LISP:DOUBLE-FLOAT
COALTON-USER> (coalton:coalton 1.0f0)
1.0f0
COALTON-USER> (coalton:coalton 1.0d0)
1.0
COALTON-USER> (coalton:coalton 1.0l0)
1.0
```

(development mode on SBCL)

Izaakwltn:
One other case:

```
COALTON-USER> (coalton (the single-float 1.0))
```

```
error: Type mismatch
  --> COALTON (NIL):1:9
   |
 1 |  (COALTON (THE SINGLE-FLOAT 1.0))
   |           ^^^^^^^^^^^^^^^^^^^^^^ Declared type 'SINGLE-FLOAT' does not match inferred type 'DOUBLE-FLOAT'
   [Condition of type COALTON-IMPL/TYPECHECKER/BASE:TC-ERROR]
```

jbouwman:
Fixed by #1113

PR #1113, opened 2024-06-03
don't enforce a default float format

jbouwman:
Does this address #1073?

Eclector reader setup is resetting \*read-default-float-format\*.


stylewarning:
I think this is the right call, even though Lisp has an annoying default.

eliaslfox (approved):

PR #1114, opened 2024-06-03
Defend inner reader

### LINKED ISSUE #1053: `A token consisting solely of multiple dots is illegal.` when `*print-length*` is a small integer
The main branch (27574d1fae0e3b1b3390acba2e8becccd1f7b4cb) produces multiple errors on SBCL 2.3.11 saying `A token consisting solely of multiple dots is illegal.` when `cl:*print-length*` is set to non-NIL, say 10:

```
; in: DEFINE-INTEGER-CONVERSIONS U8
;     (COALTON-LIBRARY/MATH/CONVERSIONS::DEFINE-INTEGER-CONVERSIONS COALTON:U8)
;
; caught COMMON-LISP:ERROR:
;   (during macroexpansion of (COALTON-TOPLEVEL
;     (DEFINE-INSTANCE #
;       ...)
;     ...))
;   A token consisting solely of multiple dots is illegal.

; in: DEFINE-INTEGER-CONVERSIONS I8
;     (COALTON-LIBRARY/MATH/CONVERSIONS::DEFINE-INTEGER-CONVERSIONS COALTON:I8)
;
; caught COMMON-LISP:ERROR:
;   (during macroexpansion of (COALTON-TOPLEVEL
;     (DEFINE-INSTANCE #
;       ...)
;     ...))
;   A token consisting solely of multiple dots is illegal.
.
.
.
```

jbouwman:
Prevent truncation of the form passed to eclector by user-level print length and level settings

Fixes #1053

### DIFF
--- a/src/reader.lisp
+++ b/src/reader.lisp
@@ -240,34 +240,37 @@ Used to forbid reading while inside quasiquoted forms.")
                      (let ((*coalton-reader-allowed* t))
                        (funcall (get-macro-character #\, (named-readtables:ensure-readtable :standard)) s c)))))
 
+(defun print-form (form)
+  "Prevent truncation of a FORM that will be immediately re-read."
+  (let ((*print-length* nil)
+        (*print-level* nil)
+        (*print-circle* t))
+    (prin1-to-string form)))
+
 (defmacro coalton:coalton-toplevel (&body forms)
   (let ((*readtable* (named-readtables:ensure-readtable 'coalton:coalton))
         (*compile-file-truename*
-          (pathname (format nil "COALTON-TOPLEVEL (~A)" *compile-file-truename*)))
-        (*print-circle* t))
-    (with-input-from-string (stream (cl:format cl:nil "~S" (cons 'coalton:coalton-toplevel forms)))
+          (pathname (format nil "COALTON-TOPLEVEL (~A)" *compile-file-truename*))))
+    (with-input-from-string (stream (print-form (cons 'coalton:coalton-toplevel forms)))
       (cl:read stream))))
 
 (defmacro coalton:coalton-codegen (&body forms)
   (let ((*readtable* (named-readtables:ensure-readtable 'coalton:coalton))
         (*compile-file-truename*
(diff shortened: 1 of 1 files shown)


src/reader.lisp
```diff
@@ -240,34 +240,37 @@ Used to forbid reading while inside quasiquoted forms.")
                      (let ((*coalton-reader-allowed* t))
                        (funcall (get-macro-character #\, (named-readtables:ensure-readtable :standard)) s c)))))
 
+(defun print-form (form)
+  "Prevent truncation of a FORM that will be immediately re-read."
+  (let ((*print-length* nil)
+        (*print-level* nil)
+        (*print-circle* t))
```
eliaslfox, line 247: I *think* setting `*default-float-format*` to double-float will fix #1073.

stylewarning:
Maybe [`with-standard-io-syntax`](http://www.ai.mit.edu/projects/iiip/doc/CommonLISP/HyperSpec/Body/mac_with-stan_rd-io-syntax.html)?

jbouwman:
oh, whoops, no -- with-standard-io-syntax blows away the readtable :|

jbouwman:
> oh, whoops, no -- with-standard-io-syntax blows away the readtable :|
> 
> edit: I'm not sure why it doesn't work; will investigate

with-standard-io-syntax sets *package* to :cl-user, which breaks the contract that the coalton-toplevel form's contents are written and then reread in whatever package is current. For instance, violating this loses the 'cl:gcd' etc. qualified symbols in library/math/integral.lisp and overrides them with their Coalton versions.

print-readably t borks -> syntax

print-circle nil breaks a weak test in tests/reader-tests

so the following would be ok:

```
  (let ((p *package*))
    (with-standard-io-syntax
      (let ((*package* p)
            (*print-circle* t)
            (*print-readably* nil))
        (prin1-to-string form)))))
```

src/reader.lisp
jbouwman: It doesn't.

eliaslfox (approved):

stylewarning:
@jbouwman Do you know why print-readably doesn't work with ->?

PR #1116, opened 2024-06-05
Configurable AST collection

jbouwman:
Instead of writing AST info to stdout, code generator provides values to an output hook function under control of caller, which can decide what to do with the values.

This is to support returning these structures to Emacs via swank.

### DIFF
--- a/src/codegen/program.lisp
+++ b/src/codegen/program.lisp
@@ -27,9 +27,19 @@
    (#:rt #:coalton-impl/runtime)
    (#:tc #:coalton-impl/typechecker))
   (:export
+   #:*codegen-hook*
    #:compile-translation-unit))
 
 (in-package #:coalton-impl/codegen/program)
+
+(defvar *codegen-hook* nil
+  "A hook that provides access to intermediate program representations during code generation.
+
+A function bound here will be called with a keyword category, and one or more additional arguments, depending on the value of that keyword. Keyword values:
+
+  :ast name type value
+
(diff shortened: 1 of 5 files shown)


src/codegen/program.lisp
```diff
@@ -27,10 +27,13 @@
    (#:rt #:coalton-impl/runtime)
    (#:tc #:coalton-impl/typechecker))
   (:export
+   #:*codegen-hook*
    #:compile-translation-unit))
 
 (in-package #:coalton-impl/codegen/program)
 
+(defvar *codegen-hook*)
```
Izaakwltn, line 35: It might be nice to have a docstring or comment for this variable.

Izaakwltn (approved):
One nit, otherwise looks good to me

src/codegen/program.lisp
jbouwman: completely agree, added it, thank you

stylewarning:
Was `get-ast` intended to be a part of this PR? I'd have raised two minor points:

1. Keywords aren't quoted, which is extra confusing because they're passed to a lambda function.

2. `get-ast` isn't an optimal name for a function that reads from a stream.

CC @Izaakwltn

jbouwman:
`get-ast` is tracking against a swank function in coalton-emacs -- it's clear that it should move there, since the basic support is now present to implement it. I'll open a ticket to do that (and quote the keyword)

Issue #1126, opened 2024-06-11
Monomorphizer tests are excluded from the test suite, is this intentional?


Izaakwltn:
Currently, they are not loaded as part of `:coalton/tests`. 

I checked and they compile, run, and pass. 

https://github.com/coalton-lang/coalton/blob/05111b8a59e3f7346b175ce1ec621bff588e1e1f/tests/monomorphizer-tests.lisp

stylewarning:
Maybe an accident. Let's ship them if they work.

stylewarning:
(But let's also verify the tests make sense.)

Izaakwltn:
Sounds good, I'll take a look

Izaakwltn:
Closed by #1128

PR #1128, opened 2024-06-14
Fix monomorphizer underapplication

kartik-s:
The monomorphizer fails when compiling a `COMPILE-CANDIDATE` that meets the following criteria:

- The `NAME` refers to a method.
- The method is underapplied.
- The `ARGS` are only statically known dictionaries and nothing else.
- The method is not inline for the argument dictionaries.

Such a `COMPILE-CANDIDATE` could be found, for example, when `COALTON-LIBRARY/LIST:MAXIMUM` is applied to a `(List Big-Float)` because `COALTON-LIBRARY/LIST:MAXIMUM` implicitly underapplies `>` to `Big-Float`'s instance of `Ord`, which does not inline the `>` method.

This PR fixes this case in the monomorphizer, adds a test for it, and adds the monomorphizer tests to the Coalton test suite. 

`cl-quil/discrete` now builds and passes its tests with [this `(monomorphize)`](https://github.com/quil-lang/quilc/blob/c181970798e398fade1d2b374e7f4d5806d0e897/src/discrete/rz-approx/generate-solution.lisp#L151) added back.

### DIFF
--- a/tests/monomorphizer-tests.lisp
+++ b/tests/monomorphizer-tests.lisp
@@ -31,6 +31,19 @@
 
 
 (coalton-toplevel
+  (monomorphize)
+  (declare underapplication-c
+           ((List coalton-library/big-float:Big-Float) -> coalton-library/big-float:Big-Float))
+  (define (underapplication-c xs)
+    (match (list:maximum (append xs (map (fn (x) (* x x)) xs)))
+      ((Some x) x)
+      ((None) 0))))
+
+(define-test test-monomorphizer-under-application-non-inline-method ()
+  (is (== 9 (underapplication-c (make-list -1 -2 -3)))))
+
+
+(coalton-toplevel
   (define (partial-monomorphization-a x y z)
     (== x y)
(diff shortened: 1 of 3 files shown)


stylewarning (approved):
lgtm

Izaakwltn:
Closes #1126

tests/monomorphizer-tests.lisp
```diff
@@ -30,6 +30,19 @@
   (is (== 4 (underapplication-b 2))))
 
 
+(coalton-toplevel
+  (monomorphize)
+  (declare underapplication-c
+           ((List coalton-library/big-float:Big-Float) -> coalton-library/big-float:Big-Float))
+  (define (underapplication-c xs)
+    (match (coalton-library/list:maximum (append xs (map (fn (x) (* x x)) xs)))
```
eliaslfox, line 38: `list:maximum`

eliaslfox (approved):

PR #1129, opened 2024-06-18
lisp-toplevel in coalton files

jbouwman:
Add lisp-toplevel form.

`lisp-toplevel` may appear at any point in a coalton-toplevel or coalton file context.

During code generation, lisp forms are merged with coalton definitions according to the rule:

1. coalton forms are emitted in scc order
2. every time a coalton form is emitted, all lisp forms lexically preceding that form are emitted first

Convert math/num.lisp from interleaved lisp and coalton-toplevel forms to a single coalton-toplevel form containing multiple lisp-toplevel forms.

This will support direct conversion to .coal by removing the outer coalton-toplevel macro and adding a native package form.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -27,7 +27,12 @@
   :version (:read-file-form "VERSION.txt")
   :around-compile (lambda (compile)
                     (let (#+sbcl (sb-ext:*derive-function-types* t)
-                          #+sbcl (sb-ext:*block-compile-default* :specified))
+                          #+sbcl (sb-ext:*block-compile-default* :specified)
+                          ;; The lisp-toplevel form is currently
+                          ;; restricted to standard library
+                          ;; implementation by checking for the
+                          ;; presence of this feature.
+                          (*features* (cons ':coalton-lisp-toplevel *features*)))
                       (funcall compile)))
   :defsystem-depends-on (#:coalton-asdf)
   :depends-on (#:coalton-compiler
--- a/src/codegen/program.lisp
+++ b/src/codegen/program.lisp
@@ -23,6 +23,7 @@
    #:optimize-bindings)
   (:local-nicknames
    (#:util #:coalton-impl/util)
+   (#:parser #:coalton-impl/parser)
    (#:settings #:coalton-impl/settings)
    (#:global-lexical #:coalton-impl/global-lexical)
    (#:rt #:coalton-impl/runtime)
@@ -42,41 +43,116 @@ A function bound here will be called with a keyword category, and one or more ad
 
     Toplevel definitions, after type checking and before compilation.")
 
+;; The following functions control the output order of compiled
+;; definitions and interleaved lisp expressions.
+;;
+;; Toplevel define and instance forms are compiled to 1 or more named,
+;; lisp-source-valued output definitions: when these definitions are
+;; generated, they are associated with the starting source offset of their
+;; toplevel form:
+;;
+;;   toplevel definition:
+;;     #<def .... (300 . 345)>
+;;
+;;   bindings (lisp definitions):
+;;     (300 b1 .. bn)
+;;
+;; Then when compile-definitions emits the full set of output
+;; definitions, any lisp source forms that occurred earlier in the
+;; file are emitted first.
+
+(defun bindings-offset (bindings offsets)
+  "Given a list of binding names, and a name -> offset map, return the earliest binding start offset."
+  (reduce #'min
+          (mapcar (lambda (binding)
+                    (gethash (car binding) offsets 0))
+                  bindings)))
+
+(defun merge-forms (forms-a forms-b &optional merged)
+  "Stably merge two lists of forms.
+
+1. The inputs and output are lists of 2-lists, structured as (OFFSET FORM).
+2. The order of both lists is preserved.
+3. Lists are merged by recursively selecting the head of the list with the lowest offset.
+
+Example:
+
+  (merge-forms '((2 \"b\") (8 \"d\") (1 \"a\"))
+               '((5 \"x\") (7 \"x\")))
+
+    => ((2 \"b\") (5 \"x\") (7 \"x\") (8 \"d\") (1 \"a\"))
+
+(Note that the order of elements in the first list is preserved)"
+  (cond ((endp forms-a)
+         (return-from merge-forms
+           (nreconc merged forms-b)))
+        ((endp forms-b)
+         (return-from merge-forms
+           (nreconc merged forms-a)))
+        ((< (caar forms-a)
+            (caar forms-b))
+         (push (pop forms-a) merged))
+        (t
+         (push (pop forms-b) merged)))
+  (merge-forms forms-a forms-b merged))
+
+(defun compile-definitions (sccs definitions lisp-forms offsets env)
+  "Compile SCCs and generate a final output definition list, merging any present lisp sources."
+  (let ((bindings (loop :for scc :in sccs
+                        :for bindings := (remove-if-not (lambda (binding)
+                                                          (find (car binding) scc))
+                                                        definitions)
+                        :collect (cons (bindings-offset bindings offsets)
+                                       (compile-scc bindings env))))
+        (lisp-forms (mapcar (lambda (lisp-form)
+                              (cons (car (parser:toplevel-lisp-form-source lisp-form))
+                                    (parser:toplevel-lisp-form-body lisp-form)))
+                            lisp-forms)))
+    (mapcan #'cdr (merge-forms bindings lisp-forms))))
+
+(defun definition-bindings (definitions env offsets)
+  "Translate the DEFINITIONS in this TU into bindings, updating an OFFSETS hashtable to record the source offset of each binding's source definition."
+  (loop :for define :in definitions
+        :for offset := (car (tc:toplevel-define-source define))
+        :for name := (tc:node-variable-name (tc:toplevel-define-name define))
+        :for compiled-node := (translate-toplevel define env name)
+
+        :when *codegen-hook*
+          :do (funcall *codegen-hook* ':AST
+                       name
+                       (tc:lookup-value-type env name)
+                       (tc:binding-value define))
+
+        :do (setf (gethash name offsets) offset)
+        :collect (cons name compiled-node)))
+
+(defun instance-bindings (instances env offsets)
+  "Translate the INSTANCES defined by this TU into bindings, updating an OFFSETS hashtable to record the source offset of each binding's source instance."
+  (loop :for instance :in instances
+        :for offset := (car (tc:toplevel-define-instance-source instance))
+        :for instance-bindings := (translate-instance instance env)
+
+        :do (dolist (binding instance-bindings)
+              (setf (gethash (car binding) offsets) offset))
+        :append instance-bindings))
+
 (defun compile-translation-unit (translation-unit monomorphize-table env)
   (declare (type tc:translation-unit translation-unit)
            (type hash-table monomorphize-table)
            (type tc:environment env))
 
-  (let* ((definitions
+  (let* ((offsets (make-hash-table))
+         (definitions
            (append
-            (loop :for define :in (tc:translation-unit-definitions translation-unit)
-                  :for name := (tc:node-variable-name (tc:toplevel-define-name define))
-
-                  :for compiled-node := (translate-toplevel define env name)
-
-                  :do (when *codegen-hook*
-                        (funcall *codegen-hook*
-                                 ':AST
-                                 name
-                                 (tc:lookup-value-type env name)
-                                 (tc:binding-value define)))
-                  :collect (cons name compiled-node))
-
-            ;; HACK: this load bearing reverse should be replaced with an actual solution
-            (loop :for instance :in (reverse (tc:translation-unit-instances translation-unit))
-                  :append (translate-instance instance env))))
-
-         (definition-names
-           (mapcar #'car definitions)))
+            (definition-bindings (tc:translation-unit-definitions translation-unit) env offsets)
+            (instance-bindings (tc:translation-unit-instances translation-unit) env offsets)))
+         (definition-names (mapcar #'car definitions)))
 
     (multiple-value-bind (definitions env)
-        (optimize-bindings
-         definitions
-         monomorphize-table
-         *package*
-         env)
+        (optimize-bindings definitions monomorphize-table *package* env)
 
-      (let ((sccs (node-binding-sccs definitions)))
+      (let ((sccs (node-binding-sccs definitions))
+            (lisp-forms (tc:translation-unit-lisp-forms translation-unit)))
 
         (values
          `(progn
@@ -103,13 +179,7 @@ A function bound here will be called with a keyword category, and one or more ad
                 (list
                  `(declaim (sb-ext:start-block ,@definition-names))))
 
-            ,@(loop :for scc :in sccs
-                    :for bindings
-                      := (remove-if-not
-                          (lambda (binding)
-                            (find (car binding) scc))
-                          definitions)
-                    :append (compile-scc bindings env))
+            ,@(compile-definitions sccs definitions lisp-forms offsets env)
 
             #+sbcl
             ,@(when (eq sb-ext:*block-compile-default* :specified)
--- a/src/faux-macros.lisp
+++ b/src/faux-macros.lisp
@@ -45,6 +45,9 @@
 (define-coalton-editor-macro coalton:define-instance (instance &body method-definitions)
   "Define an instance of a type class. (Coalton top-level operator.)")
 
+(define-coalton-editor-macro coalton:lisp-toplevel (options &body lisp-toplevel-forms)
+  "Include lisp forms. (Coalton top-level operator.)")
+
 (define-coalton-editor-macro coalton:specialize (name from-ty to-ty)
   "Declare a specialization for a function. (Coalton top-level operator.)")
 
--- a/src/parser/toplevel.lisp
+++ b/src/parser/toplevel.lisp
@@ -117,6 +117,11 @@
    #:toplevel-define-instance-compiler-generated ; ACCESSOR
    #:toplevel-define-instance-list               ; TYPE
    #:toplevel-package-name                       ; ACCESSOR
+   #:toplevel-lisp-form                          ; STRUCT
+   #:make-toplevel-lisp-form                     ; CONSTRUCTOR
+   #:toplevel-lisp-form-body                     ; ACCESSOR
+   #:toplevel-lisp-form-source                   ; ACCESSOR
+   #:toplevel-lisp-form-list                     ; TYPE
    #:toplevel-specialize                         ; STRUCT
    #:make-toplevel-specialize                    ; CONSTRUCTOR
    #:toplevel-specialize-from                    ; ACCESSOR
@@ -128,6 +133,7 @@
    #:make-program                                ; CONSTRUCTOR
    #:program-package                             ; ACCESSOR
    #:program-file                                ; ACCESSOR
+   #:program-lisp-forms                          ; ACCESSOR
    #:program-types                               ; ACCESSOR
    #:program-structs                             ; ACCESSOR
    #:program-declares                            ; ACCESSOR
@@ -204,6 +210,8 @@
 ;;;; toplevel-define-instance := "(" "define-instance" "(" ty-predicate ")" docstring? instance-method-definition* ")"
 ;;;;                           | "(" "define-instance" "(" ty-predicate "=>" ty-predicate ")" docstring? instance-method-definition ")"
 ;;;;                           | "(" "define-instance" "(" ( "(" ty-predicate ")" )+ "=>" ty-predicate ")" docstring? instance-method-definition+ ")"
+;;;;
+;;;; toplevel-lisp-form := "(" "lisp-toplevel" "(" ")" lisp-form* ")"
 ;;;;
 ;;;; toplevel-specialize := "(" identifier identifier ty ")"
 
@@ -409,6 +417,19 @@
 (deftype toplevel-define-instance-list ()
   '(satisfies toplevel-define-instance-list-p))
 
+(defstruct (toplevel-lisp-form
+            (:copier nil))
+  (body   (util:required 'body)   :type cons  :read-only t)
+  (source (util:required 'source) :type cons  :read-only t))
+
+(eval-when (:load-toplevel :compile-toplevel :execute)
+  (defun toplevel-lisp-form-list-p (x)
+    (and (alexandria:proper-list-p x)
+         (every #'toplevel-lisp-form-p x))))
+
+(deftype toplevel-lisp-form-list ()
+  '(satisfies toplevel-lisp-form-list-p))
+
 (defstruct (toplevel-specialize
             (:copier nil))
   (from   (util:required 'from)   :type node-variable :read-only t)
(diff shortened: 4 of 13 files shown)


eliaslfox:
Should this the commit to fix CCL be on [this pr](https://github.com/coalton-lang/coalton/pull/1091) instead?

src/parser/toplevel.lisp
```diff
@@ -1048,6 +1050,11 @@ Ensure the package exists, import the COALTON package and required dependencies,
        (push instance (program-instances program))
        t))
 
+    ((coalton:lisp-toplevel)
+     (loop :for form :in (cst:raw (cst:rest form))
+           :do (vector-push-extend form (program-lisp program)))
```
eliaslfox, line 1055: Either push to a list or update the other places in this file to use a vector.

src/typechecker/translation-unit.lisp
```diff
@@ -28,4 +29,5 @@
   (definitions     nil                      :type toplevel-define-list           :read-only t)
   (instances       nil                      :type toplevel-define-instance-list  :read-only t)
   (classes         nil                      :type ty-class-list                  :read-only t)
+  (lisp            nil                                                           :read-only t)
```
eliaslfox, line 32: `lisp-forms`

src/parser/toplevel.lisp
```diff
@@ -1048,6 +1050,11 @@ Ensure the package exists, import the COALTON package and required dependencies,
        (push instance (program-instances program))
        t))
 
+    ((coalton:lisp-toplevel)
+     (loop :for form :in (cst:raw (cst:rest form))
```
eliaslfox, line 1054: You need to check for attributes here and error appropriately.

eliaslfox (dismissed):

src/parser/toplevel.lisp
jbouwman: This uses a list now.

jbouwman:
@stylewarning @eliaslfox do you have guidance on this change?

src/codegen/program.lisp
```diff
@@ -41,41 +42,96 @@ A function bound here will be called with a keyword category, and one or more ad
...
+(defun bindings-offset (bindings offsets)
+  "Given a list of binding names, and a name -> offset map, return the earliest binding start offset."
+  (apply #'min
+         (mapcar (lambda (binding)
+                   (or (gethash (car binding) offsets) 0))
+                 bindings)))
+
+(defun merge-forms (lisp defs)
+  "Merge a list of toplevel-lisp forms and compiled definitions, ensuring that lisp forms are always placed before the original source locations of the compiled definitions."
+  (let (result)
```
macrologist, line 72: nit. If `result` is to be used as a list it should signal this fact here by binding the value to `nil`.

If `result` were simply being declared, later to be provided a value via `(setf result "moo")`, then leaving it uninitialized is acceptable (even preferred). 

I personally don't mind this style, but some do. Looking at you @stylewarning

src/codegen/program.lisp
```diff
@@ -41,41 +42,96 @@ A function bound here will be called with a keyword category, and one or more ad
...
+
+(defun bindings-offset (bindings offsets)
+  "Given a list of binding names, and a name -> offset map, return the earliest binding start offset."
+  (apply #'min
+         (mapcar (lambda (binding)
+                   (or (gethash (car binding) offsets) 0))
+                 bindings)))
+
+(defun merge-forms (lisp defs)
+  "Merge a list of toplevel-lisp forms and compiled definitions, ensuring that lisp forms are always placed before the original source locations of the compiled definitions."
```
macrologist, line 71: Some structural description of the arguments would be appreciated. 

E.g. ```LISP is a list of lists, (OFFSET SOMETHING-ELSE SOMETHING-ELSE-ELSE ...), and DEFS is ....```

or whatever.

src/codegen/program.lisp
```diff
@@ -41,41 +42,96 @@ A function bound here will be called with a keyword category, and one or more ad
...
+  (let (result)
+    (loop :for def :in defs
+          :for next-def-offset := (car def)
+          :do (loop :while (< (caar lisp) next-def-offset)
```
macrologist, line 75: Do we know that `LISP` will never be empty during this loop? If so, cool.  If not.... kaboom.

Looking at this function right here right now, I can't tell that this implicit invariant holds. You know... but maybe it does in practice due to context that I can't appreciate by looking at this function.

src/codegen/program.lisp
```diff
@@ -41,41 +42,96 @@ A function bound here will be called with a keyword category, and one or more ad
...
+                            lisp-forms)))
+    (mapcan #'cdr (merge-forms bindings lisp-forms))))
+
+(defun definition-bindings (definitions env offsets)
+  (loop :for define :in definitions
+        :for offset := (car (tc:toplevel-define-source define))
+        :for name := (tc:node-variable-name (tc:toplevel-define-name define))
+        :for compiled-node := (translate-toplevel define env)
+
+        :do (when *codegen-hook*
```
macrologist, line 102: nit. just out of curiosity, why not `:when *codegen-hook* :do (funcalll *codegen-hook* ...)`? Not a suggestion, the choice just jumped out at me.

Lean in to the `loop`... become the `loop` :P

src/parser/toplevel.lisp
```diff
@@ -604,6 +629,20 @@ consume all attributes"))))
 
     package))
 
+(defun parse-lisp-toplevel-form (form program)
+  "Parse lisp forms that are to be literally included in compiler output.
+
+If the outermost form matches (eval-when (compile-toplevel) ..), evaluate the enclosed forms."
+  (loop :for form :in (cst:raw (cst:rest form))
+        :when (and (eql (car form) 'eval-when)
+                   (some (lambda (k)
```
macrologist, line 638: Optimization Nit. Might be worth factoring this lambda out since it does not closer over any outer binding environment. Either that or `#.(lambda (k) (string-equal k 'compile-toplevel)`.

eliaslfox:
I don't think the compiler should be sorting output forms based on source spans. `coalton-toplevel` is currently intentionally unordered. The unordered nature of forms was done because it provides a much nicer programming language UX.

This implementation adds significant implementation complexity and the behavior will be difficult for users to understand.

```
(coalton-toplevel

  (define (f x) ...)
  
   (lisp-toplevel ...) ; first toplevel depends on g
   
   (define (g x) (h x))
   
   (lisp-toplevel ...) ; second toplevel depends on h
   
   (define (h x) (f x)))
```

I don't know what the order of evaluation is for the above example, but I can't think of an order that feels intuitive.

I think that all lisp toplevel forms should be lifted to the top of files. Complex macros can be defined in a lisp file loaded before the Coalton file where they're used.  The standard library has a lot of complex interaction between Coalton and lisp, and it serves as a bad example of how this feature would be often used. I think we should design around the simpler interaction patterns found in application code.

src/codegen/program.lisp
stylewarning: agree

stylewarning:
Initial thoughts regarding @eliaslfox's comment:

LISP-TOPLEVEL will be confusing and unintuitive if it is allowed to syntactically sit in the middle of the file, and it will be compiled as if it were at the top of the file.

There is no purpose for LISP-TOPLEVEL if all instantiations of it will be moved to the top, since you can just write a lisp file to compile before it.

For a toplevel form called LISP-TOPLEVEL intended to allow a mixture of Coalton and Lisp code, @jbouwman's design is correct and what any reasonable Lisp programmer would expect. It is also reflective of how Coalton's library code is written.

The subtext is that the Coalton library is a special case that isn't going to be reflective of real Coalton code, but I think this assumption will turn out to be slightly wrong. Interop between Lisp and Coalton is a chief selling point, and the standard library shows that front-and-center. 

@eliaslfox's criticism is, really, a criticism to not have LISP-TOPLEVEL. While the criticism is well-intentioned, it appears to only be supported by a rather obtuse and adversarial example, and buttressed by an appeal to the inevitable poor DX someone will have when trying to make use of such an example. Are such adversarial examples likely to show up in practice? Or are we protecting a user that probably won't exist?

**If** LISP-TOPLEVEL offers more detriment than benefit—I'm skeptical of this—then either it should be removed from consideration, or it should be renamed to LISP-PROLOG and LISP-EPILOG and only be syntactically allowed at the start and end of the file. It doesn't make sense any other way. However, the definition of the standard library, along with the desire to have one package per file, have forced our hand into interspersing Lisp and Coalton code as a natural solution to interacting Coalton-Lisp applications.

In another view, we could actually see the root cause of the problem as not being LISP-TOPLEVEL, but our insistence to act like Haskell with implicit file-scoped letrec. Discussions to have explicit letrec have been shot down for looking too gaudy or annoying to type... but maybe explicit is better than implicit for the flexibility in organizing our code, developing editor support, getting Lisp interop, and the other things we get in return? I don't necessarily want to hijack a discussion about LISP-TOPLEVEL as a discussion about letrec, but it seems to be relevant again in Coalton's language design.

Those are my thoughts for now.

src/codegen/program.lisp
jbouwman: I rewrote this in a more generic and empty-input-tolerant tail-recursive fashion, collecting the result in an optional argument, and adding better documentation. Does this style look ok?

src/codegen/program.lisp
jbouwman: done

src/codegen/program.lisp
jbouwman: done, see above comment

src/codegen/program.lisp
jbouwman: done

src/parser/toplevel.lisp
jbouwman: I moved this to a helper eval-toplevel-p function and poked around at the disassembly -- it looks like it's not allocating, and SBCL didn't like #.:

```
toplevel.lisp:719:8:
  error: Objects of type COMPILED-FUNCTION can't be dumped into fasl files.
  note: The first argument never returns a value.
```

jbouwman:
Following up discussion with @stylewarning on 8/2, this changeset now places `lisp-toplevel` behind a `*features*` flag, for the time being, restricting usage to library source code until there is better consensus on semantics.

coalton.asd
```diff
@@ -27,7 +27,8 @@
   :version (:read-file-form "VERSION.txt")
   :around-compile (lambda (compile)
                     (let (#+sbcl (sb-ext:*derive-function-types* t)
-                          #+sbcl (sb-ext:*block-compile-default* :specified))
+                          #+sbcl (sb-ext:*block-compile-default* :specified)
+                          (*features* (cons :coalton-lisp-toplevel *features*)))
```
stylewarning, line 31: quote

jbouwman: 🙇

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+                                                        definitions)
+                        :collect (cons (bindings-offset bindings offsets)
+                                       (compile-scc bindings env))))
+        (lisp-forms (mapcar (lambda (lisp-form)
+                              (cons (car (parser:toplevel-lisp-form-source lisp-form))
+                                    (parser:toplevel-lisp-form-body lisp-form)))
+                            lisp-forms)))
+    (mapcan #'cdr (merge-forms bindings lisp-forms))))
+
+(defun definition-bindings (definitions env offsets)
```
stylewarning, line 103: add docstring

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+
+        :when *codegen-hook*
+          :do (funcall *codegen-hook* :ast
+                       name
+                       (tc:lookup-value-type env name)
+                       (tc:binding-value define))
+
+        :collect (cons name compiled-node)))
+
+(defun instance-bindings (instances env offsets)
```
stylewarning, line 117: add docstring

stylewarning (changes requested):
partial review

src/parser/toplevel.lisp
```diff
@@ -979,6 +1023,30 @@ consume all attributes"))))
        (push instance (program-instances program))
        t))
 
+    ((coalton:lisp-toplevel)
+     (unless (alexandria:featurep :coalton-lisp-toplevel)
```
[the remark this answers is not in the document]
jbouwman: done

src/parser/toplevel.lisp
```diff
@@ -979,6 +1023,30 @@ consume all attributes"))))
...
+       (error 'parse-error
+              :err (se:source-error
+                    :span (cst:source (cdr (aref attributes 0)))
+                    :file file
+                    :message "Invalid attribute for lisp-toplevel"
+                    :primary-note "lisp-toplevel cannot have attributes"
+                    :notes
+                    (list
+                     (se:make-source-error-note
+                      :type :secondary
```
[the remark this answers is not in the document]
jbouwman: done, throughout

src/parser/toplevel.lisp
```diff
@@ -688,6 +713,25 @@ consume all attributes"))))
 
 ;; end package support
 
+(defun eval-toplevel-p (form)
+  "T if form is an instance of EVAL-WHEN containing a COMPILE-TOPLEVEL symbol."
+  (and (eql (car form) 'eval-when)
+       (some (lambda (k)
+               (string-equal 'compile-toplevel k))
```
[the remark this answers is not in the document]
jbouwman: added

src/parser/toplevel.lisp
```diff
@@ -688,6 +713,25 @@ consume all attributes"))))
 
 ;; end package support
 
+(defun eval-toplevel-p (form)
+  "T if form is an instance of EVAL-WHEN containing a COMPILE-TOPLEVEL symbol."
+  (and (eql (car form) 'eval-when)
```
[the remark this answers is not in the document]
jbouwman: both done

src/parser/toplevel.lisp
```diff
@@ -688,6 +713,25 @@ consume all attributes"))))
...
+       (some (lambda (k)
+               (string-equal 'compile-toplevel k))
+             (cadr form))))
+
+(defun parse-lisp-toplevel-form (form program)
```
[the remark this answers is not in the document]
jbouwman: That is the reason: the macro definitions are needed immediately in the case that macros expand to coalton code, for instance

```
  (lisp-toplevel ()
    ...
    (cl:eval-when (:compile-toplevel :load-toplevel)
      (cl:defmacro define-eq (type)
        `(define-instance (Eq ,type)
    ... )

  (define-eq Integer)
```

coalton.asd
```diff
@@ -27,7 +27,8 @@
   :version (:read-file-form "VERSION.txt")
   :around-compile (lambda (compile)
                     (let (#+sbcl (sb-ext:*derive-function-types* t)
-                          #+sbcl (sb-ext:*block-compile-default* :specified))
+                          #+sbcl (sb-ext:*block-compile-default* :specified)
+                          (*features* (cons ':coalton-lisp-toplevel *features*)))
```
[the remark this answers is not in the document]
jbouwman: done

src/faux-macros.lisp
```diff
@@ -45,6 +45,9 @@
 (define-coalton-editor-macro coalton:define-instance (instance &body method-definitions)
   "Define an instance of a type class. (Coalton top-level operator.)")
 
+(define-coalton-editor-macro coalton:lisp-toplevel (name from-ty to-ty)
```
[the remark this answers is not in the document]
jbouwman: done

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+
+1. The inputs and output are lists of 2-lists, structured as (OFFSET FORM).
+2. The order of both lists is preserved.
+3. Lists are merged by recursively selecting the head of the list with the lowest offset."
+  (cond ((endp forms-a)
+         (return-from merge-forms
+           (append (nreverse merged) forms-b)))
+        ((endp forms-b)
+         (return-from merge-forms
+           (append (nreverse merged) forms-a)))
```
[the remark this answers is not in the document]
jbouwman: where has nreconc been all my life

src/codegen/program.lisp
jbouwman: done

src/codegen/program.lisp
jbouwman: done

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+
+(defun merge-forms (forms-a forms-b &optional merged)
+  "Stably merge two lists of forms.
+
+1. The inputs and output are lists of 2-lists, structured as (OFFSET FORM).
+2. The order of both lists is preserved.
+3. Lists are merged by recursively selecting the head of the list with the lowest offset."
+  (cond ((endp forms-a)
+         (return-from merge-forms
+           (append (nreverse merged) forms-b)))
```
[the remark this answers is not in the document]
jbouwman: done

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+                        :for bindings := (remove-if-not (lambda (binding)
+                                                          (find (car binding) scc))
+                                                        definitions)
+                        :collect (cons (bindings-offset bindings offsets)
+                                       (compile-scc bindings env))))
+        (lisp-forms (mapcar (lambda (lisp-form)
+                              (cons (car (parser:toplevel-lisp-form-source lisp-form))
+                                    (parser:toplevel-lisp-form-body lisp-form)))
+                            lisp-forms)))
+    (mapcan #'cdr (merge-forms bindings lisp-forms))))
```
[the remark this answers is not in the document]
jbouwman: yes, all inputs are newly consed and not reused

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+;;
+;; Then when compile-definitions emits the full set of output
+;; definitions, any lisp source forms that occurred earlier in the
+;; file are emitted first.
+
+(defun bindings-offset (bindings offsets)
+  "Given a list of binding names, and a name -> offset map, return the earliest binding start offset."
+  (apply #'min
+         (mapcar (lambda (binding)
+                   (or (gethash (car binding) offsets) 0))
```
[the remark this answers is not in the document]
jbouwman: done

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+;;   bindings (lisp definitions):
+;;     (300 b1 .. bn)
+;;
+;; Then when compile-definitions emits the full set of output
+;; definitions, any lisp source forms that occurred earlier in the
+;; file are emitted first.
+
+(defun bindings-offset (bindings offsets)
+  "Given a list of binding names, and a name -> offset map, return the earliest binding start offset."
+  (apply #'min
```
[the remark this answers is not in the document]
jbouwman: done - reduce

src/parser/toplevel.lisp
```diff
@@ -205,6 +211,8 @@
 ;;;;                           | "(" "define-instance" "(" ty-predicate "=>" ty-predicate ")" docstring? instance-method-definition ")"
 ;;;;                           | "(" "define-instance" "(" ( "(" ty-predicate ")" )+ "=>" ty-predicate ")" docstring? instance-method-definition+ ")"
 ;;;;
+;;;; toplevel-lisp-form := "(" "lisp-toplevel" lisp-form* ")"
```
stylewarning, line 214: add the extra () to the syntax

src/parser/toplevel.lisp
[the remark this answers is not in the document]
stylewarning: yeah i see. normally in compile-file this isn't necessary because defmacro gets done early enough

tests/parser-test-files/lisp-toplevel.txt
```diff
@@ -0,0 +1,22 @@
+================================================================================
```
stylewarning, line 1: can you add a test that just has plain invalid Lisp syntax inside of a lisp-toplevel form?

stylewarning (changes requested):
i feel this is a little anemic on tests. can we add some?

tests/parser-test-files/lisp-toplevel.txt
jbouwman: Invalid lisp syntax like `(defmacro 777)` will only be detected at parse time if an eval-when clause is detected and contents are evaluated. I can a handler and tests for this case.

Hard syntax errors like unterminated strings, unknown character literals, and so on will signal inside of eclector, before anything in this changeset is run. For instance, an unterminated `(package` form produces:

```
Unhandled ECLECTOR.READER:UNTERMINATED-LIST in thread #<SB-THREAD:THREAD tid=259 "main thread" RUNNING
                                                         {7004FD0733}>:
  While reading list, expected the character ) when input ended.

Backtrace for: #<SB-THREAD:THREAD tid=259 "main thread" RUNNING {7004FD0733}>
```
This is existing (not so good) behavior -- leaving unterminated forms in, e.g. library/types.lisp causes: 
```
Unhandled ECLECTOR.READER:UNTERMINATED-LIST in thread #<SB-THREAD:THREAD tid=259 "main thread" RUNNING
                                                         {7004FD0733}>:
  While reading list, expected the character ) when input ended.

Backtrace for: #<SB-THREAD:THREAD tid=259 "main thread" RUNNING {7004FD0733}>
```

The third case is lisp code that isn't evaluated at parse time, and is emitted unchanged to the .lisp output that is subsequently given to compile-file. I'm not sure how to handle that one, though, I am pretty sure that it would not be hard to map an error offset thrown from compile-file back to an originating form.

src/parser/toplevel.lisp
jbouwman: done

src/codegen/program.lisp
```diff
@@ -41,41 +42,104 @@ A function bound here will be called with a keyword category, and one or more ad
...
+         (mapcar (lambda (binding)
+                   (or (gethash (car binding) offsets) 0))
+                 bindings)))
+
+(defun merge-forms (forms-a forms-b &optional merged)
+  "Stably merge two lists of forms.
```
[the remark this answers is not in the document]
jbouwman: added example

tests/parser-test-files/lisp-toplevel.txt
stylewarning: file a ticket for the 2nd case

tests/parser-tests.lisp
```diff
@@ -44,4 +44,6 @@
       (parse-file file))))
 
 (deftest test-parse-package-suite ()
+  (let ((*features* (cons ':coalton-lisp-toplevel *features*)))
```
stylewarning, line 47: can we bind this in the `.asd` like the other one?

jbouwman: I just tried, and, it has no effect, because the feature is checked during test execution.

tests/parser-test-files/lisp-toplevel.txt
```diff
@@ -0,0 +1,64 @@
...
+
+--------------------------------------------------------------------------------
+
+error: Invalid lisp-toplevel form
+  --> test:4:2
+   |
+ 3 |   (lisp-toplevel
+   |  _-
+ 4 | |   (defvar *x* nil))
+   | |   ^^^^^^^^^^^^^^^^ saw 'def' form: in lisp-toplevel, code must be preceded by an empty options list
```
stylewarning, line 41: 💯

stylewarning (approved):
Errors are much better!

We will merge the inlining bugfix PR first since code is broken, then this one. Some rebasing will likely be necessary.

jbouwman:
> Errors are much better!
> 
> We will merge the inlining bugfix PR first since code is broken, then this one. Some rebasing will likely be necessary.

@amorphedstar cleaned up the nits in the inliner, so I merged that, and did the consequent very minor rebasing

stylewarning (approved):

PR #1131, opened 2024-06-24
Fail if no input files are selected by tests

jbouwman:
Parser tests are not enabled because the suite is looking in 'bad', and files are in 'bad-files'.

This reenables tests, and adds a guardrail. At least one assertion will need updating.

### DIFF
--- a/src/error.lisp
+++ b/src/error.lisp
@@ -223,13 +223,24 @@ NOTES and HELP-NOTES may optionally be supplied notes and help messages."
       ;; Print the error message and location
       (multiple-value-bind (line-number line-start-index)
           (get-line-from-index file-stream (coalton-error-location error))
-        (format stream
-                "~(~A~): ~A~%  --> ~A:~D:~D~%"
-                (coalton-error-type error)
-                (coalton-error-message error)
-                (coalton-file-name (coalton-error-file error))
-                line-number
-                (- (coalton-error-location error) line-start-index)))
+
+        ;; get-line-from-index sometimes returns a line-start-index
+        ;; that is 1 larger than (coalton-error-location error),
+        ;; resulting in a column number of -1. The following lines
+        ;; compensate for that and can be removed if and when that is
+        ;; fixed.
+        (let ((column-number (- (coalton-error-location error) line-start-index)))
+          (when (< column-number 0)
+            (decf line-number)
+            (setf column-number (1- (length (get-nth-line file-stream line-number)))))
+
+          (format stream
+                  "~(~A~): ~A~%  --> ~A:~D:~D~%"
+                  (coalton-error-type error)
+                  (coalton-error-message error)
+                  (coalton-file-name (coalton-error-file error))
+                  line-number
+                  column-number)))
 
       ;; Print the error notes
       (let* (;; We need to keep track of the current depth of multiline
--- a/src/parser/toplevel.lisp
+++ b/src/parser/toplevel.lisp
@@ -1745,15 +1745,13 @@ consume all attributes")))
                  :message "Invalid type variable"
                  :primary-note "expected keyword symbol"
                  :help-notes
-                 (cond
-                   ((eq *package* (symbol-package (cst:raw form)))
-                    (list
-                     (make-coalton-error-help
-                      :span (cst:source form)
-                      :replacement
-                      (lambda (existing)
-                        (concatenate 'string ":" existing))
-                      :message "add `:` to symbol")))))))
+                 (list
+                  (make-coalton-error-help
+                   :span (cst:source form)
+                   :replacement
+                   (lambda (existing)
+                     (concatenate 'string ":" existing))
+                   :message "add `:` to symbol")))))
 
   (make-keyword-src
    :name (cst:raw form)
--- a/tests/parser-tests.lisp
+++ b/tests/parser-tests.lisp
@@ -2,7 +2,10 @@
 
 (deftest test-parser ()
   (labels ((test-files (pattern)
-             (directory (merge-pathnames pattern (asdf:system-source-directory "coalton/tests"))))
+             (let ((files (directory (merge-pathnames pattern (asdf:system-source-directory "coalton/tests")))))
+               (when (endp files)
+                 (error "No test files match pattern '~A'" pattern))
+               files))
 
            (parse-file (file)
              (with-open-file (stream file
@@ -22,7 +25,7 @@
                      "no errors")
                  (error:coalton-base-error (c)
                    (princ-to-string c))))))
-    (dolist (file (test-files "tests/parser-test-files/bad/*.coal"))
+    (dolist (file (test-files "tests/parser-test-files/bad-files/*.coal"))
       (let ((error-file (make-pathname :type "error"
                                        :defaults file)))
         (cond ((uiop:file-exists-p error-file)
@@ -33,5 +36,5 @@
                (signals parser:parse-error
                  (parse-file file))))))
 
-    (dolist (file (test-files "tests/parser-test-files/good/*.coal"))
+    (dolist (file (test-files "tests/parser-test-files/good-files/*.coal"))
       (parse-file file))))
--- a/tests/parser-test-files/bad-files/define-instance.7.error
+++ b/tests/parser-test-files/bad-files/define-instance.7.error
@@ -1,5 +1,5 @@
 error: Malformed instance definition
-  --> test:4:17
+  --> test:4:16
    |
  4 |  (define-instance)
    |                  ^ expected an instance head
--- a/tests/parser-test-files/bad-files/parse-attribute.5.error
+++ b/tests/parser-test-files/bad-files/parse-attribute.5.error
@@ -1,5 +1,5 @@
 error: Malformed repr :native attribute
-  --> test:4:14
+  --> test:4:13
    |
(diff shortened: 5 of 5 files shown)


src/error.lisp
```diff
@@ -223,13 +223,24 @@ NOTES and HELP-NOTES may optionally be supplied notes and help messages."
       ;; Print the error message and location
       (multiple-value-bind (line-number line-start-index)
           (get-line-from-index file-stream (coalton-error-location error))
```
Izaakwltn, line 225: Should `get-line-from-index` be changed instead? Would the logic be equivalent?

src/parser/toplevel.lisp
```diff
@@ -1619,91 +1619,61 @@ consume all attributes")))
...
-     :name (make-identifier-src
-            :name (node-variable-name (parse-variable (cst:first method-form) file))
-            :source (cst:source (cst:first method-form)))
-     :docstring docstring
-     :type (parse-qualified-type (if docstring
-                                     (cst:third method-form)
-                                     (cst:second method-form))
-                                 file)
-     :source (cst:source method-form))))
+  (macrolet ((fail (local-form note)
```
Izaakwltn, line 1638: Are these changes just reorganizational or are they changing behavior?

src/parser/toplevel.lisp
```diff
@@ -1619,91 +1619,61 @@ consume all attributes")))
...
+         (setf name (parse-variable (cst:first method-form) file))
+         (setf type (parse-qualified-type (cst:second method-form) file)))
+        (3                              ; good case 2: (<name> "docstring" <type>)
+         (unless (and (cst:atom (cst:first method-form))
+                      (identifierp (cst:raw (cst:first method-form))))
+           (fail (cst:first method-form) "expected symbol"))
+         (setf name (parse-variable (cst:first method-form) file))
+         (unless (and (cst:atom (cst:second method-form))
+                      (stringp (cst:raw (cst:second method-form))))
+           (fail (cst:second method-form) "expected docstring"))
```
Izaakwltn, line 1666: I think this should be changed to `"unexpected trailing form"` on the `third` element, as discussed below.

tests/parser-test-files/bad-files/define-class.4.error
```diff
@@ -1,7 +1,7 @@
 error: Malformed method definition
-  --> test:5:7
+  --> test:5:4
    |
  4 |  (define-class (C :a)
    |                ------ in this class definition
  5 |   (m :a :b))
-   |         ^^ unexpected trailing form
+   |      ^^ expected docstring
```
Izaakwltn, line 7: I know I wrote this, but I'm wondering whether this implies that a docstring is necessary, as opposed to just the second element needing to be a docstring if there are three in the form.

Maybe `"unexpected trailing form"` on the third form is the proper error.

tests/parser-tests.lisp
```diff
@@ -33,5 +36,5 @@
                (signals parser:parse-error
                  (parse-file file))))))
 
-    (dolist (file (test-files "tests/parser-test-files/good/*.coal"))
+    (dolist (file (test-files "tests/parser-test-files/good-files/*.coal"))
```
Izaakwltn, line 39: 👍

tests/parser-tests.lisp
```diff
@@ -22,7 +25,7 @@
                      "no errors")
                  (error:coalton-base-error (c)
                    (princ-to-string c))))))
-    (dolist (file (test-files "tests/parser-test-files/bad/*.coal"))
+    (dolist (file (test-files "tests/parser-test-files/bad-files/*.coal"))
```
Izaakwltn, line 28: 👍

tests/parser-test-files/bad-files/define-class.4.error
Izaakwltn: Long story short, I wonder whether the first error is still more accurate and advantageous.

src/error.lisp
jbouwman: When I attempted to change `get-line-from-index` to match the expectation here, a number of other tests broke, so I think that may be something to address separately.

src/parser/toplevel.lisp
jbouwman: These are purely reorganizational, though they do drop some repeated probing for nullity to check the length of the method form.

Izaakwltn: If it's just reorganizational, I wonder if that should be a separate PR applying the same structure to each parse function? I think it's reasonable to find a more condensed structure but it should probably be one fell swoop.

src/error.lisp
Izaakwltn: Sounds good!

tests/parser-test-files/bad-files/define-class.4.error
jbouwman: That does correspond better to what seems to be going on with the program input, I'll change it.

src/parser/toplevel.lisp
jbouwman: I recall what the motivation was: reporting 'unexpected trailing form' needs to consider whether or not a docstring is present in order to decide whether to make the third or fourth form. The original version marks the wrong one:

```
Unhandled FIASCO::FAILED-ASSERTION in thread #<SB-THREAD:THREAD tid=259 "main thread" RUNNING
                                                {7004FD06F3}>:
  Test assertion failed when running TEST-PARSER:

Strings differ at offset 49 of expected error define-class.21.error (A) and generated error (B):
A: error: Malformed method definition
  --> test:5:13
   |
 4 |  (define-class (C :a)
   |                ------ in this class definition
 5 |   (m "doc" :a :b))
   |               ^^ unexpected trailing form

B: error: Malformed method definition
  --> test:5:10
   |
 4 |  (define-class (C :a)
   |                ------ in this class definition
 5 |   (m "doc" :a :b))
   |            ^^ unexpected trailing form
```

(also, the method is 40 ll shorter)

tests/parser-test-files/bad-files/define-class.4.error
jbouwman: done!

src/parser/toplevel.lisp
jbouwman: done, as according to the comment on define-class.4.error

eliaslfox:
@jbouwman in the future please split changes like this into multiple pull requests? Combining changesets makes review more difficult.

src/parser/toplevel.lisp
Izaakwltn: Potential minimal patch:
```
1 file changed, 20 insertions(+), 18 deletions(-)
src/parser/toplevel.lisp | 38 ++++++++++++++++++++------------------

modified   src/parser/toplevel.lisp
@@ -1645,7 +1645,7 @@ consume all attributes")))
               (cst:null (cst:rest (cst:rest (cst:rest method-form)))))
     (error 'parse-error
            :err (coalton-error
-                 :span (cst:source (cst:first (cst:rest (cst:rest method-form))))
+                 :span (cst:source (cst:first (cst:rest (cst:rest (cst:rest method-form)))))
                  :file file
                  :message "Malformed method definition"
                  :primary-note "unexpected trailing form"
@@ -1672,28 +1672,30 @@ consume all attributes")))
                    :span (cst:source (cst:second form))
                    :message "in this class definition")))))
 
-  ;; either list of length 2 or list of length 3 with docstring
-  (unless (or (cst:null (cst:rest (cst:rest method-form)))
-              (and (cst:atom (cst:second method-form))
-                   (stringp (cst:raw (cst:second method-form)))))
-    (error 'parse-error
-           :err (coalton-error
-                 :span (cst:source (cst:second method-form))
-                 :file file
-                 :message "Malformed method definition"
-                 :primary-note "expected docstring"
-                 :notes
-                 (list
-                  (make-coalton-error-note
-                   :type :secondary
-                   :span (cst:source (cst:second form))
-                   :message "in this class definition")))))
-
   (let (docstring)
     (when (and (cst:atom (cst:second method-form))
                (stringp (cst:raw (cst:second method-form))))
       (setf docstring (cst:raw (cst:second method-form))))
 
+    ;; either list of length 2 or list of length 3 with docstring
+    (unless (or (cst:null (cst:rest (cst:rest method-form)))
+                (and (cst:atom (cst:second method-form))
+                     (stringp (cst:raw (cst:second method-form)))))
+      (error 'parse-error
+                   :span (cst:source (if docstring
+                                         (cst:fourth method-form)
+                                         (cst:third method-form)))
+                   :file file
+                   :message "Malformed method definition"
+                   :primary-note "unexpected trailing form"
+                   :notes
+                   (list
+                    (make-coalton-error-note
+                     :type :secondary
+                     :span (cst:source (cst:second form))
+                     :message "in this class definition")))))
+
     (make-method-definition
      :name (make-identifier-src
             :name (node-variable-name (parse-variable (cst:first method-form) file))


```

jbouwman: In the name of getting tests back into CI, the most straightforward thing to do seem to be to postpone any changes here and update the define-class.4 parser test to reflect current behavior of docstring parsing -- I've done that.

Izaakwltn (approved):
Looks good to me- Though it may need to be adjusted depending on whether this PR or #1132 get merged first

PR #1135, opened 2024-06-25
File format for grouped parser tests

jbouwman:
Parser verification tests consist of two files per test case, matched by sequence number. This makes it cumbersome to review test coverage for any given parser feature. This PR proposes a delimited input format that supports grouping of test cases:
```
================================================================================
header
================================================================================
program
--------------------------------------------------------------------------------
error
```

See tests/parser-test-files/bad-files/parse-type.txt


jbouwman:
stashing for now, to propose later

PR #1136, opened 2024-06-26
Rewrite get-line-from-index

jbouwman:
Allow get-line-from-index to handle character offsets that point to line endings.

- add assertion for requests beyond eof
- rewrite in 'loop'
- remove utf code point calculation: all errors are character index relative
- remove column number patch
- adjust test assertions

This deletes a patch that was introduced in https://github.com/coalton-lang/coalton/pull/1131

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -43,12 +43,14 @@
                #:concrete-syntax-tree
                #:eclector-concrete-syntax-tree
                #:named-readtables
-               #:source-error)
+               #:source-error
+               #:trivial-gray-streams)
   :pathname "src/"
   :serial t
   :components ((:file "package")
                (:file "settings")
                (:file "utilities")
+               (:file "stream")
                (:file "global-lexical")
                (:file "constants")
                (:module "algorithm"
@@ -333,6 +335,7 @@
   :serial t
   :components ((:file "package")
                (:file "utilities")
+               (:file "stream-tests")
                (:file "tarjan-scc-tests")
                (:file "reader-tests")
                (:file "error-tests")
--- a/source-error/src/error.lisp
+++ b/source-error/src/error.lisp
@@ -219,23 +219,13 @@
       (multiple-value-bind (line-number line-start-index)
           (get-line-from-index file-stream (source-error-location error))
 
-        ;; get-line-from-index sometimes returns a line-start-index
-        ;; that is 1 larger than (coalton-error-location error),
-        ;; resulting in a column number of -1. The following lines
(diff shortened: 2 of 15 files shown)


src/error.lisp
```diff
@@ -522,22 +512,17 @@ Returns (VALUES LINE-NUM LINE-START-INDEX)"
...
-          :when (char= char #\Newline)
-            :do (setf line (1+ line)
-                      line-start-index (1+ i))
-          :when (>= i index)
-            :return nil)
-    (values line line-start-index)))
+  (loop :with line-num := 1
+        :with line-start-index := 0
+        :for char := (read-char file nil nil)
+        :for char-index := 0 :then (1+ char-index)
```
stylewarning, line 518: for char-index from 0

src/error.lisp
```diff
@@ -522,22 +512,17 @@ Returns (VALUES LINE-NUM LINE-START-INDEX)"
...
+        :unless char
```
stylewarning, line 519: i prefer :when (null char) only because char isn't truly a boolean

stylewarning (approved):
My review is just nits.

We may want to be explicit about our external format in Coalton.

I'm not sure if byte offsets was intentional (hence UTF8 stuff) or useful, so that's the only bit I'm uncertain about. I would double check with the git blamed.

src/error.lisp
jbouwman: done

src/error.lisp
jbouwman: done

src/error.lisp
```diff
@@ -224,23 +224,13 @@ NOTES and HELP-NOTES may optionally be supplied notes and help messages."
...
-
-          (format stream
-                  "~(~A~): ~A~%  --> ~A:~D:~D~%"
-                  (coalton-error-type error)
-                  (coalton-error-message error)
-                  (coalton-file-name (coalton-error-file error))
-                  line-number
-                  column-number)))
+        (format stream
+                "~(~A~): ~A~%  --> ~A:~D:~D~%"
```
Izaakwltn, line 228: Is there something with the formatting here, like extraneous spaces, or is it just a github bug?

Izaakwltn (approved):
Looks good to me

jbouwman:
@stylewarning was right, there is at least one place where character vs octet indexing needs clarification -- https://github.com/coalton-lang/coalton/pull/1137 ought to be diagnosed and resolved before this is merged.

Izaakwltn (approved):
lgtm

PR #1138, opened 2024-06-28
Extract source error code to a separate asdf system

eliaslfox:
This makes it easier to reuse the error rendering machinery in other projects. It also enforces a strict seperating between the error system and Coalton proper.


colescott (approved):
nice

stylewarning:
:local-nicknames (#:oof #:source-error)

PR #1139, opened 2024-07-03
Replace hash tables in environment with alists

jbouwman:
Hash table instances aren't, by default, printed readably in Allegro and CCL, preventing generation of environment updates in source for coalton when there are classes and structs present.

Convert the few occurrences of these into association lists, which are probably more efficient, since the entry counts are always small.

This emerges from incremental conversion of the standard library to .coal files, in #1129

### DIFF
--- a/src/doc/generate-documentation.lisp
+++ b/src/doc/generate-documentation.lisp
@@ -427,8 +427,8 @@
                     :type (tc:type-entry-type entry)
                     :tyvars (tc:type-entry-tyvars entry)
                     :fields (tc:struct-entry-fields struct-entry)
-                    :field-docstrings (tc:struct-entry-field-docstrings struct-entry)
-                    :field-tys (tc:struct-entry-field-tys struct-entry)
+                    :field-docstrings (tc:get-table (tc:struct-entry-field-docstrings struct-entry))
+                    :field-tys (tc:get-table (tc:struct-entry-field-tys struct-entry))
                     :instances applicable-instances)
                    output-structs))
 
--- a/src/typechecker/map.lisp
+++ b/src/typechecker/map.lisp
@@ -1,0 +1,65 @@
+;;; Environment maps provide readable representations of hash
+;;; table-valued fields present in Coalton environment updates
+;;; in generated Lisp code.
+
+(defpackage #:coalton-impl/typechecker/map
+  (:use
+   #:cl)
+  (:local-nicknames
+   (#:util #:coalton-impl/util))
+  (:export
+   #:environment-map
+   #:make-map
+   #:get-table
+   #:get-value))
+
+(in-package #:coalton-impl/typechecker/map)
+
+(defstruct environment-map
+  "A readable, hash table-backed map."
+  (table (util:required 'table) :type hash-table :read-only t)
+  (test  (util:required 'test)  :type symbol     :read-only t))
+
+(defun make-map (&key initial-contents (test 'equal))
+  "Construct an environment map containing the keys and values of the association list ALIST.
+
+Environment maps have a readable representation as an alist.
+Hash table is initialized using TEST."
+  (let ((table (make-hash-table :test test)))
+    (dolist (cons initial-contents)
+      (alexandria:ensure-gethash (car cons) table (cdr cons)))
+    (make-environment-map :table table
+                          :test test)))
+
+(defmethod make-load-form ((self environment-map) &optional environment)
+  (make-load-form-saving-slots self :environment environment))
+
+(defmethod print-object ((self environment-map) stream)
+  (if *print-readably*
+      (let ((alist (alexandria:hash-table-alist (environment-map-table self))))
+        (write-string "#.(" stream)
+        (write 'make-map :stream stream)
+        (dolist (value `(:initial-contents ',alist :test ',(environment-map-test self)))
+          (write-char #\Space stream)
+          (write value :stream stream))
+        (write-string ")" stream))
+      (call-next-method)))
+
+(declaim (inline get-table))
+(defun get-table (environment-map)
+  "Return the underlying hash table that provides storage for environment map ENVIRONMENT-MAP."
+  (declare (type environment-map environment-map))
+  (environment-map-table environment-map))
+
+(declaim (inline get-value))
+(defun get-value (environment-map key)
+  "Return the value in environment map ENVIRONMENT-MAP associated with KEY."
+  (declare (type environment-map environment-map))
+  (gethash key (environment-map-table environment-map)))
+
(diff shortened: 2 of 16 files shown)


src/doc/generate-documentation.lisp
```diff
@@ -26,8 +26,8 @@
   (type             (util:required 'type)             :type tc:ty                     :read-only t)
   (tyvars           (util:required 'tyvars)           :type tc:tyvar-list             :read-only t)
   (fields           (util:required 'fields)           :type util:string-list          :read-only t)
-  (field-docstrings (util:required 'field-docstrings) :type hash-table                :read-only t)
-  (field-tys        (util:required 'field-tys)        :type hash-table                :read-only t)
+  (field-docstrings (util:required 'field-docstrings) :type (or null cons)            :read-only t)
```
stylewarning, line 29: or null cons = list

stylewarning (changes requested):
I'm supportive of this change. 


Before I accept this commit, could we discuss the merit of making a DEFSTRUCT-like macro called DEFASSOC that defines an association-list type, constructor, getter, putter, and setter?

I wouldn't mind even if it defined a new structure type that encapsulated the association list.

I'm in general not a fan of such frivolous macros, but (1) schlepping around `equal`, and (2) losing sight of the "true" data type when you use `nil` as the "constructor" makes the code a little less easy to read in my view.

jbouwman:
> I'm in general not a fan of such frivolous macros, but (1) schlepping around `equal`, and (2) losing sight of the "true" data type when you use `nil` as the "constructor" makes the code a little less easy to read in my view.

That sounds perfectly reasonable to me. As a starting point, 1 & 2 could be covered by bundling data and test with a struct, possibly

```
(defstruct (alist (:copier nil))
  (data nil
   :type list)
  (test #'equal
   :type function))

(defun alist-get (alist key)
  (cdr (assoc key (alist-data alist)
              :test (alist-test alist))))

(defun alist-set (alist key value)
  (push (cons key value)
        (alist-data alist)))

(defun alist-set-p (alist key)
  (not (null (assoc key (alist-data alist)
                    :test (alist-test alist)))))

(defun alist-put (alist key value)
  (make-alist :data (cons (cons key value)
                          (alist-data alist))
              :test (alist-test alist)))
```
Looking at that snippet, there are a number of tendentious things like putting the instance in 1st position, embedding the type name in the function name, distinguishing destructive = set and nondestructive = put, and near collision with ANSI CL's alist-copy. But tell me what you think.

src/doc/generate-documentation.lisp
jbouwman: done, thanks

stylewarning:
A problem with stashing the function away is that those might not be externalizable. The immutable nature of the standard alist is also appealing. Maybe the test function is just implicit in the functions defined, and the struct is just there to give type safety (but otherwise the contents are :READ-ONLY T).

eliaslfox (changes requested):
I'm not in favor of making this change. I don't understand why we need to dump the environment update to a readable s-expression. 

Alists also have terrible performance for lookups, which should be the most common operation. The current uses for hash tables (fields on a struct, methods in a class) will generally have small datasets, but I don't want to lose the ability to use hash tables for larger datasets should the need arise.

stylewarning:
@eliaslfox Of course the best thing would be to benchmark, but if I understand correctly, the sets are small in practice. Hiding them behind an abstract data type would allow the choice of representation to change in the future, at the cost of having a little bit more verbose code.

@jbouwman will have to discuss why they need to be externalizable objects. If I recall correctly it had to do with some non-SBCL issue. ANSI 3.2.4.2.2 doesn't permit conforming programs to depend on some hash tables being externalizable, namely those which have a non-unique bijection (under "similarity) of keys. Whether this matters or not, I'm not sure...

jbouwman:
Outcome of steering meeting: add make-load-forms for hash tables.

eliaslfox (approved):
LGTM

src/typechecker/map.lisp
```diff
@@ -0,0 +1,66 @@
...
+
+Environment maps have a readable representation as an alist.
+Hash table is initialized using TEST."
+  (let ((table (make-hash-table :test test)))
+    (dolist (cons initial-contents)
+      (alexandria:ensure-gethash (car cons) table (cdr cons)))
+    (make-environment-map :table table
+                          :test test)))
+
+(defmethod make-load-form ((self environment-map) &optional environment)
```
stylewarning, line 35: @jbouwman This will make a load form for the object itself, but don't we retain the problem of not having a load form for the underlying hash table? This is a question out of ignorance; I'm not sure.

I'd have assumed we needed to handle the hash table specifically, since we were having trouble with the hash-tables to begin with, motivating this change in the first place.

src/typechecker/map.lisp
```diff
@@ -0,0 +1,66 @@
...
+  (environment-map-table environment-map))
+
+(declaim (inline get-value))
+(defun get-value (environment-map key)
+  "Return the value in environment map ENVIRONMENT-MAP associated with KEY."
+  (declare (type environment-map environment-map))
+  (gethash key (environment-map-table environment-map)))
+
+(declaim (inline set-value))
+(defun set-value (environment-map key value)
```
stylewarning, line 62: Not needed, but you could have also done

```
(defun (setf get-value) (new-value map key)
  (set-value map key new-value))
```

tests/typechecker/map-tests.lisp
```diff
@@ -0,0 +1,16 @@
+(in-package #:coalton-tests)
+
+(defstruct map-container
+  map)
+
+(deftest test-environment-map ()
```
stylewarning, line 6: how about a compile-file test?

src/typechecker/map.lisp
jbouwman: I believe the supported Lisps do already provide load forms for hash tables: otherwise the existing compilation scheme of macroexpansion inside compile-file wouldn't be working. This make-load-form method allows that to continue to work -- there are implementations of make-load-form for all other types that are reachable in persisted environments.

The print-object method that follows is the key to representing hash tables in portable, standalone lisp source code.

tests/typechecker/map-tests.lisp
jbouwman: good point, shall add.

src/typechecker/map.lisp
stylewarning: Got it, thanks!

tests/typechecker/map-tests.lisp
jbouwman: added!

stylewarning (approved):

PR #1140, opened 2024-07-05
Make spelling of 'indices' consistent

jbouwman:


Izaakwltn (approved):
Are we sure this is should be the canonical spelling of `indiecies`?

PR #1147, opened 2024-07-08
add faux macros for iteration syntax

### LINKED ISSUE #1143: faux lisp macros don't exist for looping forms
things like `for` etc. don't exist as faux lisp macros, and so they don't complete in the minibuffer.

stylewarning:
fixes #1143


stylewarning:
A small deficiency is that it won't deal with loop labels quite right. `trivial-indent` might help, but the documentation isn't so good and I couldn't figure it out. So I'll punt it for a later time.

eliaslfox (approved):

PR #1150, opened 2024-07-09
coalton-codegen-ast: return a value, like coalton-codegen

jbouwman:
Arising from discussion with a Coalton User who asked why codegen-ast doesn't return a value for use in further debugging: have it return a value, like codegen does.


eliaslfox (changes requested):
The purpose of `coalton-codegen-ast` is to be a debugging tool. It's not intended to be a public interface.

PR #1155, opened 2024-07-11
Roundtrip a hashtable-laden environment struct to trip up CCL and ACL

jbouwman:
Test case related to #1139


jbouwman:
(this branch served to exercise CCL and ACL while testing readable hash tables)

Issue #1151, opened 2024-07-10
Add instances `(Into Seq List)` and `(Into Seq Vector)`


Izaakwltn:

Izaakwltn:
closed by #1156

PR #1156, opened 2024-07-11
Into implementation for Sequence type

garlic0x1:
Solves https://github.com/coalton-lang/coalton/issues/1151

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -253,6 +253,14 @@ a new `Seq` instance."
                             (iter:zip! (iter:into-iter a)
                                        (iter:into-iter b)))))))
 
+  (define-instance (Into (Seq :a) (List :a))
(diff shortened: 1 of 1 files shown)


library/seq.lisp
```diff
@@ -253,6 +253,14 @@ a new `Seq` instance."
                             (iter:zip! (iter:into-iter a)
                                        (iter:into-iter b)))))))
 
+  (define-instance (Into (Seq :a) (List :a))
+    (define (into lst)
```
Izaakwltn, line 257: One small nit- the parameter should probably be named `seq` or similar since the input isn't a list, or a vector in the following instance.

Izaakwltn (approved):
Looks good, just one small nit!

eliaslfox (approved):

Issue #1145, opened 2024-07-08
add option to print types of values compiled in COALTON-TOPLEVEL


stylewarning:
It would be nice for debugging to see what the compiler derived/stored for the types of things defined in COALTON-TOPLEVEL. Maybe we can piggyback on `*compile-verbose*`.

stylewarning:
More specifically, while Coalton is being compiled, if `cl:*compile-print*` is true, then print out each compiled value's name and type while compiling. 

CLHS definition of `*compile-print*`:

> If [the variable] is true, information about top level forms in the file being compiled is printed to standard output. Exactly what is printed is implementation-dependent, but nevertheless some information is printed. If [the variable] is nil, no information is printed.

We could also just have our own variable too... it's probably a better idea. Maybe `*compile-print-types*`.

PR #1164, opened 2024-07-15
when settings:*compile-print-types* print definition types to stdout

garlic0x1:
https://github.com/coalton-lang/coalton/issues/1145

### DIFF
--- a/src/codegen/program.lisp
+++ b/src/codegen/program.lisp
@@ -165,6 +165,14 @@ A function bound here will be called with a keyword category, and one or more ad
                       ,name
                       ,(codegen-expression node nil env)))
 
+  ;; Print types of definitions
+  (when *compile-print* 
+    (dolist (binding bindings)
+      (let ((name (car binding)))
+        (format t "NAME: ~a, TYPE: ~a~%"
+                name
+                (tc:lookup-value-type env name)))))
+
   ;; Docstrings
   (loop :for (name . node) :in bindings
         :for entry := (tc:lookup-name env name :no-error t)
--- a/src/codegen/program.lisp
+++ b/src/codegen/program.lisp
@@ -166,10 +166,10 @@ A function bound here will be called with a keyword category, and one or more ad
                       ,(codegen-expression node nil env)))
 
   ;; Print types of definitions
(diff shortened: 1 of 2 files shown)


src/settings.lisp
```diff
@@ -55,6 +56,10 @@ Enable release mode either by setting the UNIX environment variable COALTON_ENV
 (declaim (type boolean *emit-type-annotations*))
 (defvar *emit-type-annotations* t)
 
+(declaim (type boolean *compile-print-types*))
+(defvar *compile-print-types* nil
+  "Print types of definitions to stdout on compile.")
```
stylewarning, line 61: nit: "standard output" and not the UNIXism 🤢 "stdout"

src/codegen/program.lisp
```diff
@@ -165,6 +165,14 @@ A function bound here will be called with a keyword category, and one or more ad
                       ,name
                       ,(codegen-expression node nil env)))
 
+  ;; Print types of definitions
+  (when (or *compile-print* settings:*compile-print-types*) 
+    (dolist (binding bindings)
```
stylewarning, line 170: Do we want to print the package prefix of the symbol being defined? Probably not...

stylewarning (approved):
I have one very tiny nit.

eliaslfox:
It looks like `*compile-print*` is `t` when compiling with `C-c C-c` in emacs. Was the intent of this pr that types should always be printed when developing locally?

stylewarning:
> It looks like `*compile-print*` is `t` when compiling with `C-c C-c` in emacs. Was the intent of this pr that types should always be printed when developing locally?
> 
> 
> 
> cc @stylewarning

It was more to respect the CL standard variable (allowing the user to turn it off through the CL standard variable). 

Personally I've been finding the feature extremely useful when doing development.

garlic0x1:
> It looks like `*compile-print*` is `t` when compiling with `C-c C-c` in emacs. Was the intent of this pr that types should always be printed when developing locally?
> 
> cc @stylewarning

Interesting, in Sly it is nil, I assume you are using Slime?

PR #1170, opened 2024-07-17
Adds `as` macro

Izaakwltn:
This adds the `as` macro, changing:

```
(the <type> (into <expr>))
```

into:

```
(as <type> <expr>)
```

### DIFF
--- a/docs/intro-to-coalton.md
+++ b/docs/intro-to-coalton.md
@@ -620,6 +620,41 @@ Type declarations can also be added in let expressions
       (g a b))))
 ```
 
+### Type Casting, Coercing, and Conversion
+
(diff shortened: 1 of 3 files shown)


stylewarning (approved):

stylewarning:
Would like @eliaslfox to take a look. Obviously like other things this would be better as a bonafide language feature, but it's useful enough now that I think it warrants inclusion.

stylewarning:
@macrologist We are doing this for many stdlib stuff now. It has its pros and cons.

eliaslfox (approved):
I'm not really a fan of defining this, especially defining it as a macro. In my experience `into` calls rarely need type annotations. This is also a deviation from Rust, which uses `as` only for simple compiler built in conversions. Other languages with built in conversion syntax like C also only use it for simple compiler defined conversions.

The code looks fine though.

stylewarning:
@eliaslfox As a compromise, I will do a little more due diligence on existing code to determine how frequent the pattern occurs.

Issue #1144, opened 2024-07-08
make a nicer printed representation for `Seq`


stylewarning:
The printed representation of `Seq` is low-level and not useful. Instead, at minimum, have it print as

```
#.(seq:make 1 2 3)
```

or something.

PR #1172, opened 2024-07-18
print-object method for seq:seq

garlic0x1:
Solves https://github.com/coalton-lang/coalton/issues/1144

```lisp
COALTON-USER> (coalton (coalton-library/seq:make 1 2 3))
#.(SEQ 1 2 3)
```

To avoid the methods defined [here](https://github.com/coalton-lang/coalton/blob/main/src/codegen/codegen-type-definition.lisp#L90) being used, I short-circuited them by using an `:around` method.

Regarding the suggested output `#.(seq:make 1 2 3)`, I replaced `seq:make` with just `seq` to match the implementations for Queue and Cell.

There will be nits, I will resolve them tomorrow :)

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -562,6 +562,15 @@ It attempts to rebalance with a minimum of array copying."
                           (cl:rest leaf-arrays)
                           :initial-value `(LeafArray ,(cl:funcall make-ary (cl:first leaf-arrays))))))))))
 
+;; This method implementation uses :around because sum types implement
+;; cl:print-object for each representation, to avoid a brittle design
+;; by overwriting them manually, the :around method short-circuits them.
+(cl:defmethod cl:print-object :around ((self seq) stream)
+  (cl:print-unreadable-object (self stream :type cl:nil)
+    (cl:format stream "SEQ~{ ~A~}"
+               (coalton ((the ((Seq :a) -> (List :a)) (fn (seq) (into seq)))
+                         (lisp (Seq :a) () self)))))
+  self)
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/SEQ")


library/seq.lisp
```diff
@@ -562,6 +562,11 @@ It attempts to rebalance with a minimum of array copying."
                           (cl:rest leaf-arrays)
                           :initial-value `(LeafArray ,(cl:funcall make-ary (cl:first leaf-arrays))))))))))
 
+(cl:defmethod cl:print-object :around ((self seq) stream)
+  (cl:format stream "#.(SEQ~{ ~A~})"
```
stylewarning, line 566: I would print this as an unreadable object since it won't be readable outside of a Coalton context (unlike the other examples).

    (defmethod ...
     (print-unreadable-object (... :type nil :identity nil)
        (prin1 'seq) ; print package as necessary
        (format stream "~{ ..." ...)))

stylewarning (changes requested):

library/seq.lisp
garlic0x1: `define-type` does not implement the `print-object` method for the `seq` class, it implements it for both subclasses LeafArray and RelaxedNode, if I just define the method for the `seq` superclass, the "more specific" subclass method implementations are used.  There may be some CLOS trick I don't know about to make it use the more generic method here.

I will make it print unreadably, but why is `#.(CELL 1)` supposed to work?  I can't get that to do anything valid in my REPL in either CL or Coalton.  Also would using the originally suggested `#.(SEQ:MAKE 1 2 3)` satisfy the readability issue?

garlic0x1:
Now it prints like this:
```lisp
COALTON-USER> (coalton (coalton-library/seq:make 1 2))
#<SEQ 1 2>
```

library/seq.lisp
```diff
@@ -562,6 +562,12 @@ It attempts to rebalance with a minimum of array copying."
                           (cl:rest leaf-arrays)
                           :initial-value `(LeafArray ,(cl:funcall make-ary (cl:first leaf-arrays))))))))))
 
+(cl:defmethod cl:print-object :around ((self seq) stream)
```
stylewarning, line 565: can you write a brief comment describing that this is a bit of a hack because of blah blah blah... (why :around, why not a directly on seq, etc)

stylewarning (changes requested):
looks great, just one request

stylewarning (approved):

stylewarning:
squash it all, and i'll merge.

great contribution and sleuthing! thanks!

PR #1179, opened 2024-07-22
bugfix for `apply-ast-substitution` on `node-lisp` and other pre-inlining changes

amorphedstar:
`apply-ast-substitution` would previously return `node-lisp` nodes unchanged, but if the node captures Coalton variables which must be substituted, then the application will need to generate a `node-let` wrapping the `node-lisp`.

The codegen will generate code that violates package locks if `node-lisp` nodes capture variable names which match existing function names and are then inlined into other packages. Accordingly, I renamed some arguments in `slice.lisp` and `vector.lisp`.

The `codegen-let` recursive data constructor requires `node-direct-application`s, but didn't explicitly check for them, which caused errors when inlining.

There are also a few other minor cleanup changes from browsing through the code.


src/typechecker/collect.lisp
```diff
@@ -1,226 +0,0 @@
-;;;;
-;;;; Generic traversals for collecting lists of items referenced from
-;;;; within a nested AST structure.
-;;;;
-
-(defpackage #:coalton-impl/parser/collect
```
amorphedstar, line 6: This file still thinks it's in the parser, and @eliaslfox verified it's fine to delete

src/codegen/codegen-expression.lisp
```diff
@@ -289,7 +289,8 @@
 
 (defun data-letrec-able-p (initform env)
   (let ((ctor-ent (find-constructor initform env)))
-    (and ctor-ent
+    (and (node-direct-application-p initform)
```
amorphedstar, line 292: The implementation of recursive data requires a direct application:
https://github.com/coalton-lang/coalton/commit/11968b8180aaacc8710a76e91e1e48bfe7a7232e

src/codegen/ast-substitutions.lisp
```diff
@@ -38,9 +38,31 @@
 
   (:method (subs (node node-lisp))
     (declare (type ast-substitution-list subs)
-             (ignore subs)
              (values node))
-    node)
+    (multiple-value-bind (let-bindings lisp-var-bindings)
```
amorphedstar, line 42: The bindings generated here are very similar to the inlining case
https://github.com/coalton-lang/coalton/pull/1029/files#diff-bfb68d1f5b43842fcc3cf2f8161617a2e306044bcb45ae3a30d6c2a030175857R341

tests/float-tests.lisp
```diff
@@ -99,7 +99,7 @@
   (declare float-checklist ((math:Dividable Integer :a) => (List :a)))
   (define float-checklist
     (coalton-prelude:zipWith
-     math:general/ (coalton:the (coalton:List coalton:Integer) (coalton-prelude:range -100 100)) (coalton-prelude:range 200 1))))
+     math:general/ (coalton:the (coalton:List coalton:Integer) (coalton-prelude:range -100 99)) (coalton-prelude:range 200 1))))
```
eliaslfox, line 102: ?

library/math/bounded.lisp
```diff
@@ -59,10 +59,10 @@
   (define-instance (Bounded IFix)
     (define minBound
       (lisp IFix ()
-        cl:most-positive-fixnum))
+        cl:most-negative-fixnum))
```
eliaslfox, line 62: oof

eliaslfox (approved):

tests/float-tests.lisp
amorphedstar: This doesn't change behavior (since `zipWith` stops once it reaches the end of either iterator), but just makes more explicit that the zipped list goes from (-100 200) to (99 1).
(I had just stumbled on this when trying to debug the test, which ended up just being a problem with Windows and libraries.)

library/math/bounded.lisp
stylewarning: lol

stylewarning (approved):

PR #1180, opened 2024-07-22
Native standard library

jbouwman:
Provisional conversion of standard library from .lisp to .coal files.


stylewarning:
I think this is now pretty out of date and we should consider this again in the future.

PR #1183, opened 2024-07-24
package hygiene

jbouwman:
- remove duplicate exports in library/classes.lisp
- correct symbol type labels in codegen/ast.lisp
- drop unused and misspelled exports in typechecker/types.lisp
- remove orphaned ty-binding-list type declaration

The machine integer types exported there may have been intended for further context reduction steps: just a guess.  Removing on the principle of minimizing hardcoded type information in the compiler.

### DIFF
--- a/library/classes.lisp
+++ b/library/classes.lisp
@@ -4,11 +4,6 @@
   (:local-nicknames
    (#:types #:coalton-library/types))
   (:export
-   #:Addressable #:eq?)
-  (:export
-   #:Eq #:==
-   #:Num #:+ #:- #:* #:fromInt)
-  (:export
    #:Signalable
    #:error
    #:Tuple
@@ -35,8 +30,6 @@
    #:Into
    #:TryInto
    #:Iso
-   #:Signalable
-   #:error
    #:Unwrappable #:unwrap-or-else #:with-default #:unwrap #:expect #:as-optional
    #:default #:defaulting-unwrap))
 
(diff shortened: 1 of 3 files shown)


src/typechecker/types.lisp
```diff
@@ -217,17 +199,7 @@
 (defvar *boolean-type*      (make-tycon :name 'coalton:Boolean      :kind +kstar+))
 (defvar *unit-type*         (make-tycon :name 'coalton:Unit         :kind +kstar+))
 (defvar *char-type*         (make-tycon :name 'coalton:Char         :kind +kstar+))
-(defvar *u8-type*           (make-tycon :name 'coalton:U8           :kind +kstar+))
```
stylewarning, line 220: Were these variables just unused?

Izaakwltn: From a cursory search, I don't see them ever being used

library/classes.lisp
```diff
@@ -35,8 +30,6 @@
    #:Into
    #:TryInto
    #:Iso
-   #:Signalable
```
Izaakwltn, line 38: That was me, I think. whoops!

Izaakwltn (approved):

src/typechecker/types.lisp
jbouwman: Yes, unused. these styles wired-in type constructors are used in two places: in the manual definition of early types in typechecker/environment.lisp and in typechecker/context-reduction.lisp, to resolve ambiguous lisp literals -- other access is conventional, through the environment.

Many early types, notably char and string, look like they could be ejected without much effort from the compiler, and defined directly in the standard library.

PR #1184, opened 2024-07-26
Refactor traverse

amorphedstar:
This PR aims to refactor the codegen's `traverse` function in order to provide a more general structure that allows
- operations performed uniformly on every node, and in particular, type variable collection and substitution for inlining multiple copies of polymorphic functions (this can also be used for syntactic inlining heuristics like node count), and
- arbitrary data passing through the traversal, not just a fixed `bound-variables` list, to allow for various applications like constant propagation ( see https://github.com/coalton-lang/coalton/pull/1029/commits/8d9c0995d7eb03cd84a2163dae62af57126b1e01#diff-d5c2bd51ab00b09becf60d0d656f35fd12dec0c8e7e0fb377f9eabfebcfe3b99R155-R156 ) and so that immutable data can be used with `traverse`, instead of requiring mutation like the `:before-*` `push` and `:after-*` `pop` used in several applications of `traverse`.

The traversal still assumes the result is a `node`, which means that collecting other data out of an AST node tree still requires either mutation with `traverse`, or a completely different traversal function. This is fine, since the main use of `traverse` is to avoid rewriting methods just to recursively propagate an operation through the AST when only a few nodes need custom behavior. Defaulting to recursive `traverse` calls only makes sense when the result is a node.

This PR also refactors `apply-ast-substitution` to use `traverse`. It creates a specialized `traverse-with-binding-list` for the three usages of `traverse` that actually used the `bound-variables`. And it implements type variable collection and substitution in ASTs.


src/codegen/transformations.lisp
```diff
@@ -310,6 +310,43 @@
             (append funs binding-list-traversals)
             (list :bound-variables nil)))
 
+(defmethod tc:apply-substitution (subs (node node))
```
amorphedstar, line 313: Refactoring `tc:apply-substitution` like this breaks `(monomorphize)`, but I'm not sure why...
```
; caught COMMON-LISP:ERROR:
;   COMMON-LISP:READ error during COMMON-LISP:COMPILE-FILE:
;   
;     The value
;       #S(COALTON-IMPL/CODEGEN/AST:NODE-LITERAL :TYPE INTEGER :VALUE 2)
;     is not of type
;       COALTON-IMPL/CODEGEN/AST:NODE-VARIABLE
;   
;     (in form starting at line: 5, column: 0, position: 86)
```

coalton.asd
```diff
@@ -120,11 +120,11 @@
                 :serial t
                 :components ((:file "pattern")
                              (:file "ast")
+                             (:file "transformations")
```
amorphedstar, line 123: `ast-substitutions` now uses `traverse`

src/codegen/transformations.lisp
```diff
@@ -11,204 +11,181 @@
...
+           (type list   args)
+           (type node   node)
+           (values node &optional))
+  (or
+   (alexandria:when-let ((f (cdr (assoc key funs))))
+     (apply f (cons node args)))
+   node))
 
-(defun call-if (node key funs bound-variables)
+(defmacro traverse-if (key funs args before-node else-expr)
```
amorphedstar, line 32: Is a macro a good way to allow a default traversal expression which is only evaluated if a custom traversal isn't specified, or is it better to pass a `lambda` or something?

src/codegen/transformations.lisp
```diff
@@ -246,3 +223,171 @@
...
+              :body (traverse (node-bind-body node) funs new-args)))))))
+
+(defun traverse-with-binding-list (node funs)
+  (declare (type node node)
+           (type list funs)
+           (values node &optional))
+  (traverse node
+            (append funs binding-list-traversals)
+            (list :bound-variables nil)))
+
```
amorphedstar, line 312: Is there a nicer way to specify these `bound-variables`-specific traversal functions?

stylewarning (commented):
I think the abstraction presented here is quite bizarre and we should see if we can get away with using CLOS methods instead of keyword dispatch. Happy to discuss.

src/codegen/ast.lisp
```diff
@@ -112,8 +113,19 @@
 ;;;
 
 
-(defstruct (node (:constructor nil))
-  (type (util:required 'type) :type tc:ty :read-only t))
+(defstruct (node (:constructor nil)
```
stylewarning, line 116: this is going to export (setf node-type). My recommendation is to avoid doing that, personally, as it's not allowed by the API. But it means we'll need the %type shenanigans, or a different conc-name.

amorphedstar: Yeah, I ended up switching it back from `%type` to `type` because there are like 1000 places in the code where node structs are created using `:type`. Do you know a nice way of handling that? I tried a few things and ended up writing another stupid 18-case generic function to allow `copy-node` with editing some fields, without changing any of the prior AST node struct definitions.

library/builtin.lisp
```diff
@@ -54,7 +54,7 @@
 
   (declare boolean-and (Boolean -> Boolean -> Boolean))
   (define (boolean-and x y)
-    "Are both `x` and `y` true? Note that this is a *function* which means both `x` and `y` will be evaluated. Use the AND macro for short-circuiting behavior."
+    "Are both `x` and `y` true? Note that this is a *function* which means both `x` and `y` will be evaluated. Use the `and` macro for short-circuiting behavior."
```
amorphedstar, line 57: (for consistency with `or` on line 50 ^^)

src/codegen/ast-substitutions.lisp
```diff
@@ -25,150 +29,60 @@
 (deftype ast-substitution-list ()
   '(satisfies ast-substitution-list-p))
 
-(defgeneric apply-ast-substitution (subs node)
```
amorphedstar, line 28: I replaced this explicit generic with a call to `traverse`, emphasizing the four cases in which something actually gets done

src/codegen/ast.lisp
```diff
@@ -428,97 +430,111 @@ both CL namespaces appearing in NODE"
      (node-variables-g (node-bind-expr node) :variable-namespace-only variable-namespace-only)
      (node-variables-g (node-bind-body node) :variable-namespace-only variable-namespace-only))))
 
-(defmethod tc:apply-substitution (subs (node node-literal))
```
amorphedstar, line 431: I moved this method definition to the `transformations.lisp` file, for all `node`s at once using `traverse`

src/codegen/ast.lisp
```diff
@@ -428,97 +430,111 @@ both CL namespaces appearing in NODE"
...
+     :type (or type (node-type node))
+     :expr (or expr (node-return-expr node))))
+
+  (:method ((node node-field) &key type name dict)
+    (make-node-field
+     :type (or type (node-type node))
+     :name (or name (node-field-name node))
+     :dict (or dict (node-field-dict node))))
+
+  (:method ((main-node node-dynamic-extent) &key type name node body)
```
amorphedstar, line 528: :(

src/codegen/transformations.lisp
```diff
@@ -11,204 +11,159 @@
...
+           (values node &optional))
+  (or
+   (alexandria:when-let ((f (cdr (assoc key funs))))
+     (apply f (cons node args)))
+   node))
 
-(defun call-if (node key funs bound-variables)
+(defmacro traverse-if (key funs args before-node else-expr)
+  (let ((before-key
+          (alexandria:make-keyword (concatenate 'string "BEFORE-" (symbol-name (cadr key)))))
```
amorphedstar, line 34: If we quote the keywords, then the macro needs their symbol from `cadr`

src/codegen/transformations.lisp
```diff
@@ -246,3 +201,169 @@
     (fset:do-map (name entry (immutable-map-data (tc:environment-function-environment env)))
       (setf (gethash name table) (tc:function-env-entry-arity entry)))
     table))
+
+(defvar binding-list-traversals
```
amorphedstar, line 205: This is a list of the five cases which got special treatment inside of `traverse` previously, where `bound-variables` gets updated and passed, instead of passed through the recursion with no change. Is there a better way of doing this than `defvar`? (Lisp seemed to not like it when I tried to define it as a constant...) Should I inline this into `traverse-binding-list` below?

src/codegen/transformations.lisp
```diff
@@ -246,3 +201,169 @@
...
+                                    (make-match-branch
+                                     :pattern (tc:apply-substitution subs (match-branch-pattern branch))
+                                     :body (match-branch-body branch)))
+                                  (node-match-branches node)))))
+    (cons ':after-while-let
+          (lambda (node)
+            (declare (type node-while-let node)
+                     (values node-while-let &optional))
+            (copy-node node
+                       :pattern (tc:apply-substitution subs (node-while-let-pattern node))))))))
```
amorphedstar, line 317: @eliaslfox mentioned that `pattern`s have types, which should be substituted, too

src/codegen/transformations.lisp
```diff
@@ -246,3 +201,169 @@
...
+
+(defmethod tc:type-variables ((node node))
```
amorphedstar, line 319: This will collect all type variables from `node`s and `pattern`s in the tree of `node`. (Should I include something like this as a docstring in the code, or is it too obvious?)

src/codegen/ast.lisp
stylewarning: Yes, add

    (:conc-name %node-)

just before the `(:constructor nil)` clause so that the accessor becomes

    %node-type

and

    (setf %node-type)

Define your own

    node-type

reader based off of this. The `:node` keyword argument will still work.

src/codegen/ast.lisp
stylewarning: Can you add a short comment in the code?

    ;;; tc:apply-substitutions used to be here, but ...

src/codegen/ast.lisp
```diff
@@ -428,97 +430,111 @@ both CL namespaces appearing in NODE"
...
+    (make-node-match
+     :type     (or type     (node-type node))
+     :expr     (or expr     (node-match-expr node))
+     :branches (or branches (node-match-branches node))))
+
+  (:method ((node node-while) &key type label expr body)
+    (make-node-while
+     :type  (or type  (node-type node))
+     :label (or label (node-while-label node))
+     :expr  (or expr  (traverse (node-while-expr node) funs args))
```
stylewarning, line 485: is the traverse supposed to be here?

src/codegen/ast.lisp
```diff
@@ -428,97 +430,111 @@ both CL namespaces appearing in NODE"
...
-   :node (tc:apply-substitution subs (node-dynamic-extent-node node))
-   :body (tc:apply-substitution subs (node-dynamic-extent-body node))))
-
-(defmethod tc:apply-substitution (subs (node node-bind))
-  (make-node-bind
-   :type (tc:apply-substitution subs (node-type node))
-   :name (node-bind-name node)
-   :expr (tc:apply-substitution subs (node-bind-expr node))
-   :body (tc:apply-substitution subs (node-bind-body node))))
+(defgeneric copy-node (node &key)
```
stylewarning, line 433: I think this will have subtle bugs if someone passes in nil. You'll need to do the laborious thing of using suppliedp arguments...

src/codegen/transformations.lisp
stylewarning: In this case the keywords are syntax, not data.

src/codegen/transformations.lisp
stylewarning: I think this macro needs revision, but it would be better to discuss it.

src/codegen/transformations.lisp
stylewarning:
```
(defun make-binding-list-traversals ()
    (load-time-value
      (list ...)
      t))
```

src/codegen/transformations.lisp
stylewarning: does this need to be a method?

and yes, lean toward writing docs.

src/codegen/transformations.lisp
```diff
@@ -246,3 +201,169 @@
...
+           (values node &optional))
+  (alexandria:if-let ((old-type-variables (tc:type-variables node)))
+    (tc:apply-substitution
+     (mapcar (lambda (tyvar)
+               (tc:make-substitution
+                :from tyvar
+                :to (tc:make-variable (tc:kind-of tyvar))))
+             old-type-variables)
+     node)
+    node))
```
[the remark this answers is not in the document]
stylewarning: how about

    (defun fresh-type-renamer (tyvar)
      (make-variable (kind-of tyvar)))

    (defun rename-type-variables (renamer node)
      "Rename the type variables according to the function RENAMER ..."
      ...)

then the caller can do

    (rename-type-variables #'fresh-type-renamer node)

Just an idea.

stylewarning (changes requested):

src/codegen/transformations.lisp
amorphedstar: Ok, for now I reverted the keywords being passed to this macro to unquoted

src/codegen/transformations.lisp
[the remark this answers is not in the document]
amorphedstar: Yeah, that's pretty nice! I decided to swap the node and renamer so that `renamer` can be `optional`, defaulting to `#'tc:fresh-type-renamer`

src/codegen/transformations.lisp
amorphedstar: `tc:type-variables` was already a method for collecting type variables out of `tc:ty`, accessor, node, etc. Of course, we could use a different name for a function which collects type variables out of the AST nodes, in which case it wouldn't need to be a method

src/codegen/ast.lisp
amorphedstar: no haha

src/codegen/ast.lisp
amorphedstar: yeah, I checked that even `node-literal-value` can't be a boolean, so I think all the cases should be fine, but the type updating is all i _really_ wanted, so I went ahead and reverted and stashed this generic, going back to the `copy-node (node &optional type)` we had previously

src/codegen/transformations.lisp
```diff
@@ -11,204 +11,191 @@
...
-              :body (traverse (node-bind-body node) funs new-bound-variables))))
-      (call-if node :bind funs bound-variables))))
+(defun traverse (initial-node custom-funs &optional (initial-args nil))
+  "Create a recursive \"generic\" traversal with pre-, in-, and post-operations fixed by `funs` (with the default traversals as a fallback), and evaluate this traversal with `initial-node` and `initial-args`."
+  (declare (type node initial-node)
+           (type list custom-funs)
+           (type list initial-args)
+           (values node &optional))
+  (let ((funs (append custom-funs (make-default-traversals))))
+    (labels ((traverse (current-node args)
```
amorphedstar, line 187: The original `defgeneric traverse` is now entirely gone. Both `funs` and the `traverse` actions are now entirely abstracted to a common alist associating `(cons (or :before :traverse :after) 'coalton-impl/codegen/ast:node*)` to function objects. Instead of passing `funs` everywhere to characterize the traversal, now the fixpoint `traverse` function itself gets passed. Instead of abstracting away the particular `bound-variables` logic and gluing the traversal to itself with some wrapping function, now only the glue remains.

src/codegen/transformations.lisp
```diff
@@ -37,6 +37,28 @@
   (type     'node                     :type node-subtype :read-only t)
   (function (util:required 'function) :type function     :read-only t))
 
+(defmacro action ((when type &rest args) &body body)
```
amorphedstar, line 40: This macro seems to generate lots of warnings about the node class names being undefined...

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+  '(member :before :after :traverse))
+
+(defstruct (action (:constructor make-action (when type function)))
+  "Container for a function to call on nodes of a particular type - before, after, or instead of the recursive call in the traversal"
+  (when     ':after                   :type when-keyword :read-only t)
+  (type     'node                     :type node-subtype :read-only t)
+  (function (util:required 'function) :type function     :read-only t))
+
+(defmacro action ((when type &rest args) &body body)
+  "Construct an action block, turning `body` into a lambda function accepting `args` if provided"
```
stylewarning, line 41: Since macros are new syntax, you must write something comprehensive about the syntax in the doc string.

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+  (check-type when         keyword)
+  (check-type type         symbol)
+  (assert (subtypep type 'node))
```
stylewarning, line 44: node-subtype-p

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+  (cond
+    ((not args)
+     (assert (= 1 (length body)))
+     `(make-action ',when ',type ,@body))
+    ((equal ':traverse when)
```
stylewarning, line 49: eq

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+  (type     'node                     :type node-subtype :read-only t)
+  (function (util:required 'function) :type function     :read-only t))
+
+(defmacro action ((when type &rest args) &body body)
+  "Construct an action block, turning `body` into a lambda function accepting `args` if provided"
+  (check-type when         keyword)
+  (check-type type         symbol)
+  (assert (subtypep type 'node))
+  (cond
+    ((not args)
```
stylewarning, line 46: endp

- not for booleans
- endp for lists 
- null for nil

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+     (assert (= 1 (length body)))
+     `(make-action ',when ',type ,@body))
+    ((equal ':traverse when)
+     (assert (< 1 (length args)))
+     `(make-action ',when ',type
+                   (lambda (,@args)
+                     (declare (type ,(car args) ,type)
```
stylewarning, line 53: car cadr-> first second

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+                              (type ,(cadr args) function))
+                     ,@body)))
+    (t
+     `(make-action ',when ',type
+                   (lambda (,@args)
+                     (declare (type ,(car args) ,type))
```
stylewarning, line 59: first

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+           (type list         args)
+           (type node         node)
+           (values node &optional))
+  (or
+   (alexandria:when-let*
+       ((action (find-if (lambda (action)
+                           (declare (type action action)
+                                    (values boolean &optional))
+                           (and (equal when-key (action-when action))
+                                (equal type-key (action-type action))))
```
stylewarning, line 177: eq

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+                         actions))
+        (f (action-function action)))
+     (if (equal ':traverse when-key)
```
stylewarning, line 180: eq

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+       :body (apply traverse (node-dynamic-extent-body node) args)))
+    (action (:traverse node-bind node traverse &rest args)
+      (make-node-bind
+       :type (node-type node)
+       :name (node-bind-name node)
+       :expr (apply traverse (node-bind-expr node) args)
+       :body (apply traverse (node-bind-body node) args))))
+   t))
+
+(defun optional-call (when-key type-key actions traverse args node)
```
stylewarning, line 162: maybe now we can use the "action" terminology now? `fire-actions` `fire-matching-actions` maybe something better

src/codegen/transformations.lisp
```diff
@@ -246,3 +238,173 @@
     (fset:do-map (name entry (immutable-map-data (tc:environment-function-environment env)))
       (setf (gethash name table) (tc:function-env-entry-arity entry)))
     table))
+
+(defun make-binding-list-traversals ()
```
stylewarning, line 242: change name to reflect actions?

src/codegen/transformations.lisp
```diff
@@ -246,3 +238,173 @@
...
+   t))
+
+(defun traverse-with-binding-list (node actions)
+  "Traverse `node` while keeping track of a list of `bound-variables` which gets passed as a keyword argument whenever a function from `actions` gets called."
+  (declare (type node        node)
+           (type action-list actions)
+           (values node &optional))
+  (traverse node
+            (append actions (make-binding-list-traversals))
+            ':bound-variables nil))
```
stylewarning, line 304: if this is immutable: `'(:bound-variables nil)`

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
+        (f (action-function action)))
+     (if (equal ':traverse when-key)
+         (apply f node traverse args)
+         (apply f node args)))
+   node))
 
-(defun call-if (node key funs bound-variables)
-  (declare (type node node)
-           (type symbol key)
+(defun traverse (initial-node custom-actions &rest initial-args)
```
stylewarning, line 185: we absolutely need to write a very helpful and perhaps even comprehensive explainer on TRAVERSE, in the doc string or in a comment. We need to show how to use it, with some dead simple examples, so a new coalton contributor can understand it quickly. The traverse+actions is a powerful analysis workhorse, and deserves a lot more than a couple emaciated docstrings.

Depending on how comprehensive, it can even be a .md in docs with a link to it here.

src/codegen/transformations.lisp
```diff
@@ -7,208 +7,200 @@
...
 
-(defun traverse-bindings (bindings funs)
-  (declare (type binding-list bindings)
-           (values binding-list))
-  (loop :for (name . node) :in bindings
-        :collect (cons name (traverse node funs nil))))
+(defun node-subtype-p (symbol)
+  (and (symbolp symbol) (subtypep symbol 'node)))
+
+(deftype node-subtype ()
```
stylewarning, line 28: document these types, at least referencing ACTION and TRAVERSE.

stylewarning (changes requested):
this is coming together a *lot* more nicely now I think. Mostly left is finishing touches with documentation.

src/codegen/transformations.lisp
amorphedstar: I had thought to try that, but
```
; caught ERROR:
;   during macroexpansion of (ACTION (:AFTER NODE-LISP NODE) (ALEXANDRIA:UNIONF NODE-VARS #) ...). Use *BREAK-ON-SIGNALS* to intercept.
;    The function COALTON-IMPL/CODEGEN/TRANSFORMATIONS::NODE-SUBTYPE-P is undefined.It is defined earlier in the file but is not available at compile-time.
```
when
```suggestion
  (assert (node-subtype-p type))
```
... is there another way to get it to work?

src/codegen/transformations.lisp
amorphedstar: I updated `traverse` to spread `args` in every case, so any form with general `args` uses a list and `apply`. This means that a keyword argument like `bound-variables` can be put into `traverse` directly now, instead of needing to wrap it as a list.

Upon thinking more, the reason why using keyword arguments in the traversal would have been nice before is that different traversals might keep track of different data and only use some of it, but the main definition of `traverse` would have had to keep track of it all. Now that `traverse` passes arbitrary extra `args`, there can be no ambiguity, and anything using `traverse-binding-list` can be guaranteed that there is just the one arg `bound-variables`, so I changed it from a keyword argument to a regular argument.

src/codegen/traverse.lisp
```diff
@@ -0,0 +1,359 @@
...
+           (values node &optional))
+  (traverse node
+            (append actions (make-binding-list-actions))
+            nil))
+
+;;;
+;;; Examples
+;;;
+
+(defun count-nodes (node)
```
amorphedstar, line 329: I pulled the general `traverse` stuff out to its own file and added these examples which I hope could be seen as `dead simple`

src/codegen/optimizer.lisp
```diff
@@ -24,21 +27,57 @@
...
    (#:parser #:coalton-impl/parser)
    (#:tc #:coalton-impl/typechecker))
   (:export
+   #:make-function-table
    #:optimize-bindings
    #:optimize-node))
 
 (in-package #:coalton-impl/codegen/optimizer)
 
+(defun update-function-env (bindings env)
```
amorphedstar, line 48: I brought these two functions out of `transformations` to the place they are used, in `optimizer`, since they don't seem like transformations to me, but if that's wrong I can put them back.

src/codegen/traverse.lisp
stylewarning: skull emoji

stylewarning (approved):

PR #1186, opened 2024-07-30
Integer literal folding (part 2)

garlic0x1:
I have an ugly cond block that optimizes `translate-expression` on type `tc:node-integer-literal` for inferred types other than regular ints.

In the following example, `1` expands to `1.0` instead of `(from-int 1)`, essentially just doing that work at comptime in the expression translator.

```lisp
                (declare adder (Single-Float -> Single-Float))
                (define (adder x) (+ 1 x)))
(COMMON-LISP:PROGN
 (COALTON-IMPL/GLOBAL-LEXICAL:DEFINE-GLOBAL-LEXICAL ADDER
                                                    COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY)
 (COMMON-LISP:DEFUN ADDER (X-310)
   (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE X-310))
   (COALTON-LIBRARY/MATH/NUM::|INSTANCE/NUM SINGLE-FLOAT-COALTON-LIBRARY/CLASSES:+|
    1.0 X-310))
 (COMMON-LISP:SETF ADDER (COALTON-IMPL/RUNTIME/FUNCTION-ENTRY::F1 #'ADDER))
 (COMMON-LISP:SETF (COMMON-LISP:DOCUMENTATION 'ADDER 'COMMON-LISP:VARIABLE)
                     "ADDER :: (SINGLE-FLOAT → SINGLE-FLOAT)")
 (COMMON-LISP:SETF (COMMON-LISP:DOCUMENTATION 'ADDER 'COMMON-LISP:FUNCTION)
                     "ADDER :: (SINGLE-FLOAT → SINGLE-FLOAT)")
 (COMMON-LISP:VALUES))
```

Repeatedly checking `tc:tycon-p` is not great but I struggled to create clean control flow here...

### DIFF
--- a/src/codegen/translate-expression.lisp
+++ b/src/codegen/translate-expression.lisp
@@ -137,35 +137,52 @@ Returns a `node'.")
       (assert (null (tc:qualified-ty-predicates qual-ty)))
 
       (let* ((classes-package (util:find-package "COALTON-LIBRARY/CLASSES"))
-
              (num-class (util:find-symbol "NUM" classes-package))
-
              (num-pred (tc:make-ty-predicate :class num-class
                                              :types (list (tc:qualified-ty-type qual-ty))))
-
              (from-int-method (util:find-symbol "FROMINT" classes-package))
-
+             (val (tc:node-integer-literal-value expr))
              (ty (tc:qualified-ty-type qual-ty)))
+        (flet ((make-full-call ()
+                 (make-node-application
+                  :type (tc:qualified-ty-type qual-ty)
+                  :rator (make-node-variable
+                          :type (tc:make-function-type*
+                                 (list
+                                  (pred-type num-pred env)
+                                  tc:*integer-type*)
+                                 (tc:qualified-ty-type qual-ty))
+                          :value from-int-method)
+                  :rands (list
+                          (resolve-dict num-pred ctx env)
+                          (make-node-literal
+                           :type tc:*integer-type*
+                           :value val)))))
 
-        ;; Avoid casting INTEGER to INTEGER with FROMINT at runtime
-        (if (and (tc:tycon-p ty) (eq 'coalton:integer (tc:tycon-name ty)))
-            (make-node-literal
-             :type tc:*integer-type*
-             :value (tc:node-integer-literal-value expr))
-            (make-node-application
-             :type (tc:qualified-ty-type qual-ty)
-             :rator (make-node-variable
-                     :type (tc:make-function-type*
-                            (list
-                             (pred-type num-pred env)
-                             tc:*integer-type*)
-                            (tc:qualified-ty-type qual-ty))
-                     :value from-int-method)
-             :rands (list
-                     (resolve-dict num-pred ctx env)
-                     (make-node-literal
-                      :type tc:*integer-type*
-                      :value (tc:node-integer-literal-value expr))))))))
+          ;; Avoid casting INTEGER literals with FROMINT at runtime
+          ;; when the type is known, We can do that at comptime.  This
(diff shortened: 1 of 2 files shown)


src/codegen/translate-expression.lisp
```diff
@@ -145,27 +145,46 @@ Returns a `node'.")
...
-                             (pred-type num-pred env)
-                             tc:*integer-type*)
-                            (tc:qualified-ty-type qual-ty))
-                     :value from-int-method)
-             :rands (list
-                     (resolve-dict num-pred ctx env)
-                     (make-node-literal
-                      :type tc:*integer-type*
-                      :value (tc:node-integer-literal-value expr))))))))
+        (cond ((and (tc:tycon-p ty) (eq 'coalton:integer (tc:tycon-name ty)))
```
stylewarning, line 153: put the otherwise case in an flet called `make-full-call` and do something like

```
(if (not (tyconp ty))
    (make-full-call)
    (case (tycon-name ty)
      (coalton:integer
       (make-node-literal ...))
      ...
      (otherwise
       (make-full-call))))
```

src/codegen/translate-expression.lisp
```diff
@@ -145,27 +145,46 @@ Returns a `node'.")
...
+                :type tc:*ufix-type*
+                :value val))
+              ((and (tc:tycon-p ty) (eq 'coalton:ifix (tc:tycon-name ty)))
+               (make-node-literal
+                :type tc:*ifix-type*
+                :value val))
+              ((and (tc:tycon-p ty) (eq 'coalton:single-float (tc:tycon-name ty)))
+               (make-node-literal
+                :type tc:*single-float-type*
+                :value (float val)))
```
stylewarning, line 168: please use coerce here to be explicit

stylewarning (changes requested):
little things

stylewarning:
also figure out where `*ufix-type*` is because it's saying it's not in the package

also write a comment in the code that this should ideally be done by constant folding, but the benefit is too great to wait for

stylewarning:
@garlic0x1 CASE will be a lot nicer than COND. :)

src/codegen/translate-expression.lisp
```diff
@@ -137,35 +137,52 @@ Returns a `node'.")
...
+                           :value val)))))
+
+          ;; Avoid casting INTEGER literals with FROMINT at runtime
+          ;; when the type is known, We can do that at comptime.  This
+          ;; is a temporary hack to get an easy performance boost but
+          ;; should be solved later with proper constant folding.
+          (if (not (tc:tycon-p ty))
+              (make-full-call)
+              (case (tc:tycon-name ty)
+                (coalton:integer
```
stylewarning, line 169: nit: Coalton type names should be capitalized

stylewarning (approved):

src/codegen/translate-expression.lisp
```diff
@@ -137,35 +137,52 @@ Returns a `node'.")
...
+          ;; when the type is known, We can do that at comptime.  This
+          ;; is a temporary hack to get an easy performance boost but
+          ;; should be solved later with proper constant folding.
+          (if (not (tc:tycon-p ty))
+              (make-full-call)
+              (case (tc:tycon-name ty)
+                (coalton:Integer
+                 (make-node-literal :type tc:*integer-type*
+                                    :value val))
+                (coalton:Ifix
```
stylewarning, line 172: UFix and IFix 😵

stylewarning (approved):

PR #1190, opened 2024-08-05
Minor Library Additions

YarinHeffes:
This PR includes some functionality that I have found useful while programming with Coalton. Additionally, this PR does take and expand on the code in the `as-macro` branch ( https://github.com/coalton-lang/coalton/pull/1170 ) to imitate a currying capability.


library/classes.lisp
```diff
@@ -346,7 +346,12 @@ Typical `fail` continuations are:
     "Unwrap an `unwrappable`, returning `(default)` of the wrapped type on failure. "
     (unwrap-or-else (fn (elt) elt)
                     (fn () (default))
-                    container)))
+                    container))
+
+  (declare default? ((Default :a) (Eq :a) => :a -> Boolean))
+  (define (default? x)
+      "Is `x` the default item of its type?"
+      (== (default) x)))
```
stylewarning, line 354: nit: change arg order so parens stack

library/functions.lisp
```diff
@@ -44,6 +46,12 @@
       (lisp :a (str item) (cl:format cl:t "~A: ~A~%" str item))
       Unit))
 
+  (declare print ((Into :a String) => :a -> :a))
+  (define (print item)
```
stylewarning, line 50: might as well add `print-line` too

library/functions.lisp
```diff
@@ -44,6 +46,12 @@
...
+      "Print the String representation of an item to `cl:*standard-output*` and return the item."
```
stylewarning, line 51: nit:

```
"Print the string representation of `item` to the standard output, and return the item."
```

Maybe it should actually just return `Unit`? What do you think?

library/functions.lisp
```diff
@@ -111,6 +119,10 @@
     "Compute the complement of a unary Boolean function."
     (not (f x)))
 
+  (declare curry ((Tuple :left :right -> :result) -> :left -> :right -> :result))
+  (define (curry func left right)
+      (func (tuple left right)))
```
stylewarning, line 124: `Tuple`

indentation

library/list.lisp
```diff
@@ -142,6 +143,12 @@
     "Returns a list containing one element."
     (Cons x Nil))
 
+  (declare singleton? (List :a -> Boolean))
+  (define (singleton? xs)
```
stylewarning, line 147: can we have singleton for other sequence structures too?

library/list.lisp
```diff
@@ -142,6 +143,12 @@
...
+    "Returns `True` if `xs` is a list containing one element."
+    (and (not (null? xs))
```
stylewarning, line 149: use `match`

```
(match xs
  ((Cons _ (Nil)) True)
  (_ False))
```

library/string.lisp
```diff
@@ -39,6 +39,13 @@
 ;;; String
 ;;;
 
+(cl:eval-when (:compile-toplevel :load-toplevel)
+  (cl:defmacro define-instance-into-rational-string (type)
+    `(define-instance (Into ,type String)
+       (define (into z)
+         (lisp String (z)
+           (cl:format cl:nil "~D" z))))))
```
stylewarning, line 47: `~D` won't work for rationals I don't think (??); I forgot.

library/string.lisp
```diff
@@ -207,6 +227,13 @@ does not have that suffix."
                  (Err (concat "Cannot parse string as integer: " s))
                  (Ok z))))))
 
+  (define-instance ((Into :a String) (Into :b String) => (Into (Tuple :a :b) String))
```
stylewarning, line 230: I think this is probably a net negative, and could trip one up in debugging.

stylewarning (changes requested):

library/classes.lisp
YarinHeffes: Fixed.

library/functions.lisp
YarinHeffes: How should `print-line` differ from `print`?

library/functions.lisp
YarinHeffes: I made it return `Unit`, and added a function `and-return`, so that, for debugging purposes, one can write `(and-return print x)` for example. I also used `and-return` to implement `vector:singleton`.

library/functions.lisp
YarinHeffes: Fixed.

library/list.lisp
YarinHeffes: For now, I added this to `vector`.

library/list.lisp
YarinHeffes: Fixed.

library/string.lisp
YarinHeffes: I tested for a simple case for all the types which I implemented with this macro and it seemed to work just fine.

library/string.lisp
YarinHeffes: Removed.

library/functions.lisp
```diff
@@ -44,6 +48,17 @@
       (lisp :a (str item) (cl:format cl:t "~A: ~A~%" str item))
       Unit))
 
+  (declare print ((Into :a String) => :a -> Unit))
+  (define print
+    "Print the String representation of an item to `cl:*standard-output*`."
+    (compose trace into))
+
+  (declare and-return ((:a -> :b) -> :a -> :a))
+  (define (and-return func item)
```
stylewarning, line 57: Is there a use case for this? Not sure this is a useful enough utility to graduate to the standard library.

In any case, I'd probably call it `intercept` or something and reverse the argument order.

library/functions.lisp
```diff
@@ -111,12 +126,21 @@
     "Compute the complement of a unary Boolean function."
     (not (f x)))
 
+  (declare curry ((Tuple :left :right -> :result) -> :left -> :right -> :result))
+  (define (curry func left right)
+      (func (Tuple left right)))
```
stylewarning, line 131: Indentation weird.

library/functions.lisp
```diff
@@ -111,12 +126,21 @@
...
+
   (declare uncurry ((:left -> :right -> :result) -> Tuple :left :right -> :result))
   (define (uncurry func tpl)
     (match tpl
       ((Tuple left right)
        (func left right))))
 
+  (declare pair-with ((:left -> :right) -> :left -> Tuple :left :right))
+  (define (pair-with func left)
```
stylewarning, line 140: How about

    (pair-with left f right)

Also not sure this is the most useful, but I can see it.

library/string.lisp
stylewarning: CLHS 22.3.2.2 says `~D` "should" have an argument which is an integer.

But it does say that if it's not an integer, it'll fall back on `~A`. So your code should work, but it's just a little strange to see.

src/language-macros.lisp
stylewarning, line 1: Rebase on main.

stylewarning (changes requested):

library/functions.lisp
YarinHeffes: The main use case is really to implement a `debug-print`, so that, when testing code, once can just wrap an expression with `(and-return print (...))` or `(debug-print (...))` and the code will continue to work as expected. However, I'll remove this, since it is easily implemented by the user when needed.

library/functions.lisp
YarinHeffes: This is an extension of `iter:pair-with!` and can be used in general to make pairings `(x, f(x))`. This would allow any collection to do `(map (pair-with f) collection)`, including `iterator`, so `iter:pair-with!` could then be deprecated.

library/string.lisp
YarinHeffes: I changed this to `define-instance-into-integral-string` and I handled fraction separately with `"~A"` now.

library/functions.lisp
YarinHeffes: Fixed.

src/language-macros.lisp
YarinHeffes: Should be fixed.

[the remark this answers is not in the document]
YarinHeffes: I actually did, and then I made a change to the macro to allow functionality such as `(map (as String) (make-list 1 2 3))` for example. If you think this is a bad change, then I can remove it, but I certainly would find this functionality useful.

library/functions.lisp
```diff
@@ -111,12 +119,21 @@
...
+      "Create a `Tuple` from `left` paired with the its output through `func`."
```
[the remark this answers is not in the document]
YarinHeffes: fixed.

library/functions.lisp
```diff
@@ -44,6 +47,11 @@
       (lisp :a (str item) (cl:format cl:t "~A: ~A~%" str item))
       Unit))
 
+  (declare print ((Into :a String) => :a -> Unit))
+  (define print
```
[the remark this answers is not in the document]
YarinHeffes: Fixed.

stylewarning (approved):

PR #1193, opened 2024-08-06
Implement inlining with basic syntactic heuristics

amorphedstar:
This PR implements an inlining pass based on the work of @digikar99 ( https://github.com/coalton-lang/coalton/pull/1029 ). The updated `traverse` is used to allow passing a call stack, which enables separate bounds on recursive function unrolling and inlining depth. Heuristics are assumed to be simple syntactic functions of a function body for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,8 @@ jobs:
         LISP_IMPL: ["sbcl", "allegro", "ccl"]
         COALTON_ENV: ["release", "development"]
         SAFETY: ["0", "3"]
-        DISABLE_SPECIALIZATION: ["1", "0"] 
+        DISABLE_SPECIALIZATION: ["1", "0"]
+        HEURISTIC_INLINING: ["1", "0"]
         exclude:
           # Since we don't have a way of switching safety on allegro or ccl, we can
           # just ignore high safety tests.
@@ -47,6 +48,7 @@ jobs:
         env:
           COALTON_ENV: ${{ matrix.COALTON_ENV }}
           COALTON_DISABLE_SPECIALIZATION: ${{ matrix.DISABLE_SPECIALIZATION }}
+          COALTON_HEURISTIC_INLINING: ${{ matrix.HEURISTIC_INLINING }}
         run: |
           cat <<EOF > run-tests.lisp
           (sb-ext:restrict-compiler-policy 'safety ${{ matrix.SAFETY }})
@@ -61,6 +63,7 @@ jobs:
         env:
           COALTON_ENV: ${{ matrix.COALTON_ENV }}
           COALTON_DISABLE_SPECIALIZATION: ${{ matrix.DISABLE_SPECIALIZATION }}
+          COALTON_HEURISTIC_INLINING: ${{ matrix.HEURISTIC_INLINING }}
         run: |
           cat <<EOF > run-tests.lisp
           (ql:quickload :coalton/tests)
@@ -74,6 +77,7 @@ jobs:
         env:
           COALTON_ENV: ${{ matrix.COALTON_ENV }}
           COALTON_DISABLE_SPECIALIZATION: ${{ matrix.DISABLE_SPECIALIZATION }}
+          COALTON_HEURISTIC_INLINING: ${{ matrix.HEURISTIC_INLINING }}
         run: |
           cat <<EOF > run-tests.lisp
           (ql:quickload :coalton/tests)
--- a/src/codegen/ast.lisp
+++ b/src/codegen/ast.lisp
@@ -281,20 +281,6 @@ call to (break)."
       (loop :for (name . node) :in bindings
             :collect (cons name (intersection binding-names (node-variables node))))))))
 
-(defun node-application-symbol-rator (node)
-  "Returns the name of the function being called if it is known"
-  (declare (type (or node-application node-direct-application) node)
-           (values (or null parser:identifier)))
-  (etypecase node
-    (node-direct-application
-     (node-direct-application-rator node))
-
-    (node-application
-     (unless (node-variable-p (node-application-rator node))
-       (return-from node-application-symbol-rator))
-
-     (node-variable-value (node-application-rator node)))))
-
 (defun node-rands (node)
   (declare (type (or node-application node-direct-application))
            (values node-list))
@@ -306,6 +292,7 @@ call to (break)."
      (node-application-rands node))))
 
 (defun node-rator-name (node)
+  "Returns the name of the function being called if it is known"
   (declare (type (or node-application node-direct-application))
            (values (or null parser:identifier)))
   (etypecase node
--- a/src/codegen/optimizer.lisp
+++ b/src/codegen/optimizer.lisp
@@ -321,12 +321,19 @@
      (list
       (action (:after node-application) #'rewrite-application)))))
 
+(defun aggressive-heuristic (node)
+  "This will probably inline more than is optimal."
+  (and (<= (count-applications node)
+           8)
+       (<= (count-nodes node)
+           32)))
+
 (defun heuristic-inline-applications (node
                                       env
                                       &key
-                                        (max-depth  0)
-                                        (max-unroll 0)
-                                        (heuristic  (constantly nil)))
+                                        (max-depth  16)
+                                        (max-unroll 4)
+                                        (heuristic  #'aggressive-heuristic))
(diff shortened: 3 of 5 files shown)


src/codegen/ast.lisp
```diff
@@ -281,20 +281,6 @@ call to (break)."
       (loop :for (name . node) :in bindings
             :collect (cons name (intersection binding-names (node-variables node))))))))
 
-(defun node-application-symbol-rator (node)
```
amorphedstar, line 284: This was apparently an unused duplicate of `node-rator-name` below.

stylewarning:
We should add @digikar99 as git commit co-author.

src/codegen/optimizer.lisp
```diff
@@ -316,6 +320,106 @@
...
+           (declare (type node-let               node)
+                    (type parser:identifier-list call-stack)
+                    (values node-let &optional))
+           (make-node-let
+            :type     (node-type node)
+            :bindings (loop :for (name . subnode) :in (node-let-bindings node)
+                            :collect (cons name
+                                           (if (and (typep subnode '(or node-application
+                                                                        node-direct-application))
+                                                    (eq 'coalton:Cons (node-rator-name subnode))
```
amorphedstar, line 398: @eliaslfox was right - preventing inlining `Cons` was the only thing necessary to get the recursive data working, or at least to get the tests to all pass.

src/codegen/traverse.lisp
```diff
@@ -242,18 +232,19 @@ other nodes, then it would be inappropriate to also define an action
...
+                      (values node &optional))
+             (alexandria-2:line-up-last
+              current-node
+              (fire-action ':before   'node                                actions args)
+              (fire-action ':before   (class-name (class-of current-node)) actions args)
+              (fire-action ':traverse (class-name (class-of current-node)) actions args)
+              (fire-action ':after    (class-name (class-of current-node)) actions args)
+              (fire-action ':after    'node                                actions args))))
+      (let ((*traverse* #'current-traverse))
+        (apply *traverse* initial-node initial-args)))))
```
amorphedstar, line 247: Is this the right way to bind it?

src/codegen/traverse.lisp
```diff
@@ -355,6 +346,25 @@ gets called."
        )))
     counter))
 
+(defun count-applications (node)
+  "Count the number of function application nodes in the AST
+corresponding to `node`."
```
amorphedstar, line 351: For use in a heuristic to (indirectly) control how much branching is possible

src/codegen/traverse.lisp
```diff
@@ -7,11 +7,19 @@
...
+   #:*traverse*
    #:traverse
    #:traverse-with-binding-list))
 
 (in-package #:coalton-impl/codegen/traverse)
 
+(defvar *traverse* nil
+  "The recursive function reference for the current traversal. This will
+be bound for the dynamic extent of any traversal. Inside of an
+`action`, `(funcall *traverse* node)` may be used to invoke recursion.")
```
amorphedstar, line 21: Most changes in this file are renaming and documentation updates to reflect this

src/codegen/optimizer.lisp
```diff
@@ -316,6 +320,106 @@
...
+                          :into subs
+                        :finally (return (values bindings subs)))
+                (make-node-let
+                 :type     (node-type node)
+                 :bindings bindings
+                 :subexpr  (funcall *traverse*
+                                    (rename-type-variables (apply-ast-substitution
+                                                            subs
+                                                            (node-abstraction-subexpr code)))
+                                    (cons name call-stack)))))
```
amorphedstar, line 386: This main inlining substitution is copied almost verbatim from https://github.com/coalton-lang/coalton/pull/1029 , just with type renaming and the stack

src/codegen/optimizer.lisp
```diff
@@ -316,6 +320,106 @@
...
+           (type (integer 0)    max-depth)
+           (type (integer 0)    max-unroll)
+           (type function       heuristic)
+           (values node &optional))
+  (flet ((try-inline (node call-stack)
+           (declare (type (or node-application node-direct-application) node)
+                    (type parser:identifier-list                        call-stack)
+                    (values (or node-let node-application node-direct-application) &optional))
+           (or
+            (alexandria:when-let*
```
amorphedstar, line 356: It seems `alexandria:if-let` is not an option because only `alexandria:when-let*` has early exiting

src/codegen/optimizer.lisp
```diff
@@ -316,6 +320,106 @@
...
+           8)
+       (<= (count-nodes node)
+           32)))
+
+(defun heuristic-inline-applications (node
+                                      env
+                                      &key
+                                        (max-depth  16)
+                                        (max-unroll 4)
+                                        (heuristic  #'aggressive-heuristic))
```
amorphedstar, line 334: Now that the inliner inlines direct applications just the same as applications, it takes forever with the heuristic `(constantly t)`, even as low as `:max-depth 2 :max-unroll 1`. It seems that some of the tests must have significantly branching recursive functions with large bodies.

.github/workflows/main.yml
```diff
@@ -18,7 +18,8 @@ jobs:
         LISP_IMPL: ["sbcl", "allegro", "ccl"]
         COALTON_ENV: ["release", "development"]
         SAFETY: ["0", "3"]
-        DISABLE_SPECIALIZATION: ["1", "0"] 
+        DISABLE_SPECIALIZATION: ["1", "0"]
+        HEURISTIC_INLINING: ["1", "0"]
```
amorphedstar, line 22: Is it fine with double the tests, with heuristic inlining enabled or disabled like this?

src/codegen/traverse.lisp
```diff
@@ -7,11 +7,19 @@
...
+   #:*traverse*
    #:traverse
    #:traverse-with-binding-list))
 
 (in-package #:coalton-impl/codegen/traverse)
 
+(defvar *traverse* nil
+  "The recursive function reference for the current traversal. This will
+be bound for the extent of any traversal. Inside of an
+`action`, `(funcall *traverse* node)` may be used to invoke recursion.")
```
amorphedstar, line 21: Most changes in this file are renaming and documentation updates to reflect this

.github/workflows/main.yml
stylewarning: Yes.

src/codegen/optimizer.lisp
stylewarning: @eliaslfox 🎉

stylewarning: @amorphedstar Can we please abstract this test out and write a nice comment?

It's such unusual "business logic" that it's worth not muddying the code too much *and* properly documenting it.

src/codegen/traverse.lisp
stylewarning: If this has no meaning outside of the traversal, I recommend making it unbound.

```
(defvar *traverse*)
(setf (documentation '*traverse 'variable)
      "doc string")
```

It's a little obtuse but it ensures `*traverse*` will error when evaluated outside of its context.

amorphedstar: Done: https://github.com/coalton-lang/coalton/pull/1193/files?diff=unified&w=0#diff-6613c23f2ebfe4977201bab1561ab3e597cdc43410389416ddfaae8a03194f98R19-R24

src/codegen/traverse.lisp
```diff
@@ -372,3 +384,36 @@ without any slot information."
...
+                                          :type (node-type subnode)
+                                          :rator-type (node-rator-type subnode)
+                                          :rator 'coalton:Cons
+                                          :rands (mapcar
+                                                  (lambda (rand)
+                                                    (apply *traverse* rand args))
+                                                  (node-rands subnode)))
+                                         (apply *traverse* subnode args))))
+      :subexpr  (apply *traverse* (node-let-subexpr node) args)))
+   t))
```
amorphedstar, line 419: The `Cons`-skipping is now abstracted here, and could conceivably be used to override the default `:traverse node-let` action on any traversals which have a risk of rewriting a `node-application` or `node-direct-application` to another type

src/codegen/optimizer.lisp
amorphedstar: here: https://github.com/coalton-lang/coalton/pull/1193/files?diff=unified&w=0#r1707561858

PR #1194, opened 2024-08-06
Propagate file:span information through toplevel definitions

jbouwman:
Most AST types have a `source` field of type `(cons integer integer)` that indicates the character span of the source form associated with the parsed object.

In order to support runtime lookup of source definitions, change the type of `source` to `source-location`, which contains both a span and a filename.

This is stacked on #1209 and left in draft state while that change is considered.


jbouwman:
Superseded by #1216

PR #1196, opened 2024-08-07
inline applications of abstractions

stylewarning:


src/codegen/optimizer.lisp
```diff
@@ -351,43 +351,95 @@ requires direct constructor calls."
...
-           (declare (type (or node-application node-direct-application) node)
-                    (type parser:identifier-list                        call-stack)
-                    (values (or node-let node-application node-direct-application) &optional))
-           (or
-            (alexandria:when-let*
-                ((name (node-rator-name node))
-                 (code (tc:lookup-code env name :no-error t))
-                 (_    (and (node-abstraction-p code)
+  (labels ((inline-abstraction-from-application (node abstraction process-body)
+             "If NODE is an abstraction (f e1 e2 ... en) and ABSTRACTION is a
```
amorphedstar, line 355: Do we prefer "`` NODE ``" instead of "`` `node` ``"?

src/codegen/optimizer.lisp
```diff
@@ -351,43 +351,95 @@ requires direct constructor calls."
...
-                (make-node-let
-                 :type     (node-type node)
-                 :bindings bindings
-                 :subexpr  (funcall *traverse*
-                                    (rename-type-variables (apply-ast-substitution
-                                                            subs
-                                                            (node-abstraction-subexpr code)))
-                                    (cons name call-stack)))))
-            node)))
+                               (length (node-rands node))))
```
amorphedstar, line 435: this could be lined up with when-let like the other case if desired

src/codegen/optimizer.lisp
stylewarning: In Lisp code I've preferred capitalized names. But I actually forgot what our convention was.

Lisp code docstrings aren't emitted to markdown anywhere.

Issue #1197, opened 2024-08-08
TRANSLATE-EXPRESSION on NODE-FOR is buggy


stylewarning:
**EDIT**:

I thought this was a bug in the inliner, but it appears that the `translate-expression` method of `node-for` is doing manual type constructions, and those type constructions are incorrect.

Original post:

repro:

Enable inlining, and

```
(coalton-codegen
  (define (f)
    (for x in (iter:up-to 5)
      Unit)))
```

results in

```
Failed to unify types ITERATOR and OPTIONAL
   [Condition of type COALTON-IMPL/TYPECHECKER/TYPE-ERRORS:UNIFICATION-ERROR]

Restarts:
 0: [RETRY] Retry SLIME REPL evaluation request.
 1: [*ABORT] Return to SLIME's top level.
 2: [ABORT] abort thread (#<THREAD "repl-thread" RUNNING {1001378003}>)

Backtrace:
  0: ((:METHOD COALTON-IMPL/TYPECHECKER/UNIFY::MGU (COALTON-IMPL/TYPECHECKER/TYPES:TYCON COALTON-IMPL/TYPECHECKER/TYPES:TYCON)) ITERATOR OPTIONAL) [fast-method]
      Locals:
        COALTON-IMPL/TYPECHECKER/UNIFY::TYPE1 = ITERATOR
        COALTON-IMPL/TYPECHECKER/UNIFY::TYPE2 = OPTIONAL
  1: ((:METHOD COALTON-IMPL/TYPECHECKER/UNIFY::MGU (COALTON-IMPL/TYPECHECKER/TYPES:TAPP COALTON-IMPL/TYPECHECKER/TYPES:TAPP)) (ITERATOR INTEGER) (OPTIONAL INTEGER)) [fast-method]
      Locals:
        COALTON-IMPL/TYPECHECKER/UNIFY::TYPE1 = (ITERATOR INTEGER)
        COALTON-IMPL/TYPECHECKER/UNIFY::TYPE2 = (OPTIONAL INTEGER)
  2: (COALTON-IMPL/TYPECHECKER/UNIFY:UNIFY COMMON-LISP:NIL (ITERATOR INTEGER) (OPTIONAL INTEGER))
[...]

PR #1199, opened 2024-08-08
fix translate-expression for node-for

stylewarning:
This PR fixes two bugs, #1197 and #1200 

#1197 was caused by incorrect types being assigned to the resulting translated `node-for` in `translate-expression`. I think this was just human mistake in constructing a somewhat complicated AST node with lots of sub-nodes. I believe I have it correct, but I'm not 100% confident. (It does pass a new test I added.) Is there a better way to do it?

#1200 was problematic because all looping constructs implicitly returned `cl:nil`, but the type checker unified loops to `Unit`. I've changed all of the loops to just explicitly codegen a `Unit` return.

### DIFF
--- a/src/codegen/codegen-expression.lisp
+++ b/src/codegen/codegen-expression.lisp
@@ -129,8 +129,8 @@
       `(loop
          :named ,(break-label label)
          :while ,pred-expr
-         :do
-            (block ,(continue-label label) ,body-expr))))
+         :do (block ,(continue-label label) ,body-expr)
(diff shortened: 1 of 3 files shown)


amorphedstar (approved):
Looks to me like this fixed the `(Iterator (Iterator #T...))` and other weird types coming out of the translation.

eliaslfox (approved):

Izaakwltn (approved):

PR #1203, opened 2024-08-09
Add node-block and node-return-from for inlining functions with node-return

amorphedstar:
Inlining bodies of functions with return nodes would make it so they returned to the most recent non-inlined function, which is wrong. This PR creates `node-block` and `node-return-from` and implements a traversal to fix this.

### DIFF
--- a/library/system.lisp
+++ b/library/system.lisp
@@ -88,7 +88,7 @@ While the result will always contain microseconds, some implementations may retu
       (cl:let ((env (uiop:getenvp var)))
         (cl:if env
                (Some env)
-               (None)))))
+               None))))
 
   
   (declare setenv! (String -> String -> Unit))
@@ -164,7 +164,7 @@ While the result will always contain microseconds, some implementations may retu
       (cl:let ((arg (uiop:argv0)))
         (cl:if arg
                (Some (uiop:argv0))
-               (None))))))
+               None)))))
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/SYSTEM")
--- a/src/codegen/ast.lisp
+++ b/src/codegen/ast.lisp
@@ -15,88 +15,93 @@
    #:binding-list                       ; TYPE
    #:node-literal                       ; STRUCT
    #:make-node-literal                  ; CONSTRUCTOR
-   #:node-literal-value                 ; ACCESSOR
+   #:node-literal-value                 ; READER
    #:node-variable                      ; STRUCT
    #:make-node-variable                 ; CONSTRUCTOR
    #:node-variable-p                    ; FUNCTION
-   #:node-variable-value                ; ACCESSOR
+   #:node-variable-value                ; READER
    #:node-application                   ; STRUCT
    #:make-node-application              ; CONSTRUCTOR
    #:node-application-p                 ; FUNCTION
-   #:node-application-rator             ; ACCESSOR
-   #:node-application-rands             ; ACCESSOR
+   #:node-application-rator             ; READER
+   #:node-application-rands             ; READER
    #:node-direct-application            ; STRUCT
    #:make-node-direct-application       ; CONSTRUCTOR
-   #:node-direct-application-rator-type ; ACCESSOR
-   #:node-direct-application-rator      ; ACCESSOR
-   #:node-direct-application-rands      ; ACCESSOR
+   #:node-direct-application-rator-type ; READER
+   #:node-direct-application-rator      ; READER
+   #:node-direct-application-rands      ; READER
    #:node-direct-application-p          ; FUNCTION
    #:node-abstraction                   ; STRUCT
    #:make-node-abstraction              ; CONSTRUCTOR
-   #:node-abstraction-vars              ; ACCESSOR
-   #:node-abstraction-subexpr           ; ACCESSOR
+   #:node-abstraction-vars              ; READER
+   #:node-abstraction-subexpr           ; READER
    #:node-abstraction-p                 ; FUNCTION
    #:node-let                           ; STRUCT
    #:make-node-let                      ; CONSTRUCTOR
    #:node-let-p                         ; FUNCTION
-   #:node-let-bindings                  ; ACCESSOR
-   #:node-let-subexpr                   ; ACCESSOR
+   #:node-let-bindings                  ; READER
+   #:node-let-subexpr                   ; READER
    #:node-lisp                          ; STRUCT
    #:make-node-lisp                     ; CONSTRUCTOR
-   #:node-lisp-vars                     ; ACCESSOR
-   #:node-lisp-form                     ; ACCESSOR
+   #:node-lisp-vars                     ; READER
+   #:node-lisp-form                     ; READER
    #:match-branch                       ; STRUCT
    #:make-match-branch                  ; CONSTRUCTOR
-   #:match-branch-pattern               ; ACCESSOR
-   #:match-branch-bindings              ; ACCESSOR
-   #:match-branch-body                  ; ACCESSOR
+   #:match-branch-pattern               ; READER
+   #:match-branch-bindings              ; READER
+   #:match-branch-body                  ; READER
    #:branch-list                        ; TYPE
    #:node-match                         ; STRUCT
    #:make-node-match                    ; CONSTRUCTOR
-   #:node-match-expr                    ; ACCESSOR
-   #:node-match-branches                ; ACCESSOR
+   #:node-match-expr                    ; READER
+   #:node-match-branches                ; READER
    #:node-while                         ; STRUCT
    #:make-node-while                    ; CONSTRUCTOR
-   #:node-while-label                   ; ACCESSOR
-   #:node-while-expr                    ; ACCESSOR
+   #:node-while-label                   ; READER
+   #:node-while-expr                    ; READER
    #:node-while-body                    ; ACESSOR
    #:node-while-let                     ; STRUCT
    #:make-node-while-let                ; CONSTRUCTOR
    #:node-while-let-label               ; ACESSOR
-   #:node-while-let-pattern             ; ACCESSOR
-   #:node-while-let-expr                ; ACCESSOR
+   #:node-while-let-pattern             ; READER
+   #:node-while-let-expr                ; READER
    #:node-while-let-body                ; ACESSOR
    #:node-loop                          ; STRUCT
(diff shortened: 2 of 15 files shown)


src/codegen/transformations.lisp
```diff
@@ -123,3 +124,37 @@ both CL namespaces appearing in `node`"
...
+   (list
+    (action (:before node-return _node)
+      (declare (ignore _node))
+      (return-from node-returns-p t))))
+  nil)
+
+(defun localize-returns (node)
+  "If `node` contains any `node-return` subnodes, wrap `node` in a new
+`node-block` and rewrite all `node-return` subnodes to return from
+this block."
```
amorphedstar, line 143: Is there an established name for this?

src/codegen/transformations.lisp
```diff
@@ -123,3 +124,37 @@ both CL namespaces appearing in `node`"
...
+  (declare (type node node)
+           (values node &optional))
+  (unless (node-returns-p node)
+    (return-from localize-returns node))
+
+  (let ((block-name (gentemp)))
```
amorphedstar, line 149: Should I make the inliner pass a function name so the block name can be something more informative like `INLINED-FUNCTIONNAME` or `INLINED-LAMBDA`?

src/codegen/ast.lisp
```diff
@@ -251,6 +259,16 @@ call to (break)."
...
 
+(defstruct (node-return-from (:include node))
+  "A return statement, used for early returns in inlined functions"
+  (name (util:required 'name) :type symbol :read-only t)
+  (expr (util:required 'expr) :type node   :read-only t))
+
+(defstruct (node-block (:include node))
+  "A return target, used for early returns from inlined functions"
+  (name (util:required 'node) :type symbol :read-only t)
+  (expr (util:required 'expr) :type node   :read-only t))
```
amorphedstar, line 270: Should I rename this slot to `body` ?

stylewarning: body makes sense

src/codegen/ast.lisp
```diff
@@ -82,6 +82,14 @@
    #:node-return                        ; STRUCT
    #:make-node-return                   ; CONSTRUCTOR
    #:node-return-expr                   ; ACCESSOR
+   #:node-return-from                   ; STRUCT
+   #:make-node-return-from              ; CONSTRUCTOR
+   #:node-return-from-name              ; ACCESSOR
```
stylewarning, line 87: teChnically these read-only things are readers

src/codegen/codegen-expression.lisp
```diff
@@ -241,6 +245,12 @@
     (assert (not (null current-function)))
     `(return-from ,current-function ,(codegen-expression (node-return-expr expr) current-function env)))
 
+  (:method ((expr node-return-from) current-function env)
+    `(return-from ,(block-label (node-return-from-name expr)) ,(codegen-expression (node-return-from-expr expr) current-function env)))
```
stylewarning, line 249: nit: newline after the label expression

src/codegen/codegen-expression.lisp
```diff
@@ -241,6 +245,12 @@
...
+
+  (:method ((expr node-block) current-function env)
+    `(block ,(block-label (node-block-name expr)) ,(codegen-expression (node-block-expr expr) current-function env)))
```
stylewarning, line 252: nit: newline after the label

src/codegen/optimizer.lisp
```diff
@@ -399,8 +400,9 @@ requires direct constructor calls."
                                       ;; so they aren't
                                       ;; inadvertently unified
                                       ;; across substitutions.
-                                      (rename-type-variables
-                                       (apply-ast-substitution subs body)))))))
+                                      (localize-returns
```
stylewarning, line 403: add to the comment above, about the returns

src/codegen/transformations.lisp
```diff
@@ -123,3 +124,37 @@ both CL namespaces appearing in `node`"
...
   (null (intersection (node-variables node) bound-variables)))
+
+(defun node-returns-p (node)
+  "Does `node` contain a `node-return` subnode?"
+  (declare (type node node)
+           (values boolean &optional))
+  (traverse
+   node
+   (list
+    (action (:before node-return _node)
```
stylewarning, line 135: I forgot the semantics of return. Does

    (fn () (return Unit))

work by returning from the nearest enclosed lambda?

src/codegen/transformations.lisp
stylewarning: i would say this is about making returns explicit rather than localizing them.

src/codegen/transformations.lisp
stylewarning: we might do an earlier pass altogether that transforms ALL functions to always have a block around the contents. thoughts?

stylewarning (changes requested):
generally good. two Q's:

- should we do a single pass that transforms all functions to have a named block with explicated returns?

- does return go back to the nearest enclosing lambda? if so, the logic will need some adjustment.

src/codegen/transformations.lisp
stylewarning: Answer: yes, `return` returns from the lambda, so this PR will need adjustments to account for that

src/codegen/transformations.lisp
amorphedstar: I think this is an interesting possibility, and I tried to implement this in the tc-node -> ast-node translation. The sad thing is that the output of the codegen will then look messier with these extra blocks instead, of return-from function-name, unless there is extra logic added to codegen-expression

library/system.lisp
```diff
@@ -88,7 +88,7 @@ While the result will always contain microseconds, some implementations may retu
       (cl:let ((env (uiop:getenvp var)))
         (cl:if env
                (Some env)
-               (None)))))
+               None))))
```
amorphedstar, line 91: is this right

src/typechecker/types.lisp
```diff
@@ -305,7 +305,7 @@
   (declare (type ty ty)
            (type fixnum num))
 
-  (assert (>= num (length (function-type-arguments ty))))
+  (assert (<= num (length (function-type-arguments ty))))
```
amorphedstar, line 308: This would throw type errors like
```lisp
The assertion
(COMMON-LISP:>= COALTON-IMPL/TYPECHECKER/TYPES::NUM
                #1=(COMMON-LISP:LENGTH
                    (COALTON-IMPL/TYPECHECKER/TYPES:FUNCTION-TYPE-ARGUMENTS
                     COALTON-IMPL/TYPECHECKER/TYPES:TY)))
failed with COALTON-IMPL/TYPECHECKER/TYPES::NUM = 1, #1# = 2.
   [Condition of type COMMON-LISP:SIMPLE-ERROR]
```
on code like
```clojure
COALTON-USER> (coalton (let ((f (fn (x) (fn (y) (* x y))))) ((f 1) 2)))
```
or
```clojure
COALTON-USER> (coalton-toplevel (define (f x) (fn (x) (fn (y) (* x y)))))
```

stylewarning: @amorphedstar can we add a test for this

library/system.lisp
stylewarning: yes

stylewarning (commented):
This is looking slick. Is it about ready for merging?

Would be nice to add a simple test for some things, including a test that would trip up the inliner with the returns.

amorphedstar:
> This is looking slick. Is it about ready for merging?
> 
> Would be nice to add a simple test for some things, including a test that would trip up the inliner with the returns.

I added two simple tests, but wasn't quite sure the best place to put them. And I think it should be about ready to merge if you think it's fine having all these extra `block`s now in the codegen output, and potentially unclearer return names in a few places that had special codegen behavior, like `node-bind` in the case when `node-bind-expr` is a `node-abstraction`.

stylewarning:
> > This is looking slick. Is it about ready for merging?
> 
> > 
> 
> > Would be nice to add a simple test for some things, including a test that would trip up the inliner with the returns.
> 
> 
> 
> I added two simple tests, but wasn't quite sure the best place to put them. And I think it should be about ready to merge if you think it's fine having all these extra `block`s now in the codegen output, and potentially unclearer return names in a few places that had special codegen behavior, like `node-bind` in the case when `node-bind-expr` is a `node-abstraction`.

Since blocks are only useful if there's a lexically apparent return, maybe we can (either in this PR or future) only emit a block when needed.

src/codegen/codegen-expression.lisp
```diff
@@ -403,16 +374,19 @@
...
+         `(let ((,name ,(codegen-expression node_ env)))
             (declare (ignorable ,name))
-            ,(codegen-let node (cdr sccs) current-function local-vars env))))
+            ,(codegen-let node (cdr sccs) local-vars env))))
 
       (t (error "Invalid scc binding group. This should have been detected during typechecking.")))))
 
 
-(defun argument-types (node env)
+(defun function-declare (node env)
```
stylewarning, line 384: nit: rename `function-declarations`

src/codegen/program.lisp
```diff
@@ -125,20 +126,9 @@ A function bound here will be called with a keyword category, and one or more ad
...
-                    :collect `(type ,(tc:lisp-type arg-ty env) ,name))
-              (list `(values ,(tc:lisp-type (node-type (node-abstraction-subexpr node)) env)
-                             &optional))))))
-
-    `(defun ,name ,(node-abstraction-vars node)
-       (declare (ignorable ,@(node-abstraction-vars node))
-                ,@type-decs)
-       ,(codegen-expression (node-abstraction-subexpr node) name env))))
+  `(defun ,name ,(node-abstraction-vars node)
+       ,(function-declare node env)
```
stylewarning, line 130: nit: indenting looks off

stylewarning (approved):
I have a few nits that I don't think should hold up the PR. You can address them if desired, but if not within 24h, I'll merge.

We can reduce the number of emitted blocks, but it's fine if you want to
make that a GH issue that can be addressed later. In fact I think
I'd prefer that so as to not let this PR drag.

src/codegen/codegen-expression.lisp
jbouwman: is done

src/codegen/program.lisp
jbouwman: is done

PR #1205, opened 2024-08-09
Allow Coalton config before Coalton is loaded

stylewarning:
This PR allows configuration to reside in one's lisp config directory without relying on environment variables

my sbclrc has:

```
;;; Coalton configuration

(let ((config '((:compiler-mode          "development")
                (:print-unicode          t)
                (:perform-specialization t)
                (:perform-inlining       nil)
                (:emit-type-annotations  t)
                (:print-types            t))))
  (setf (symbol-plist ':coalton-config) nil)
  (loop :for (key value) :in config
        :do (setf (get ':coalton-config key) value)))
```

as my preferred default

### DIFF
--- a/src/settings.lisp
+++ b/src/settings.lisp
@@ -18,8 +18,54 @@
 
 (in-package #:coalton-impl/settings)
 
+;;; Configuration reaches the Coalton compiler in 3 ways.
+;;;
+;;;    1. When Coalton is compiled, several environment variables are
+;;;    inquired. These are
+;;;
+;;;        - COALTON_ENV: Whether to be in development or release mode.
+;;;
+;;;        - COALTON_DISABLE_SPECIALIZATION: Whether to disable
+;;;          function specialization.
+;;;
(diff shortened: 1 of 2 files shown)


src/settings.lisp
```diff
@@ -18,8 +18,54 @@
...
+
+
+;;; Compiler and runtime configuration
+
 (declaim (type boolean *coalton-print-unicode*))
-(defvar *coalton-print-unicode* t
+(defvar *coalton-print-unicode* (config ':print-unicode :default t)
   "Whether to print coalton info using unicode symbols")
 
 (defun coalton-release-p ()
```
Izaakwltn, line 71: This should maybe mention https://github.com/coalton-lang/coalton/blob/main/docs/coalton-lisp-interop.md
which I believe is the only thorough documentation on the difference between the two modes.

Izaakwltn (approved):
lgtm

jbouwman (approved):

PR #1209, opened 2024-08-13
Supply source-error:file to maybe-read-form

### LINKED ISSUE #1206: Improve presentation of low-level reader errors
Hard syntax errors like unterminated strings, unknown character literals, and so on will signal eclector conditions that are unhandled. For instance, an unterminated form results in:

```
Unhandled ECLECTOR.READER:UNTERMINATED-LIST in thread #<SB-THREAD:THREAD tid=259 "main thread" RUNNING
                                                         {7004FD0733}>:
  While reading list, expected the character ) when input ended.

Backtrace for: #<SB-THREAD:THREAD tid=259 "main thread" RUNNING {7004FD0733}>
```

Improve reader conditional handling so that a reasonable coalton source-error is signalled and printed instead, indicating file and character range for offending input.

jbouwman:
This change builds a source-error:file object immediately when Coalton is passed control of an input stream, so that any activity against that stream can be wrapped in a handler that rethrows lower-level reader errors as parser-errors.

source-error:file is made abstract, and the error printer reads source text from a newly opened stream when printing. This allows `file` objects to be attached directly to and dumped along with coalton objects to support runtime navigation to symbols, as #1194

Closes #1206

### DIFF
--- a/coalton-asdf.lisp
+++ b/coalton-asdf.lisp
@@ -15,11 +15,6 @@
     (call-with-around-compile-hook
      c (lambda (&rest flags)
          (declare (ignore flags))
-         (with-open-file (stream coal-file
-                                 :direction ':INPUT
-                                 :element-type 'character
-                                 :external-format ':UTF8)
-           (let ((char-stream (coalton-impl/stream:make-char-position-stream stream)))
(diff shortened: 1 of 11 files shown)


jbouwman:
Superseded by #1216

PR #1211, opened 2024-08-14
do not create unnecessary blocks

amorphedstar:
Cleanup to #1203 with a traversal to check for if there actually are any returns

### DIFF
--- a/src/codegen/translate-expression.lisp
+++ b/src/codegen/translate-expression.lisp
@@ -12,6 +12,11 @@
   (:import-from
    #:coalton-impl/codegen/typecheck-node
    #:typecheck-node)
+  (:import-from
+   #:coalton-impl/codegen/traverse
+   #:action
+   #:*traverse*
+   #:traverse)
   (:local-nicknames
    (#:util #:coalton-impl/util)
    (#:tc #:coalton-impl/typechecker))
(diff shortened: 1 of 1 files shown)


src/codegen/translate-expression.lisp
```diff
@@ -34,9 +39,7 @@ TRANSLATE-EXPRESSION when an abstraction is being translated.")
            (type symbol name)
            (type pred-context extra-context))
 
-  (let* ((*current-function-name* name)
```
amorphedstar, line 37: this should only be bound when the `name` represents a `node-abstraction` (first two cases of the `cond` here).

src/codegen/translate-expression.lisp
```diff
@@ -984,10 +989,28 @@ dictionaries applied."
...
+  (traverse
+   inner
+   (list
+    (action (:traverse node-block node)
+      (unless (eq *current-function-name* (node-block-name node))
+        (funcall *traverse* (node-block-body node)))
+      (values))
+    (action (:before node-return-from node)
+      (when (eq *current-function-name* (node-return-from-name node))
+        (return-from wrap-in-block
```
stylewarning, line 1011: it's funny that we use early exit in a language that doesn't allow early exit via return in a lambda

stylewarning (approved):

PR #1212, opened 2024-08-14
Add a constant-propagation pass

digikar99:
EDIT: This PR aims to propagate constant literals or other constant objects such as obtained from node-lisp or dict-variables/codegen-syms required to resolve method instances. 

Two issues remain for future work:

1. Determining the set of constant objects might require a list of predicates which may be domain-specific. Let's leave that for future work. Implementing it is trivial, but brainstorming what the right way to do it might take some time. (See [this thread](https://github.com/coalton-lang/coalton/pull/1212#discussion_r1717913405).)
2. It turns out that while propagating constants, it also becomes possible to update the types of certain nodes. (See [this thread](https://github.com/coalton-lang/coalton/pull/1212#discussion_r1720887020).) It's also possible that there are other passes which may make it advantageous to update the node types. In that case, it seems more appropriate to implement node type updation in the traverse pass itself, rather than specifically as part of constant propagation. Thus, updating the traverse functions to update the node types is another branch of future work.

---

Consider the following example:

```lisp
(in-package :coalton-user)

(coalton-toplevel
  (declare add-df (double-float -> double-float -> double-float))
  (define (add-df x y)
    (lisp double-float (x y) (cl:+ x y)))

  (define (add-df-caller)
    (let ((x 5.0d0))
      (add-df x x))))
```

Without the constant-propagation pass, `(lookup-code 'add-df-caller)` returns a node containing a let with a variable binding followed by direct-application:

```lisp
#s(coalton-impl/codegen/ast:node-let
   :type double-float
   :bindings ((x-4321
               . #s(coalton-impl/codegen/ast:node-literal
                    :type double-float
                    :value 5.0d0)))
   :subexpr #s(coalton-impl/codegen/ast:node-direct-application
               :type double-float
               :rator-type (double-float → double-float → double-float)
               :rator add-df
               :rands (#s(coalton-impl/codegen/ast:node-variable
                          :type double-float
                          :value x-4321)
                         #s(coalton-impl/codegen/ast:node-variable
                            :type double-float
                            :value x-4321))))
```

With constant-propagation in place, the constant let bindings are substituted with the literals instead:

```lisp
#s(coalton-impl/codegen/ast:node-direct-application
   :type double-float
   :rator-type (double-float → double-float → double-float)
   :rator add-df
   :rands (#s(coalton-impl/codegen/ast:node-literal
              :type double-float
              :value 5.0d0)
             #s(coalton-impl/codegen/ast:node-literal
                :type double-float
                :value 5.0d0)))
```

### DIFF
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -24,7 +24,7 @@
           (no-error
            nil)
           (t
-           (error "Expected VAR ~S to be a constant but is not" var)))))
+           (error "Expected VAR ~S to be a constant but it is not" var)))))
 
 (defun constant-node-p (env node constant-bindings)
   "If NODE is a constant, returns the constant value corresponding to it.
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -26,21 +26,34 @@
           (t
            (error "Expected VAR ~S to be a constant but is not" var)))))
 
+(defun constant-node-p (env node constant-bindings)
+  "If NODE is a constant, returns the constant value corresponding to it.
+If not, returns NIL"
+  (declare (type tc:environment env)
+           (type node node)
+           (type list constant-bindings))
+  (cond ((node-literal-p node)
+         (if (numberp (node-literal-value node))
+             node
+             nil))
+        ((node-variable-p node)
+         (cond ((and (eq 'coalton:boolean (node-type node))
+                     (member (node-variable-value node) '(coalton:true coalton:false)))
+                node)
+               ((tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t)
+                node)
+               (t
+                (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
+        ((node-lisp-p node)
+         (if (cl:constantp (node-lisp-form node))
+             node
+             nil))))
+
 (defun propagate-constants (node env)
   (declare (optimize debug))
-  (labels ((constant-node-p (node constant-bindings)
-             ;; FIXME: Do we need more nodes classified as constants?
-             (or (node-literal-p node)
-                 (and (node-variable-p node)
-                      (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
-                 (and (node-variable-p node)
-                      (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
-           (propagate-constants-node-variable (node constant-bindings)
-             ;; FIXME: Do we need more nodes classified as constants?
+  (labels ((propagate-constants-node-variable (node constant-bindings)
              (declare (type node-variable node))
-             (if (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t)
-                 node
-                 (constant-var-value (node-variable-value node) constant-bindings :no-error t)))
+             (constant-node-p env node constant-bindings))
 
            (propagate-constants-node-let (node constant-bindings)
              (declare (type node-let node))
@@ -49,7 +62,7 @@
                    (nonconstant-bindings nil))
                (loop :for (var . value) :in node-bindings
                      :for propagated-value-node := (funcall *traverse* value constant-bindings)
-                     :do (if (constant-node-p propagated-value-node constant-bindings)
+                     :do (if (constant-node-p env propagated-value-node constant-bindings)
                              (push (cons var propagated-value-node)
                                    new-constant-bindings)
                              (push (cons var propagated-value-node)
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -40,17 +40,11 @@
                  (and (node-variable-p node)
                       (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
            (propagate-constants-node-variable (node constant-bindings)
-             (declare (type node-variable node))
              ;; FIXME: Do we need more nodes classified as constants?
-             (cond ((node-literal-p node)
-                    node)
-                   ((and (node-variable-p node)
-                         (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
-                    node)
-                   ((node-variable-p node)
-                    (constant-var-value (node-variable-value node) constant-bindings :no-error t))
-                   (t
-                    nil)))
+             (declare (type node-variable node))
+             (if (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t)
+                 node
+                 (constant-var-value (node-variable-value node) constant-bindings :no-error t)))
 
            (propagate-constants-node-let (node constant-bindings)
              (declare (type node-let node))
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -7,12 +7,8 @@
   (:import-from
    #:coalton-impl/codegen/traverse
    #:action
-   #:count-applications
-   #:count-nodes
-   #:make-traverse-let-action-skipping-cons-bindings
    #:*traverse*
-   #:traverse
-   #:traverse-with-binding-list)
+   #:traverse)
   (:export
    #:propagate-constants))
 
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -23,10 +23,16 @@
       "Bound inside CONSTANT-PROPAGATION.")
 
 (defun constant-var-value (var &key no-error)
-  (the (or node null)
-       (or (cdr (assoc var *constant-bindings*))
-           (unless no-error
-             (error "Expected VAR ~S to be a constant but is not" var)))))
+  (declare (type symbol var)
+           (values (or node null) &optional))
+  (let ((may-be-value (cdr (assoc var *constant-bindings*))))
+    (cond (may-be-value
+           ;; Whenever it exists, it is always a NODE (thus, non-NIL)
+           may-be-value)
+          (no-error
+           nil)
+          (t
+           (error "Expected VAR ~S to be a constant but is not" var)))))
 
 (defun propagate-constants (node env)
   (declare (optimize debug))
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -39,7 +39,8 @@
                       (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
                  (and (node-variable-p node)
                       (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
-           (constant-node-value (node constant-bindings &key no-error)
+           (propagate-constants-node-variable (node constant-bindings)
+             (declare (type node-variable node))
              ;; FIXME: Do we need more nodes classified as constants?
              (cond ((node-literal-p node)
                     node)
@@ -47,16 +48,9 @@
                          (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
                     node)
                    ((node-variable-p node)
-                    (constant-var-value (node-variable-value node) constant-bindings :no-error no-error))
-                   (no-error
-                    nil)
+                    (constant-var-value (node-variable-value node) constant-bindings :no-error t))
                    (t
-                    (error "Expected the node to be a constant but is not ~%~S" node))))
-
-           (propagate-constants-node-variable (node constant-bindings)
-             (declare (type node-variable node))
-             (or (constant-node-value node constant-bindings :no-error t)
-                 node))
+                    nil)))
 
            (propagate-constants-node-let (node constant-bindings)
              (declare (type node-let node))
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -103,12 +103,11 @@
               :rator (node-direct-application-rator node)
               :rands (node-direct-application-rands node))))
 
-    (let ((*constant-bindings* nil))
-      (traverse
-       node
-       (list
-        (action (:after node-variable) #'propagate-constants-node-variable)
-        (action (:traverse node-let) #'propagate-constants-node-let)
-        (action (:after node-lisp) #'propagate-constants-node-lisp)
-        (action (:after node-direct-application) #'direct-application-better-infer-types))
-       nil))))
+    (traverse
+     node
+     (list
+      (action (:after node-variable) #'propagate-constants-node-variable)
+      (action (:traverse node-let) #'propagate-constants-node-let)
+      (action (:after node-lisp) #'propagate-constants-node-lisp)
+      (action (:after node-direct-application) #'direct-application-better-infer-types))
+     nil)))
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -33,7 +33,7 @@ If not, returns NIL"
            (type node node)
            (type list constant-bindings))
   (cond ((node-literal-p node)
-         (if (numberp (node-literal-value node))
+         (if (or (numberp (node-literal-value node)) (characterp (node-literal-value node)))
              node
              nil))
         ((node-variable-p node)
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -37,8 +37,8 @@ If not, returns NIL"
              node
              nil))
         ((node-variable-p node)
-         (cond ((and (eq 'coalton:boolean (node-type node))
-                     (member (node-variable-value node) '(coalton:true coalton:false)))
+         (cond ((and (eq 'coalton:Boolean (node-type node))
+                     (member (node-variable-value node) '(coalton:True coalton:False)))
                 node)
                ((tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t)
                 node)
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -18,14 +18,10 @@
 
 (in-package #:coalton-impl/codegen/constant-propagation)
 
-(defvar *constant-bindings*)
-(setf (documentation '*constant-bindings* 'variable)
-      "Bound inside CONSTANT-PROPAGATION.")
-
-(defun constant-var-value (var &key no-error)
+(defun constant-var-value (var constant-bindings &key no-error)
   (declare (type symbol var)
            (values (or node null) &optional))
-  (let ((may-be-value (cdr (assoc var *constant-bindings*))))
+  (let ((may-be-value (cdr (assoc var constant-bindings))))
     (cond (may-be-value
            ;; Whenever it exists, it is always a NODE (thus, non-NIL)
            may-be-value)
@@ -36,14 +32,14 @@
 
 (defun propagate-constants (node env)
   (declare (optimize debug))
-  (labels ((constant-node-p (node)
+  (labels ((constant-node-p (node constant-bindings)
              ;; FIXME: Do we need more nodes classified as constants?
              (or (node-literal-p node)
                  (and (node-variable-p node)
                       (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
                  (and (node-variable-p node)
-                      (constant-var-value (node-variable-value node) :no-error t))))
-           (constant-node-value (node &key no-error)
+                      (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
+           (constant-node-value (node constant-bindings &key no-error)
              ;; FIXME: Do we need more nodes classified as constants?
              (cond ((node-literal-p node)
                     node)
@@ -51,45 +47,43 @@
                          (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
                     node)
                    ((node-variable-p node)
-                    (constant-var-value (node-variable-value node) :no-error no-error))
+                    (constant-var-value (node-variable-value node) constant-bindings :no-error no-error))
                    (no-error
                     nil)
                    (t
                     (error "Expected the node to be a constant but is not ~%~S" node))))
 
-           (propagate-constants-node-variable (node call-stack)
-             (declare (ignore call-stack)
-                      (type node-variable node))
-             (or (constant-node-value node :no-error t)
+           (propagate-constants-node-variable (node constant-bindings)
+             (declare (type node-variable node))
+             (or (constant-node-value node constant-bindings :no-error t)
                  node))
 
-           (propagate-constants-node-let (node call-stack)
+           (propagate-constants-node-let (node constant-bindings)
              (declare (type node-let node))
              (let ((node-bindings (node-let-bindings node))
                    (new-constant-bindings nil)
                    (nonconstant-bindings nil))
                (loop :for (var . value) :in node-bindings
-                     :for propagated-value-node := (funcall *traverse* value call-stack)
-                     :do (if (constant-node-p propagated-value-node)
+                     :for propagated-value-node := (funcall *traverse* value constant-bindings)
+                     :do (if (constant-node-p propagated-value-node constant-bindings)
                              (push (cons var propagated-value-node)
                                    new-constant-bindings)
                              (push (cons var propagated-value-node)
                                    nonconstant-bindings)))
-               (let ((*constant-bindings* (append new-constant-bindings *constant-bindings*)))
+               (let ((inner-constant-bindings (append new-constant-bindings constant-bindings)))
                  (if nonconstant-bindings
                      (make-node-let
                       :type (node-type node)
                       :bindings nonconstant-bindings
-                      :subexpr (funcall *traverse* (node-let-subexpr node) call-stack))
-                     (funcall *traverse* (node-let-subexpr node) call-stack)))))
+                      :subexpr (funcall *traverse* (node-let-subexpr node) inner-constant-bindings))
+                     (funcall *traverse* (node-let-subexpr node) inner-constant-bindings)))))
 
-           (propagate-constants-node-lisp (node call-stack)
-             (declare (type node-lisp node)
-                      (ignore call-stack))
+           (propagate-constants-node-lisp (node constant-bindings)
+             (declare (type node-lisp node))
              (let* ((new-lisp-vars nil)
                     (let-bindings
                       (loop :for (lisp-var . coalton-var) :in (node-lisp-vars node)
-                            :for constant-value := (constant-var-value coalton-var :no-error t)
+                            :for constant-value := (constant-var-value coalton-var constant-bindings :no-error t)
                             :if constant-value
                               :collect (let ((new-coalton-var (gentemp (symbol-name coalton-var))))
                                          (push (cons lisp-var new-coalton-var) new-lisp-vars)
@@ -109,11 +103,11 @@
                     :vars new-lisp-vars
                     :form (node-lisp-form node)))))
 
-           (direct-application-better-infer-types (node call-stack)
+           (direct-application-better-infer-types (node constant-bindings)
              ;; FIXME: Can we do something similar for non-direct APPLICATION?
              (let ((constant-propagated-rands
                      (mapcar (lambda (rand)
-                               (funcall *traverse* rand call-stack))
+                               (funcall *traverse* rand constant-bindings))
                              (node-direct-application-rands node))))
                (make-node-direct-application
                 :type (node-type node)
--- a/src/codegen/constant-propagation.lisp
+++ b/src/codegen/constant-propagation.lisp
@@ -124,6 +124,5 @@
         (action (:after node-variable) #'propagate-constants-node-variable)
         (action (:traverse node-let) #'propagate-constants-node-let)
         (action (:after node-lisp) #'propagate-constants-node-lisp)
-        (action (:after node-direct-application) #'direct-application-better-infer-types)
-        (make-traverse-let-action-skipping-cons-bindings))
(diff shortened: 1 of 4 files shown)


src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+
+(defun constant-var-value (var &key no-error)
+  (the (or node null)
+       (or (cdr (assoc var *constant-bindings*))
+           (unless no-error
+             (error "Expected VAR ~S to be a constant but is not" var)))))
+
+(defun propagate-constants (node env)
+  (declare (optimize debug))
+  (labels ((constant-node-p (node)
```
stylewarning, line 33: why don't we lift this out as a toplevel function

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+
+(in-package #:coalton-impl/codegen/constant-propagation)
+
+(defvar *constant-bindings*)
+(setf (documentation '*constant-bindings* 'variable)
+      "Bound inside CONSTANT-PROPAGATION.")
+
+(defun constant-var-value (var &key no-error)
+  (the (or node null)
+       (or (cdr (assoc var *constant-bindings*))
```
stylewarning, line 27: I know it's more typing but can you lift this out into a let+cond?

```
(let ((val ...))
  (cond
    (val ..)
    (no-error ..)
    (t ..)))
```

stylewarning (changes requested):
just two nits for now; i'll think about the algorithm and where this pass sits a little more.

great work!

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+                                   new-constant-bindings)
+                             (push (cons var propagated-value-node)
+                                   nonconstant-bindings)))
+               (let ((*constant-bindings* (append new-constant-bindings *constant-bindings*)))
+                 (if nonconstant-bindings
+                     (make-node-let
+                      :type (node-type node)
+                      :bindings nonconstant-bindings
+                      :subexpr (funcall *traverse* (node-let-subexpr node) call-stack))
+                     (funcall *traverse* (node-let-subexpr node) call-stack)))))
```
amorphedstar, line 78: The new traversal allows arbitrary arguments to be passed through, like the `call-stack` in the current inliner and `bound-variables` in other transformations. In particular, you can use constant-bindings as an argument being passed through the traversal like this:
```suggestion
           (propagate-constants-node-let (node constant-bindings)
             (declare (type node-let node))
             (let ((node-bindings (node-let-bindings node))
                   (new-constant-bindings nil)
                   (nonconstant-bindings nil))
               (loop :for (var . value) :in node-bindings
                     :for propagated-value-node := (funcall *traverse* value constant-bindings)
                     :do (if (constant-node-p propagated-value-node)
                             (push (cons var propagated-value-node)
                                   new-constant-bindings)
                             (push (cons var propagated-value-node)
                                   nonconstant-bindings)))
               (let ((inner-constant-bindings (append new-constant-bindings constant-bindings)))
                 (if nonconstant-bindings
                     (make-node-let
                      :type (node-type node)
                      :bindings nonconstant-bindings
                      :subexpr (funcall *traverse* (node-let-subexpr node) inner-constant-bindings))
                     (funcall *traverse* (node-let-subexpr node) inner-constant-bindings)))))
```
So you don't even need the `*constant-bindings*` outside of everything unless it's better for some other reason

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+                     (mapcar (lambda (rand)
+                               (funcall *traverse* rand call-stack))
+                             (node-direct-application-rands node))))
+               (make-node-direct-application
+                :type (node-type node)
+                :rator-type (tc:make-function-type*
+                             (mapcar #'node-type constant-propagated-rands)
+                             (tc:function-return-type (node-type node)))
+                :rator (node-direct-application-rator node)
+                :rands constant-propagated-rands))))
```
amorphedstar, line 118: Since this is being called `:after` the default traversal for `node-direct-application`, the `node` argument to this function `direct-application-better-infer-types` will already have had the default traversal behavior (constant propagation to sub-nodes using the same args (`constant-bindings`) as this node). So `(funcall *traverse*...)` is unnecessary here:
```suggestion




               (make-node-direct-application
                :type (node-type node)
                :rator-type (tc:make-function-type*
                             (mapcar #'node-type (node-direct-application-rands node))
                             (tc:function-return-type (node-type node)))
                :rator (node-direct-application-rator node)
                :rands (node-direct-application-rands node))))
```

Actually, maybe it would make sense to just change the default traversal action on `node-direct-applications` to do exactly what you have here (just with arbitrary `args`) so that the `rator-type` is always updated when the `rands`' types are updated?

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+
+    (let ((*constant-bindings* nil))
+      (traverse
+       node
+       (list
+        (action (:after node-variable) #'propagate-constants-node-variable)
+        (action (:after node-let) #'propagate-constants-node-let)
```
amorphedstar, line 125: For the `node-let` case, you are defining the recursion behavior yourself, calling `*traverse*` on the `node-let-subexpr` with updated `constant-bindings`, so this should be
```suggestion
        (action (:traverse node-let) #'propagate-constants-node-let)
```

(Otherwise, it will traverse once using the default of unchanged args for traversing the bindings and subexpression, and then `:after` that it will do _another_ recursive traversal)

src/codegen/constant-propagation.lisp
digikar99: > why don't we lift this out as a toplevel function

I assume you are referring to the `constant-node-p` function. With the suggestion that [`traverse` allows passing arbitrary arguments](https://github.com/coalton-lang/coalton/pull/1212#discussion_r1717172995), I think if we pass the `env` argument,  we can lift out `constant-node-p` as a toplevel function. Let me know if that sounds good.

src/codegen/constant-propagation.lisp
digikar99: Committed in https://github.com/coalton-lang/coalton/pull/1212/commits/63a82915937c7eb77faa3429b4fd9e3bcd710a9e.

src/codegen/constant-propagation.lisp
digikar99: Ah, I see, so the original code could have been applicable almost verbatim.

Passing along constant-bindings as a parameters does lead to slightly verbose code. But it feels more consistent with rest of the coalton codebase. 

Committed in https://github.com/coalton-lang/coalton/pull/1212/commits/ada84583b125402868f6cbe2701b7fc2def95bc8.

src/codegen/constant-propagation.lisp
digikar99: So, do you mean

1. Keep the `direct-application-better-infer-types` unchanged
2. Use the `:traverse` action instead of `:after` action

amorphedstar: I mean either

1. Implement the change in the diff and nothing else needs to be changed (`:after` is good as long as you aren't trying to `(funcall *traverse*...)` inside the function to be performed `:after` the main recursive step.)

OR

2. delete the function defined here entirely, and change the action in `make-default-actions` from `src/codegen/traverse.lisp` from
```
    (action (:traverse node-direct-application node &rest args)
      (make-node-direct-application
       :type (node-type node)
       :rator-type (node-direct-application-rator-type node)
       :rator (node-direct-application-rator node)
       :rands (mapcar
               (lambda (node)
                 (apply *traverse* node args))
               (node-direct-application-rands node))))
```
to
```
    (action (:traverse node-direct-application node &rest args)
      (let ((traversed-rands
              (mapcar (lambda (rand)
                        (apply *traverse* rand args))
                      (node-direct-application-rands node))))
        (make-node-direct-application
         :type (node-type node)
         :rator-type (tc:make-function-type*
                      (mapcar #'node-type traversed-rands)
                      (tc:function-return-type (node-type node)))
         :rator (node-direct-application-rator node)
         :rands traversed-rands))))
```
(basically the code you wrote here, but general)
(I am more unsure about this. Maybe it's possible that we would often want applications of `traverse` which don't touch the `rator-type` at all, even if the individual rand `node-type`s get changed?)

--
I guess it would be safe to just go with 1 for now, unless someone thinks 2 would be useful for many other types of traversals.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+           may-be-value)
+          (no-error
+           nil)
+          (t
+           (error "Expected VAR ~S to be a constant but is not" var)))))
+
+(defun propagate-constants (node env)
+  (declare (optimize debug))
+  (labels ((constant-node-p (node constant-bindings)
+             ;; FIXME: Do we need more nodes classified as constants?
```
amorphedstar, line 36: It might be good to treat things like `(lisp IFix () cl:most-positive-fixnum)` as constants, but idk how hard it is to dig into a `node-lisp` to check

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+
+    (let ((*constant-bindings* nil))
+      (traverse
+       node
+       (list
+        (action (:after node-variable) #'propagate-constants-node-variable)
+        (action (:traverse node-let) #'propagate-constants-node-let)
+        (action (:after node-lisp) #'propagate-constants-node-lisp)
+        (action (:after node-direct-application) #'direct-application-better-infer-types)
+        (make-traverse-let-action-skipping-cons-bindings))
```
amorphedstar, line 128:
```suggestion
        (action (:after node-direct-application) #'direct-application-better-infer-types))
```
I can't think of a reason why the constant propagation should need to skip `let`-bindings to `Cons` applications (this special action was a very particular fix to make sure the inliner wouldn't break recursive `Cons` bindings)

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+                    (constant-var-value (node-variable-value node) constant-bindings :no-error no-error))
+                   (no-error
+                    nil)
+                   (t
+                    (error "Expected the node to be a constant but is not ~%~S" node))))
+
+           (propagate-constants-node-variable (node constant-bindings)
+             (declare (type node-variable node))
+             (or (constant-node-value node constant-bindings :no-error t)
+                 node))
```
amorphedstar, line 59: All traversal actions will return the original node if the function returned `nil`, so this could also be written just as
```suggestion
             (constant-node-value node constant-bindings :no-error t)
```
(and at that point maybe having it be a separate function is unnecessary?)

src/codegen/constant-propagation.lisp
stylewarning: maybe a simple cl:constantp call on the innards of node-lisp. might yield false negatives but that's fine.

digikar99: Given that `alexandria:define-constant` accepts a `:test` argument, should constants be treated as domain specific? So, we will have `*constant-predicates*` as a list of predicates. This would be set to a certain default, but users can modify it per their application.

stylewarning: what do you have in mind for the contents of that list?

digikar99: When I wrote about `*constant-predicates*`, I was thinking that the contents of current `constant-node-p` can be broken into several predicates, including `node-literal-p` `constant-var-p` etc.

But, I guess we first need a better notion of a constant. Some related thoughts include: 

1. Does coalton have a predicate that checks identity of objects? 
2. What do the node-literals actually compile to?

For instance, suppose `==` checked for the identity of objects. Would the results of the `==-caller` below depend on whether or not constant propagation takes place? (You can replace the `"hello"` with something more commonly mutable, like a single-float array.)

```lisp
(coalton-toplevel  
  (define (==-caller)
    (let ((x "hello"))
      (== x x))))
```

stylewarning: There's an unsafe equality predicate: `UNSAFE-POINTER-EQ?`

Coalton hasn't been more definitive about identity. The spirit of `==` is definitely that it should be a value-equality.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,129 @@
...
+                      (constant-var-value (node-variable-value node) constant-bindings :no-error t))))
+           (constant-node-value (node constant-bindings &key no-error)
+             ;; FIXME: Do we need more nodes classified as constants?
+             (cond ((node-literal-p node)
+                    node)
+                   ((and (node-variable-p node)
+                         (tc:lookup-instance-by-codegen-sym env (node-variable-value node) :no-error t))
+                    node)
+                   ((node-variable-p node)
+                    (constant-var-value (node-variable-value node) constant-bindings :no-error no-error))
```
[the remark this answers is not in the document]
digikar99: The intended use case for `tc:lookup-instance-by-codegen-sym` was to handle `dict`-gentemps which eventually get bound to a codegen-sym of a class-instance. This case arose when trying to optimize the below code. 

Ideally, I'd like a check that a given symbol is a `dict`-gentemp corresponding to a codegen-sym of some instance.

Regarding the precedence change, the question then is: is it possible for such `dict`-gentemps to be bound to a non-global value? If yes, the precedence change makes sense. If not, then I think the current precedence keeps code a bit cleaner.

```lisp
(coalton-toplevel
  (inline)
  (define (partial-monomorphization-a x y z)
    (== x y)
    (+ 1 z)))

(coalton-toplevel
  (declare partial-monomorphization-b (Eq :a => :a -> :a -> Integer -> Integer))
  (monomorphize)
  (inline)
  (define (partial-monomorphization-b x y z)
    (partial-monomorphization-a x y z)))

(coalton-toplevel
  (monomorphize)
  (define (partial-monomorphization-b-caller)
    (partial-monomorphization-b "a" "a" 2)))
```

src/codegen/constant-propagation.lisp
digikar99: Implemented in https://github.com/coalton-lang/coalton/pull/1212/commits/37b690040c08e8f380871764e8458478d1404a58.

src/codegen/constant-propagation.lisp
digikar99: I think for this PR, it would make sense to go with option 1. 

Option 2 may be tested out internally against other codebases using coalton, and then, if nothing breaks, it should be okay. I presume all the type checks and inference steps have already taken place at this point of code. So, I don't see option 2 breaking something. If anything, it might just lead to slightly more stringent checks for whatever the traverse function happens to do. But this seems separate from this PR.

digikar99: Implemented option 1 in 2371482.

src/codegen/constant-propagation.lisp
digikar99: About `*constant-predicates*`, this involves replacing the current `constant-node-p` with approximately following:

```lisp
(declaim (type list *constant-predicates*))
(defvar *constant-predicates* nil
  "A list of two-argument functions. Each function should take a NODE and a list of CONSTANT-BINDINGS and return
- NIL if the NODE is not constant
- the constant node corresponding to NODE if it is a constant")

(defun constant-node-p (node constant-bindings)
  (declare (type node node)
           (type list constant-bindings))
  (loop :for predicate :in *constant-predicates*
          :thereis (funcall predicate node constant-bindings)))

(defmacro define-constant-predicate (name (node-var &optional (constant-bindings-var nil cbp))
                                     &body body)
  (let ((constant-bindings-var (or constant-bindings-var (gensym "CONSTANT-BINDINGS"))))
    `(progn
       (defun ,name (,node-var ,constant-bindings-var)
         (declare (type node ,node-var))
         ,(unless cbp `(declare (ignore ,constant-bindings-var)))
         ,@body)
       (pushnew ',name *constant-predicates*))))

(define-constant-predicate node-literal-constant-predicate (node)
  (when (node-literal-p node)
    node))
(define-constant-predicate node-variable-constant-p (node constant-bindings)
  (when (node-variable-p node)
    (constant-var-value (node-variable-value node) constant-bindings :no-error t)))
(define-constant-predicate node-lisp-constant-p (node)
  (when (node-lisp-p node)
    (cl:constantp (node-lisp-form node))))
```

There would be another predicate bound inside `propagate-constants` which would check whether a given node-variable is bound to a instance-codegen-sym.

amorphedstar: I still don't quite understand in what cases different things would need to be considered constants, so an explicit example of that could be helpful. But that code looks generally reasonable, aside from some inconsistency like `constant-node-p` returning type `(values (or boolean node) &optional)`

amorphedstar (commented):
looks pretty good to me, just I don't know the answers to most of the FIXME questions

src/codegen/constant-propagation.lisp
digikar99: Let me try, but I think I am only able to think about it in slightly roundabout ways. 

So, there seems to be a difference between how SLIME/SBCL (sbcl-2.4.5 atleast) treats literal objects during compilation from an emacs buffer vs evaluation in the REPL. For example, if you compile the below in a buffer `(coalton (pointer-eq-caller))` returns `T`. If you evaluate the below in a REPL, `(coalton (pointer-eq-caller))` returns `NIL`.

```lisp
(coalton-toplevel
  (define (pointer-eq-caller)
    (let ((x "hello")
          (y "hello"))
      (unsafe-pointer-eq? x y))))
```

Fortunately, it seems coalton already recognizes this and the pointer equality check is therefore called `unsafe`.

When I originally wrote about the extensible `*constant-predicates*`, I also wanted to provide ways for constant propagation to be more inclusive. For instance, if there were a coalton type corresponding to `cffi:foreign-pointer`, then `(cffi:null-pointer)` should be treated as a constant for all purposes. Similarly, objects of a `color` type defined by

```lisp
(define-type color
  (make-color u8 u8 u8))
```

would also ideally be treated as a constant, since the color only depends on the RGB values. All of this probably involves looking at a `node-application` or `node-direct-application` and then determine if it satisfies the (domain/task-specific) constraints to be a constant.

On the other hand, we probably don't want the `node-application` or `node-direct-application` corresponding to `(cell:new ...)` to be treated as a constant.

---

I think

1. If literal objects are always immutable
2. If we disallow treating `node-application` or `node-direct-application` as immutable objects

In those cases, we probably do not need the extensible notion of `*constant-predicates*`. Incorporating 2 makes me a bit sad. But it can be overcome by providing a literal representation of an object. So, above, the `color` or `cffi:foreign-pointer` may have literal representations which get treated as constants. But `(cell:new ...)` shouldn't have a literal representation.

stylewarning:
For this PR, my suggestion is to keep it very simple:

- only propagate obvious numerical and boolean constants
- add some conds/"hooks" for places where we can do more elaborate checks (but don't)
- leave out an extensible notion of constantp 

I think we should attack constantness one case at a time so we can thoroughly deliberate it. For now, let's just get the basic behavior in.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,122 @@
...
+           (error "Expected VAR ~S to be a constant but is not" var)))))
+
+(defun constant-node-p (env node constant-bindings)
+  "If NODE is a constant, returns the constant value corresponding to it.
+If not, returns NIL"
+  (declare (type tc:environment env)
+           (type node node)
+           (type list constant-bindings))
+  (cond ((node-literal-p node)
+         (if (numberp (node-literal-value node))
```
amorphedstar, line 36: Why only numbers? The allowed types are:
```lisp
(deftype literal-value ()
  "Allowed literal values as Lisp objects."
  '(or integer ratio single-float double-float string character))
```

I guess if you are considering strings as not-necessarily-constant depending on whether we care about pointer equality, then it makes sense to not include strings by default. But shouldn't characters still be considered constant?

```suggestion
         (if (or (numberp (node-literal-value node)) (characterp (node-literal-value node)))
```

digikar99: It seems that even the case of numbers is contentious. ([Ref.](http://clhs.lisp.se/Body/f_eq.htm))

> The effect is that Common Lisp makes no guarantee that eq is true even when both its arguments are ``the same thing'' if that thing is a character or number.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,122 @@
...
+             node
+             nil))
+        ((node-variable-p node)
+         (cond ((and (eq 'coalton:boolean (node-type node))
+                     (member (node-variable-value node) '(coalton:true coalton:false)))
```
amorphedstar, line 41: Other places in the codebase capitalize these
```suggestion
         (cond ((and (eq 'coalton:Boolean (node-type node))
                     (member (node-variable-value node) '(coalton:True coalton:False)))
```

src/codegen/constant-propagation.lisp
digikar99: I don't have an opinion on this. I will be happy with whatever is suitable for coalton. If I recall, the main goal of constant propagation was to obtain the value of the `dict`-gensyms bound to codegen-symbols of the instances. And this has been achieved :D.

stylewarning: I'd say anything that works with CL:EQL (excluding CL:EQ) are obvious candidates. Those would be:

- any CL number type
- characters 

(http://www.ai.mit.edu/projects/iiip/doc/CommonLISP/HyperSpec/Body/fun_eql.html)

Booleans work because of `cl:eq` but they're important enough for inlining etc. that they ought to be propagated.

Honestly though, I'm fine to punt on any cases.

stylewarning:
@amorphedstar @digikar99 Any remaining outstanding items?

digikar99:
> boolean constants

Do you mean CL booleans or Coalton booleans? If the former, could you provide a sample code where they get used in Coalton code?

src/codegen/optimizer.lisp
```diff
@@ -206,6 +209,8 @@
    node
    canonicalize
    (match-dynamic-extent-lift env)
+   ;; FIXME: Is this the right place to propagate-constants?
+   (propagate-constants env)
```
digikar99, line 213: I had put it here intuiting that it should happen as early as possible so other passes can benefit from it. At the same time, it should not change the correctness for any of the passes. I did not put it before the `match-dynamic-extent-lift` just to be on the safe side.

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,122 @@
...
+                              :form (node-lisp-form node)))
+                   (make-node-lisp
+                    :type (node-type node)
+                    :vars new-lisp-vars
+                    :form (node-lisp-form node)))))
+
+           (direct-application-better-infer-types (node constant-bindings)
+             (declare (type node-direct-application node)
+                      (ignore constant-bindings))
+             ;; FIXME: Can we do something similar for non-direct APPLICATION?
```
digikar99, line 106: Just saw that `node-application` does not have a `rator-type`. Then, we can straightaway remove this FIXME?

```suggestion
```

amorphedstar: I mean, the `node-application` still has a `rator` which is a node which has a `type`, but I don't know whether anything would be gained from trying to rewrite its type based on any operand types

amorphedstar: I mean if you look at the definitions
```lisp
(defstruct (node-variable (:include node))
  "Variables like x or y"
  (value (util:required 'value) :type parser:identifier :read-only t))

(defstruct (node-application (:include node))
  "Function application (f x)"
  (rator (util:required 'rator) :type node      :read-only t)
  (rands (util:required 'rands) :type node-list :read-only t))

(defstruct (node-direct-application (:include node))
  "Fully saturated function application of a known function"
  (rator-type (util:required 'rator-type) :type tc:ty             :read-only t)
  (rator      (util:required 'rator)      :type parser:identifier :read-only t)
  (rands      (util:required 'rands)      :type node-list         :read-only t))
```
you see that the `node-application`'s `rator` is an actual `node`, while the `node-direct-application`'s `rator` is just a `parser:identifier`, so it needs another field `rator-type` to specify the corresponding type. It would probably be an equivalent idea to write
```lisp
(defstruct (node-direct-application (:include node))
  "Fully saturated function application of a known function"
  (rator (util:required 'rator) :type node-variable :read-only t)
  (rands (util:required 'rands) :type node-list     :read-only t))
```
since this still requires that the `rator` is a "known function" / named variable. I imagine the reason for flattening it in the current `node-direct-application` is just for accessor convenience.

digikar99: Since the `rator` is a `node` for the `node-application`, can we expect that the default traverse would update the types?

src/codegen/optimizer.lisp
stylewarning: @eliaslfox any thoughts?

src/codegen/constant-propagation.lisp
amorphedstar: No, as I mentioned in another comment, the default traversal only traverses through nodes, leaving types unchanged (see `src/codegen/traverse.lisp`:
```lisp
    (action (:traverse node-application node &rest args)
      (make-node-application
       :type (node-type node)
       :rator (apply *traverse* (node-application-rator node) args)
       :rands (mapcar
               (lambda (node)
                 (apply *traverse* node args))
               (node-application-rands node))))
    (action (:traverse node-direct-application node &rest args)
      (make-node-direct-application
       :type (node-type node)
       :rator-type (node-direct-application-rator-type node)
       :rator (node-direct-application-rator node)
       :rands (mapcar
               (lambda (node)
                 (apply *traverse* node args))
               (node-direct-application-rands node))))
```
If you want to change the types of applications' `rator`s to always match the operands, you could rewrite these to update them, or you could also just redefine the function `node-rator-type` in `ast.lisp` to compute the type based on the arguments every time it gets called

digikar99: I think I asked the wrong question. I should have asked only about the `rator` not about its type. For `node-application`, we do not need to do anything more than what is currently being done by the default traverse functions to update the `rator`. 

For `node-direct-application`, we decided for the PR-specific solution.

I am now having a feeling that type updation can also be carried out for node-let, and probably for several other nodes. A PR incorporating this will also update the type of `rator` for `node-application`.

I also noted that `direct-application-better-infer-types` is never called while loading coalton-library even if heuristic inlining is enabled. However, it gets called during monomorphization, such as below:

```lisp
(coalton-toplevel
  (define (partial-monomorphization-a x y z)
    (== x y)
    (+ 1 z)))

(coalton-toplevel
  (declare partial-monomorphization-b (Eq :a => :a -> :a -> Integer -> Integer))
  (monomorphize)
  (define (partial-monomorphization-b x y z)
    (partial-monomorphization-a x y z)))
```

amorphedstar: Yes, I think a `node-direct-application` only gets _created_ when calling `src/codegen/optimizer.lisp`'s `direct-application` function. A search shows just two times where `direct-application` is called:

- inside of `src/entry.lisp`'s `expression-entry-point`:
```lisp
...
            (when (null preds)
              (return-from expression-entry-point
                (let ((node (codegen:optimize-node
                             (codegen:translate-expression node nil env)
                             env)))
                  (codegen:codegen-expression 
                   (codegen:direct-application
                    node
                    (codegen:make-function-table env))
                   env))))
...
```

- inside `src/codegen/optimizer.lisp`'s `optimize-bindings`:
```lisp
...
  (let ((bindings (optimize-bindings-initial bindings package env)))
...
      (setf bindings
                  :if (gethash name monomorphize-table)
                    :append (optimize-bindings-initial
                             (monomorphize
                              name
                              manager
                              package
                              resolve-table
                              (lambda (node env)
                                (optimize-node node env))
                              env)
                             package env)
                  :else
                    :collect (cons name node)))
...
        (setf bindings
                    :collect (cons name (direct-application node function-table))))
...
```

In both of these cases, `optimize-node` (and/or `optimize-bindings-initial`) gets called _before_ `direct-application`. So I guess it's only possible for the functions in `optimize-node` to receive a `node-direct-application` if a recursive call to the optimizer has already been completed, like how the monomorphizer recursively calls back to the optimizer?

src/codegen/constant-propagation.lisp
```diff
@@ -0,0 +1,122 @@
...
+  (declare (type symbol var)
+           (values (or node null) &optional))
+  (let ((may-be-value (cdr (assoc var constant-bindings))))
+    (cond (may-be-value
+           ;; Whenever it exists, it is always a NODE (thus, non-NIL)
+           may-be-value)
+          (no-error
+           nil)
+          (t
+           (error "Expected VAR ~S to be a constant but is not" var)))))
```
amorphedstar, line 27:
```suggestion
           (error "Expected VAR ~S to be a constant but it is not" var)))))
```

src/codegen/constant-propagation.lisp
digikar99: Ah, yes indeed! This point about when `node-direct-application` are created had also risen in earlier discussions with stylewarning and eliaslfox. Your elaboration pretty much sums it up.

But coming back to resolving this. I see two steps -

1. For this PR, leave both node-application and node-direct-application untouched, unless it directly concerns constant propagation.
2. In a separate PR, incorporate type updation in the default traverse functions.

src/codegen/optimizer.lisp
amorphedstar: The PR looks good to me as it is, but I agree that it would be good for @eliaslfox to take a look and make sure it is in the right place relative to the rest of the optimizations (and also verify that updating the `node-rator-type` is helpful here for `node-direct-application` and maybe also `node-application`?)

eliaslfox (approved):
This seems like a reasonable implementation.

stylewarning (approved):

PR #1216, opened 2024-08-23
Source Locations in Documentation

jbouwman:
There are two major parts to this change set:

The documentation generator is substantially rewritten to provide direct links to definitions. This works in the existing hugo web documentation, and supports emacs jump-to-definition. Some outstanding issues are also fixed, for instance, sorting of indexes and entries.

For most AST nodes, the source-span typed 'source' slot is replaced with a 'location' struct that has a persistable reference to the source file, as well as a span. Renaming of slot accessors and initialization keywords accounts for the majority of line changes outside of doc generator code.

source-error conditions are changed such that the source-error structure is built in-place, rather than constructed and returned by a lambda. This is also done in order to support introspectable conditions on behalf of emacs integration.


Izaakwltn (approved):
Looks good to me

jbouwman:
Obtained secondary approval from @macrologist

PR #1217, opened 2024-08-28
Source Error Slots

jbouwman:
Normalize source condition notes by merging source-condition class and source-error struct.

- A condition has a string-valued message, and zero or more notes, generally one.
- A note is minimally a string and a location

Improve the checking of type inference errors: look for the exact text, rather than the condition.


src/reader.lisp
```diff
@@ -26,8 +44,10 @@ Used to forbid reading while inside quasiquoted forms.")
 
   (parser:with-reader-context stream
     (let* ((file
-             (coalton-impl/source:make-source-file *compile-file-truename*
-                                                   :offset (file-position stream)))
+             (coalton-impl/source:make-source-file (source-file)
```
jbouwman, line 47: The source file backing the buffer may or may not be saved: the error annotations need to be based solely on the string passed to swank.

jbouwman:
generic function definitions here have been moved to separate PRs, other portions are obsolete

PR #1223, opened 2024-09-03
source:location generic function

jbouwman:
Make location generic across parser and typechecker nodes using :include to provide slot definitions.

This change is similar to the introduction of #'source:docstring in #1220 and has similar motivation.

- four existing generics are merged
- use of :include reduces the number of method definitions
- error constructors become more generic

### DIFF
--- a/src/entry.lisp
+++ b/src/entry.lisp
@@ -110,7 +110,7 @@
           (setf subs (tc:compose-substitution-lists subs subs_))
 
           (when accessors
-            (tc:tc-error (tc:accessor-location (first accessors))
+            (tc:tc-error (first accessors)
                          "Amqbiguous accessor"
                          "accessor is ambiguous"))
 
@@ -145,7 +145,7 @@
                                tvars
                                (tc:ty-scheme-type scheme))))
 
-              (tc:tc-error (tc:node-location node)
+              (tc:tc-error node
                            "Unable to codegen"
                            (format nil
                                    "expression has type ~A~{ ~S~}.~{ (~S)~} => ~S with unresolved constraint~A ~S"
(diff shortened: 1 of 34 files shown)


Izaakwltn (approved):
Seems like a reasonable (and more readable) change

stylewarning (changes requested):
My instinct is that using :include will lead to brittle design. What do we do when we want everybody to have more common slots? Behaviorally the subtype relationship also doesn't make much logical sense (it's not really an "is a" relationship).

Can we do this without using "include"? I feel all we are saving is typing.

stylewarning:
I should add that if this were a class, I would find it less objectionable, but I would prefer not to turn `node` types into classes since the compiler is already difficult to debug.

jbouwman:
Yes, it's easy to keep :include out of this and in retrospect preferable. Revising.

stylewarning:
> Yes, it's easy to keep :include out of this and in retrospect preferable. Revising.

One possibility is to create more superclasses. Maybe include a location on NODE, maybe another on something like TOPLEVEL-FORM (which we'd have to make), etc. That way we have sensible and organized classes whilst not having a lot of duplication.

src/entry.lisp
```diff
@@ -110,7 +110,7 @@
           (setf subs (tc:compose-substitution-lists subs subs_))
 
           (when accessors
-            (tc:tc-error (tc:accessor-location (first accessors))
+            (tc:tc-error (first accessors)
                          "Amqbiguous accessor"
```
stylewarning, line 114: typo

stylewarning (changes requested):
- rename coalton-object
- split tc-error into a location case and locatable case

stylewarning (approved):

PR #1229, opened 2024-09-06
Add test suite test case numbering

jbouwman:
Support execution of single tests using COALTON-TESTS:RUN-SINGLE


tests/loader.lisp
```diff
@@ -64,11 +80,25 @@
...
                (format nil "~{~A~%~}" (reverse lines))))
 
+(defun first-integer (string)
+  "Return the first integer appearing in STRING, or nil."
+  (flet ((numeric? (c)
+           (<= 48 (char-code c) 57)))
+    (let ((start (position-if #'numeric? string)))
+      (when start
+        (let* ((string (subseq string start))
+               (end (position-if (complement #'numeric?) string)))
```
Izaakwltn, line 90: This maybe should check for decimal points.

Probably unecessary, if we are diligent about adding numbers to the beginning of tests, but what if the name of the test is something like, `Ensuring proper behavior sbcl >2.4`:

`(first-integer "Ensuring proper behavior sbcl >2.4`")` returns 2.

tests/package.lisp
```diff
@@ -13,7 +13,9 @@
    (#:analysis #:coalton-impl/analysis)
    (#:entry #:coalton-impl/entry))
   (:export
-   #:run-coalton-tests))
+   #:run-coalton-tests
+   #:run-suite
```
Izaakwltn, line 17: I think this might possibly be a confusing name, maybe it should be `#:run-file-tests` or `#:run-test-page` or `#:run-test-file` something like that. Suite would make me assume it's larger than a single file.

tests/utilities.lisp
```diff
@@ -113,8 +117,24 @@ Returns (values SOURCE-PATHNAME COMPILED-PATHNAME)."
...
+test header: ~A~%~
 expected error (A) and generated error (B)"
-                                            file line description)
+                                            file line
+                                            (or number "(unassigned)")
+                                            description)
                                     expected-error
                                     generated-error))))))
+
+(defun run-single (pathname test-number)
```
Izaakwltn, line 129: I think this could just be `run-test` unless that name is already taken.

Izaakwltn (commented):
A few nits

tests/utilities.lisp
jbouwman: done

tests/package.lisp
jbouwman: done, changed to run-test-file

tests/loader.lisp
jbouwman: since this is mainly here to facilitate a developer's selection of a single test case, I think it's safe to presume they would assign an unambiguous number to a failing case, via a local edit.

At some point, a stricter header structure could be imposed -- I'm inclined to leave it loosely defined for now.

Izaakwltn (approved):
Looks good

stylewarning (approved):

PR #1233, opened 2024-09-09
Convert tests to compare printed conditions

jbouwman:
Convert tests that check that an error is signaled to instead check for specific error text. To obtain consistent messages:

- disable Unicode output while running tests that compare error messages, because ACL doesn't print unicode
- sort SCCs by source location when reporting cyclic superclass errors, because ACL and CCL's collection ordering are different from SBCLs

For SCC class definition sorting, Introduce span, location and source comparators and generics to source.lisp, a limited version of what appears in #1223


src/source.lisp
```diff
@@ -53,6 +54,10 @@
 
 ;; source input
 
+(defgeneric source< (a b)
+  (:method (a b)
```
Izaakwltn, line 58: This could probably use a docstring

src/source.lisp
```diff
@@ -140,12 +149,39 @@ OFFSET indicates starting character offset within the file."
...
   (source nil
    :read-only t)
   (span nil
-   :type (cons fixnum fixnum) :read-only t))
+   :type span :read-only t))
+
+(defgeneric location (object)
+  (:documentation "The location of a Coalton object's source definition."))
+
+(defun location< (a b)
```
Izaakwltn, line 179: This could also use a docstring

jbouwman: done

src/source.lisp
jbouwman: done

Izaakwltn (approved):

PR #1239, opened 2024-09-12
Ensure presence of source metadata in open paren dispatch

### LINKED ISSUE #1237: type-error during printing of typechecker error
Running the following from the README example (in the repl):
```
COALTON-USER> (coalton (dt (E+ (EConst 1/2) (EConst 0.5))))
```
```
Error (TYPE-ERROR) during printing: #<COALTON-IMPL/TYPECHECKER/BASE:TC-ERROR {702B19F4F3}>
   [Condition of type COALTON-IMPL/TYPECHECKER/BASE:TC-ERROR]
```

This only applies when run from the REPL, when loaded from a file it gives the expected error.

jbouwman:
- add guard to source error printer to prevent recursive error
- break up and document read-coalton-toplevel-open-paren

Closes #1237


Izaakwltn (approved):
Looks good to me- 

Would it be possible to add a test case with an error example? Or since it's a repl-based error is that infeasible/unwieldly?

jbouwman:
Test added

PR #1241, opened 2024-09-12
Fixed code examples and error message examples in docs

Izaakwltn:


jbouwman (approved):

docs/intro-to-coalton.md
```diff
@@ -493,16 +493,14 @@ COALTON-USER> (type-of '/)
 ∀ :A :B. DIVIDABLE :A :B ⇒ (:A → :A → :B)
 ```
 
-Because of the generic nature of division, if you're computing some values at the REPL, "raw division" simply does not work.
```
stylewarning, line 496: This isn't quite the right explanation.

docs/intro-to-coalton.md
```diff
@@ -511,13 +509,17 @@ COALTON-USER> (coalton (the Fraction (/ 4 2)))
 #.(COALTON-LIBRARY::%FRACTION 2 1)
 ```
 
-But division of integers does not work.
+But an Integer result from division is not possible.
```
stylewarning, line 512: Integer in backquotes

stylewarning (changes requested):
most is good, needs some tweaking

docs/intro-to-coalton.md
Izaakwltn: I think I made it better

docs/intro-to-coalton.md
```diff
@@ -491,16 +491,14 @@ COALTON-USER> (type-of '/)
 ∀ :A :B. DIVIDABLE :A :B ⇒ (:A → :A → :B)
 ```
 
-Because of the generic nature of division, if you're computing some values at the REPL, "raw division" simply does not work.
+Division defaults to `Double-Float` operations when given `Integer` values. [see Instance Defaulting](#instance-defaulting).
```
stylewarning, line 494:
```
Because of [instance defaulting](...), division of integer constants without any additional context defaults to `Double-Float` division.
```

or something like that.

docs/intro-to-coalton.md
```diff
@@ -491,16 +491,14 @@ COALTON-USER> (type-of '/)
...
 ```
 COALTON-USER> (coalton (/ 1 2))
-Failed to reduce context for DIVIDABLE INTEGER :A
-in COALTON
-   [Condition of type COALTON-IMPL/TYPECHECKER::COALTON-TYPE-ERROR-CONTEXT]
+0.5d0
 ```
 
-You have to constrain the result type of the `Dividable` instance. You can do this with the `the` operator. There are lots of `Dividable` instances made for you.
+In order to get a non-Double-Float result, you have to constrain the result type of the `Dividable` instance. You can do this with the `the` operator. There are lots of `Dividable` instances made for you.
```
stylewarning, line 501: Show an example of this.

Re-word slightly, something like: "We can tell Coalton our constants are of another type by constraining them with `the`, or relying on type inference. The type must be an instance of `Dividable`" For example, ..."

Remember to read these things as if you're an absolute beginner trying to get work done.

docs/intro-to-coalton.md
```diff
@@ -509,13 +507,17 @@ COALTON-USER> (coalton (the Fraction (/ 4 2)))
 #.(COALTON-LIBRARY::%FRACTION 2 1)
 ```
 
-But division of integers does not work.
+But an `Integer` result from division is not possible.
```
stylewarning, line 510: division with `/`

stylewarning (changes requested):

docs/intro-to-coalton.md
Izaakwltn: I believe there are two examples right below? One for `Single-Float` and another for `Fraction`

stylewarning: Whoops, didn't see in the diff. Sorry!

stylewarning (approved):

PR #1245, opened 2024-09-17
Use a package local nickname for coalton-impl/source

jbouwman:
Be more consistent with use of this package by always importing it using a nickname, rather than in the package :use clause.

Aside from changes to defpackage, this is a purely mechanical change.


Izaakwltn (approved):
`source:looks-good-to-me`

PR #1247, opened 2024-09-18
Use source:location references when constructing conditions

jbouwman:
- make cursor 'note' generic by relocating it to src/source
- parse-error constructor with `message` `note*` arguments

### DIFF
--- a/tests/parser-test-files/package.txt
+++ b/tests/parser-test-files/package.txt
@@ -36,7 +36,7 @@ error: Malformed package declaration
   --> test:1:8
    |
  1 |  (package)
-   |          ^ missing package name
+   |         ^ missing package name
 
 ================================================================================
 4 Unknown package clause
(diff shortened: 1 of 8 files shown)


tests/parser-test-files/package.txt
```diff
@@ -36,7 +36,7 @@ error: Malformed package declaration
   --> test:1:8
    |
  1 |  (package)
-   |          ^ missing package name
```
Izaakwltn, line 39: Personally I kind of prefer the pointer to be under the `)` because it makes it clear that it's pointing to the next element.

Izaakwltn (approved):

tests/parser-test-files/package.txt
jbouwman: In this case the note constructor is given the zero-indexed span (8 . 8) which ought to indicate the gap between `e` and `)`, but the condition printer is emitting 7 spaces instead of 8. I believe there are a few dozen other cases in the text-formatted error test cases. I'll open a separate cleanup issue for it.

jbouwman: This issue filed as #1248

PR #1256, opened 2024-09-20
Add instance `Quantizable CReal`

Izaakwltn:
Adds instance `Quantizable CReal`, fixing `Real CReal` dependency.


Izaakwltn:
Should fix CI for #1253

stylewarning (approved):

PR #1260, opened 2024-09-24
Add class method docstrings to lisp symbol documentation

Izaakwltn:
This adds method docstrings to lisp symbol documentation:
```
COALTON-LIBRARY/CLASSES> (cl:describe 'foldr)
COALTON-LIBRARY/CLASSES:FOLDR
  [symbol]

FOLDR names a symbol macro:
  Expansion: (COMMON-LISP:THE
              COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY
              (COALTON-IMPL/GLOBAL-LEXICAL::TOP-LEVEL-BINDING-VALUE
               (COMMON-LISP:LOAD-TIME-VALUE
                (COALTON-IMPL/GLOBAL-LEXICAL:GET-TOP-LEVEL-BINDING
                 'FOLDR
                 COALTON-IMPL/GLOBAL-LEXICAL::*TOP-LEVEL-ENVIRONMENT*))))
  Documentation:
    FOLDR :: ∀ A B C. FOLDABLE C ⇒ ((A → B → B) → B → (C A) → B)
    A right non-tail-recursive fold.
    

FOLDR names a compiled function:
  Lambda-list: (DICT _0 _1 _2)
  Derived type: (FUNCTION (T T T T) *)
  Documentation:
    FOLDR :: ∀ A B C. FOLDABLE C ⇒ ((A → B → B) → B → (C A) → B)
    A right non-tail-recursive fold.

  Inline proclamation: INLINE (inline expansion available)
```

PR #1257, opened 2024-09-20
Add constructor docstring support

Izaakwltn:
This adds docstrings for type constructors.

PR #1261, opened 2024-09-25
Adds a guide to documentation in Coalton

Izaakwltn:
This is a simple but comprehensive explanation of where to put documentation in Coalton code, along with some explanations of why different forms of documentation are more useful than others. 

These explanations should probably be expanded (perhaps with an actual description of Coalton doc generation and more detail about Lisp symbol documentation).

This guide includes features that will only be available after merging #1257 and #1260.


stylewarning (approved):

jbouwman (approved):
excellent, merge conditional on 2x dependent PRs

PR #1262, opened 2024-09-26
Include variable name in error message if missing in environment

Mira-public:
I've encountered inaccurate source locations in errors. It can be difficult to tell which symbol I need to import a package or add a package prefix for. So this adds the variable name to the error.

### DIFF
--- a/src/typechecker/tc-env.lisp
+++ b/src/typechecker/tc-env.lisp
@@ -74,8 +74,8 @@
       (error 'tc:tc-error
              :err (source:source-error
                    :location (source:location var)
-                   :message "Unknown variable"
-                   :primary-note "unknown variable"
+                   :message (format nil "Unknown variable ~a" var-name)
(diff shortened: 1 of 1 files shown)


stylewarning:
@Mira-public I too ran into some source location regressions and had trouble finding the variable names. (It doesn't highlight the right part of the code and it's difficult to tell where the error actually was.)

There's been a lot of recent work on error messages and we slipped up a bit. Working to fix the root issue.

CC @jbouwman

stylewarning:
@Mira-public The tests are going to fail because we check within them that we are getting the right error messages. Probably what we'll do
is, if this change is approved, we'll regen the tests and get the change in, possibly on a separate branch.

stylewarning:
CC @YarinHeffes who ran into similar issues.

jbouwman:
I've added the updated tests to the follow on merge request #1264

jbouwman:
Change incorporated via #1264 - thank you for the contribution!

Issue #423, opened 2022-03-12
Allow matching on other number types


eliaslfox:
Extend overloaded integer literals to patterns

Jobhdez:
can someone add more information as to what this issue is about? thanks

eliaslfox:
Currently integer literal patterns must be of type `Integer`. Ideally it would be possible to pattern match on anything that implements `Num`.

This works:
```
 (coalton-toplevel
  (define (f x)
    (match x
      (5 "five")
      (_ "not five"))))
```

This does not:
```
 (coalton-toplevel
  (declare f (IFix -> String))
  (define (f x)
    (match x
      (5 "five")
      (_ "not five"))))
```

Jobhdez:
Thank you.

gefjon:
> Ideally it would be possible to pattern match on anything that implements Num.

`Num` and `Eq`, I would think.

eliaslfox:
`Num` implies `Eq`

PR #1266, opened 2024-09-27
Allow matching on `Num` types

YarinHeffes:
Addresses #423


Izaakwltn:
I think this could use a few tests, matching on different `Num` types:

https://github.com/coalton-lang/coalton/blob/main/tests/pattern-matching-tests.lisp

YarinHeffes:
> I think this could use a few tests, matching on different `Num` types:
> 
> https://github.com/coalton-lang/coalton/blob/main/tests/pattern-matching-tests.lisp

I added some tests. This PR does not allow matching on non-literal `Num` types.

stylewarning (approved):

PR #1267, opened 2024-09-27
Refer to correct source during macroexpansion

### LINKED ISSUE #1263: Incorrect source location reported in error message
Several users report incorrect source locations in error messages: further repro and tracking information to be added here as it emerges.

cc @Mira-public @YarinHeffes @stylewarning @macrologist

jbouwman:
Bind *source* during expansion of coalton macros so that offsets refer to correct source location.

This restores the previous behavior of macroexpansion:

```
  error: 
    during macroexpansion of (COALTON HELLO). Use COMMON-LISP:*BREAK-ON-SIGNALS* to
    intercept.
    
     error: Unknown variable HELLO
      --> <macroexpansion>:1:9
       |
     1 |  (COALTON HELLO)
       |           ^^^^^ unknown variable HELLO
```

Closes #1263

### DIFF
--- a/src/reader.lisp
+++ b/src/reader.lisp
@@ -163,9 +163,8 @@ It ensures the presence of source metadata for STREAM and then calls MAYBE-READ-
   "Compile FORMS as Coalton using the indicated MODE."
   (let* ((*readtable* (named-readtables:ensure-readtable 'coalton:coalton))
          (string (print-form (cons mode forms)))
-         (*source* (unless (source-filename)
-                     (coalton-impl/source:make-source-string string
-                                                             :name (buffer-name)))))
+         (*source* (coalton-impl/source:make-source-string string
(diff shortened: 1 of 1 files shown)


src/reader.lisp
```diff
@@ -163,9 +163,8 @@ It ensures the presence of source metadata for STREAM and then calls MAYBE-READ-
   "Compile FORMS as Coalton using the indicated MODE."
   (let* ((*readtable* (named-readtables:ensure-readtable 'coalton:coalton))
          (string (print-form (cons mode forms)))
-         (*source* (unless (source-filename)
```
jbouwman, line 166: Incorrect conditional: in the context of slime evaluation, source-filename refers to the overall form within which the macro occurs, but we want offsets relative only to the printed form.

Izaakwltn (approved):
👍

PR #1270, opened 2024-09-30
Allow zero-arg return statements

### LINKED ISSUE #1269: Parse of (return) fails
A zero-argument `return` expression should return `Unit`, but fails during compilation.

Input
```
(defpackage #:work
  (:use #:coalton))

(in-package #:work)

(coalton-toplevel
  (define (f)
    (return)))
```
Error
  read-error: 
    
      There is no applicable method for the generic function
        #<COMMON-LISP:STANDARD-GENERIC-FUNCTION COALTON-IMPL/PARSER/COLLECT::COLLECT-VARIABLES-GENERIC% (31)>
      when called with arguments
        (COMMON-LISP:NIL).
    
      (in form starting at line: 1, column: 0, position: 0)
```

jbouwman:
Closes #1269 

- guard collect-variables-generic% and add test
- improve test suite loader and reporting


Izaakwltn (approved):
lgtm

PR #1271, opened 2024-10-01
Remove exports of location accessors

jbouwman:
Remove unused exports to make consistent with usage introduced in #1223


Izaakwltn (approved):
lgtm

Issue #1204, opened 2024-08-09
`:coalton/benchmark` deprecated/impendingly deprecated


Izaakwltn:
The Coalton benchmark library depends on [trivial-benchmark](https://github.com/Shinmera/trivial-benchmark), which has recently (2 months ago) gone through a full rewrite. 

`Quicklisp` is still loading a version of `trivial-benchmark` from [October](https://quickdocs.org/trivial-benchmark) of last year, so until that updates there's no true break as long as you're using `quicklisp`. If you're manually adding dependencies locally however, this is already a breaking change.

The main changes are the removal of the macros `define-benchmark-package` and `define-benchmark`.

I'm planning on tackling this but wanted to open it to discussion.

I think this refactor would also be an opportunity to bolster our benchmark library, and make sure it is included in some facets of CI, at minimum making sure the benchmarks compile, and potentially running benchmarks on new branches, comparing to main.

Izaakwltn:
Also, similarly to `coalton/testing`, I think it might be useful to have a `coalton/benchmarking` library for external coalton programs to have friendly benchmarking.

PR #1276, opened 2024-10-02
Split benchmark packages

shirok:
As we have more benchmarks, it is more convenient to allow running not only the entire benchmarks, but the specific group of benchmarks of concern. Trivial-benchmarks treats each benchmark package as a group. This patch establishes a convention to create multiple benchmark packages.

These commits try to keep changes small, for easier transition.

### DIFF
--- a/benchmarks/package.lisp
+++ b/benchmarks/package.lisp
@@ -1,40 +1,103 @@
 ;;;; package.lisp
 ;;;;
 ;;;; Benchmarks packages and common functions
+;;;;
 
-(benchmark:define-benchmark-package #:coalton-benchmarks
-  (:export #:run-benchmarks
+;;; Each bechmark group are in separate packages, benchmark-NAME
+;;; and benchmark-NAME/native.  The latter is used to define
+;;; Coalton functions, and the former calls them with benchmarking macros.
+;;;
+;;; The coalton-benchmarks package defines functions that runs each
+;;; individual benchmark group, or all benchmarks at once.
+
+(cl:defpackage #:coalton-benchmarks
+  (:use #:cl
+        #:trivial-benchmark)
+  (:export #:run-benchmark
+           #:run-benchmarks
            #:run-benchmarks-ci))
 
-(cl:defpackage #:coalton-benchmarks/native
+(cl:in-package #:coalton-benchmarks)
+
+(defparameter *all-benchmarks* '())
+
+(defmacro define-coalton-benchmark (name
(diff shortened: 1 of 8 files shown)


benchmarks/package.lisp
```diff
@@ -1,40 +1,100 @@
...
-(cl:defpackage #:coalton-benchmarks/native
+(cl:in-package #:coalton-benchmarks)
+
+(defparameter *all-benchmarks* '())
+
+(defmacro define-coalton-benchmark (name
+                                    benchmark-clauses
+                                    &rest native-clauses)
+  (let ((benchmark-package (intern (format nil "BENCHMARK-~S" name) 'keyword))
+        (native-package (intern (format nil "BENCHMARK-~S/NATIVE" name) 'keyword)))
```
stylewarning, line 28: Can these be `COALTON-BENCHMARK/...`? It's a little wordier, but worth it.

Maybe also add a default package local nickname "NATIVE" for the native package. That way the actual benchmarks can always refer to `native:...`.

In general I don't like symbol generating macros, but for this limited use case I think it's fine.

stylewarning (changes requested):

Izaakwltn:
Just a quick note:  `trivial-benchmarks` on quicklisp is out of date, and the API has changed significantly (there are no longer benchmark packages). 

Issue #1204

shirok:
Any recommendation for alternative benchmarking package?
I have several benchmarks depending on this change, so I'd like to solidify the foundation of benchmarking now.

Izaakwltn:
Are your benchmarks using mostly Coalton or would you ideally like equal support for Common Lisp benchmarks?

shirok:
I aim at specific use patterns that are expected to be optimized by the Coalton compiler, but it includes Common Lisp counterpart as base cases, so the framework needs to measure both cases.

Izaakwltn:
I wrote an initial draft of a potential implementation in Coalton, does this look like it might be able to support what you want in your benchmarks?
#1286

shirok:
Yes. At this moment, the essential feature is to have a standard way to define benchmarks and run them per-package.  I can call plain-CL version of benchmark code from Coalton anyway. 

If #1286 will be merged soon, I'll adapt my other benchmarks to it.

shirok:
Wait, the commit has unintended changes... let me check.

shirok:
Superseded by https://github.com/coalton-lang/coalton/pull/1455

PR #1277, opened 2024-10-02
Derive specialized lisp types for transparent type

shirok:
If a Coalton type is transparent type and specialized to a concrete type, propagate the specialization when deriving lisp-type. This enables underlying Lisp compiler to do more optimization.

Example: Suppose you have
```
(repr :transparent)
(define-type (Foo :t)
  (Foo (LispArray :t)))
```
Then, if you compute Lisp type of `(Foo Single-Float)`, it returns `(simple-array single-float (*))`.

### DIFF
--- a/src/typechecker/lisp-type.lisp
+++ b/src/typechecker/lisp-type.lisp
@@ -102,7 +102,13 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
             (if (typep to 'tyvar)
                 `(cl:simple-array cl:* (cl:*))
                 `(cl:simple-array ,(lisp-type to env) (cl:*))))
-           
+
+           ;; When FROM is a transparent type, and we got a specialized type,
+           ;; we propagate parameterization to the inner type.
+           ((and (not (typep to 'tyvar))
+                 (try-recurse-transparent-type from to env)))
+
+
            ;; Otherwise we fall back.
            (t
             (lisp-type (tapp-from ty) env)))))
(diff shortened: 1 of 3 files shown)


src/typechecker/lisp-type.lisp
```diff
@@ -124,3 +130,52 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
+type)."
+             (let ((v (lookup-value-type env from :no-error t)))
+               (when (typep v 'ty-scheme)
+                 (let ((qt (qualified-ty-type (ty-scheme-type v))))
+                   ;; In transparent types, we have this in QT.
+                   ;; (Arrow <innter-type>) -> <outer-type>
+                   (when (function-type-p qt)
+                     (values (function-type-from qt)
+                             (function-type-to qt)))))))
+           )
```
stylewarning, line 166: nit: stack parens

src/typechecker/lisp-type.lisp
```diff
@@ -124,3 +130,52 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
+Lisp type `(cl:simple-array cl:fixnum)`, which allows Lisp compiler to
+optimize further.  We check the conditions and returns a specialized
+Lisp type if possible.
+
+If we can't get a specialized Lisp type, this returns nil, and let the
+fallback branch take care of it.
+"
+  (labels ((transparent-type? (from)
+             (alexandria:when-let (entry (lookup-type env from :no-error t))
+               (eq (type-entry-explicit-repr entry) :transparent)))
```
stylewarning, line 153: nit: quote data keyword

src/typechecker/lisp-type.lisp
```diff
@@ -102,7 +102,13 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
             (if (typep to 'tyvar)
                 `(cl:simple-array cl:* (cl:*))
                 `(cl:simple-array ,(lisp-type to env) (cl:*))))
-           
+
+           ;; When FROM is a transparent type, and we got a specialized type,
+           ;; we propagate parameterization to the inner type.
+           ((and (not (typep to 'tyvar))
```
stylewarning, line 108: I totally forgot that an empty clause list was valid! Nice.

src/typechecker/lisp-type.lisp
```diff
@@ -124,3 +130,52 @@ USE-FUNCTION-ENTRIES specifies whether to emit FUNCTION-ENTRY for functions, emi
...
+             (let ((v (lookup-value-type env from :no-error t)))
+               (when (typep v 'ty-scheme)
+                 (let ((qt (qualified-ty-type (ty-scheme-type v))))
+                   ;; In transparent types, we have this in QT.
+                   ;; (Arrow <innter-type>) -> <outer-type>
+                   (when (function-type-p qt)
+                     (values (function-type-from qt)
+                             (function-type-to qt)))))))
+           )
+    (when (transparent-type? from)
```
stylewarning, line 167: nit: prefer IF when the return value is meaningful and there are no side effects 

prefer WHEN when the body has side effects or the return value is not
meaningful

stylewarning (changes requested):
just some small nits.

src/typechecker/lisp-type.lisp
shirok: `(and (transpernt-type? from) ...)` may be clearer.

shirok:
Comment reflected and rebased.

src/typechecker/lisp-type.lisp
stylewarning: I still think it should be IF, if AND is to be interpreted as a boolean operator and not a secret control flow operator. :)

stylewarning (approved):

src/typechecker/lisp-type.lisp
shirok: Yeah, in CL, it is doubled as boolean operator and poor man's Maybe monad.  If you're more type oriented, you'd want to distinguish the two.  I'll keep it in mind.

PR #1274, opened 2024-10-01
Put define-class tests into a single suite file

jbouwman:
Combine the programs and error messages in tests/parser-test-files/good-files/define-class.coal and tests/parser-test-files/bad-files/define-class.* into define-class.txt

A structured input format is documented in tests/parser-test-files/README.md. It supports:

- reexecution of single cases
- selective test disabling
- optional error message checking
- test numbering

And provides a place to put additional test metadata.
The tests now compile their inputs, in addition to parsing.

This incorporates changes from #1272 but only converts define-class tests and leaves others intact.

PR #1272, opened 2024-10-01
Complete conversion per-file tests to suite format

jbouwman:
Completes the conversion of single-case test files to suites.

This earlier pull request was broken into reviewable chunks, and the first parts were merged in #1274 and #1278

PR #1278, opened 2024-10-02
Convert define, define-instance, type and expression tests

jbouwman:
Merge individual test cases into per-function files according to model established in #1274

Several test cases which parse but do not compile are disabled pending review according to comments in #1272 and #1274


tests/parser-test-files/define-type.txt
```diff
@@ -0,0 +1,301 @@
...
+error: Malformed constructor
+  --> test:4:2
+   |
+ 3 |  (define-type (T :a)
+   |               ------ in this type definition
+ 4 |    0.5)
+   |    ^^^ expected symbol
+
+================================================================================
+109 icate repr atttribute
```
Izaakwltn, line 225: `Dupl`?

tests/parser-test-files/define.txt
```diff
@@ -1,8 +1,56 @@
...
+4 Define with monomorphize attribute
+================================================================================
+
+(package coalton-unit-tests)
+
+(monomorphize)
+(define x 5)
+
+================================================================================
+5 Define (?)
```
Izaakwltn, line 41: `with inline constructor` maybe

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+    (+ 1 2)))
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define (f)
```
Izaakwltn, line 54: How about
```
(define (f)
  (let ((x (the UFix 1))
        (declare y (Unit -> UFix))
        (y (fn () (+ 1 2))))
    (+ x (y))))
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+
+================================================================================
+Parse expression
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define (f a b c)
+  (lisp (Integer -> Integer) (a b c)
```
Izaakwltn, line 68: Doesn't this just return 3, a mere Integer free of arrows?
Also this doesn't and shouldn't compile
Perhaps better:
```
(define (f a b c)
  (lisp Integer (a b c)
    (cl:+ a b c))))
```
Also I don't believe that you can return an arrow type this way from a lisp form:
```
(define (f c)
  (lisp (Integer -> Integer) (c)
    (cl:lambda (a b)
      (cl:+ a b c))))
```
returns the error
```
 warning: 
    Derived type of #:G0 is
      (VALUES FUNCTION &OPTIONAL),
    conflicting with its asserted type
      (VALUES COALTON-IMPL/RUNTIME/FUNCTION-ENTRY:FUNCTION-ENTRY &OPTIONAL).
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define (f x)
+  (match x
```
Izaakwltn, line 79: This is invalid for a number of reasons, the first being that the nested `Some` makes the next match expect a type `(Optional (Optional :a))`

This isn't as robust but might be a worthy replacement in lieu of more match tests:
```
(declare f ((Default :a) => (Optional :a) -> :a))
  (define (f x)
    (match x
      ((Some y) y)
      (_ (default))))
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+    (1 z)
+    (_ 2)))
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests)
+
+(define (g x)
```
Izaakwltn, line 92: This should be moved to bad tests with the relevant error

Izaakwltn:
```
style-warning: 
    warn: non-exhaustive match
      --> <macroexpansion>:3:4
       |
     3 |      (MATCH X)))
       |      ^^^^^^^^^ non-exhaustive match
       |      --------- missing case "_"
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define (f)
+  (let x = 1)
+  (let (Some y) = 2)
```
Izaakwltn, line 104:
```
(define (f)
    (let x = 1)
    (let ((y (from-some "invalid y" (Some 2))))
      (+ x y)))
```

This will test both let forms- I don't believe you can `let` bind a constructor like this

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define (f x y)
+  (when (== x y)
```
Izaakwltn, line 176:
```
(define (f x y)
  (when (== x y)
    (let _z = 3)
    (print "hello"))
  (unless (== x y)
    (let _z = 4)
    (print "hello")))
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+    (let z = 4)
+    (print "hello")))
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests)
+
+(define x (and 1 2 3))
```
Izaakwltn, line 189: This test case should just be removed, unless it should be:
```
(define x (and True True True))
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests)
+
+(define y (or 1 2))
```
Izaakwltn, line 197:
```
(define y (or True False))
```

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests)
+
+(define (x mx)
```
Izaakwltn, line 205:
```
(define (f ax bx)
  (do
   (a <- ax)
   (b <- bx)
    (let c = (+ a b))
    (pure c)))
```

Stolen from `intro-to-coalton.md`

tests/parser-test-files/parse-expression.txt
```diff
@@ -0,0 +1,299 @@
...
+    (x (y) z)))
+
+================================================================================
+Parse expression (:disable)
+================================================================================
+
+(package coalton-unit-tests/prelude
+  (import coalton-prelude))
+
+(define x
```
Izaakwltn, line 222:
```
(define x
  (cond 
    (True 1)
    ((> 2 3) 2)))
```

tests/parser-test-files/parse-type.txt
```diff
@@ -0,0 +1,239 @@
+================================================================================
+1 Parse type (:disable)
```
Izaakwltn, line 2: I'll file an issue after this PR is merged to split these declare cases into individual tests

Izaakwltn (approved):
Looks good! 

A few nits- the alternate code suggestions are probably worth swapping in before merging, the error message comments I'll make an issue for

tests/parser-test-files/parse-expression.txt
jbouwman: removed

PR #1289, opened 2024-10-04
Add specialization for LispArray primitive types

shirok:
This allows Lisp compiler to optimize primitive array access, by allowing Coalton compiler to inline lisparray:aref and lisparray:set! .

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -256,6 +256,7 @@
                (:file "class-tests")
                (:file "struct-tests")
                (:file "list-tests")
+               (:file "lisparray-tests")
                (:file "red-black-tests")
                (:file "seq-tests")
                (:file "pattern-matching-tests")
--- a/library/lisparray.lisp
+++ b/library/lisparray.lisp
@@ -72,7 +72,41 @@ WARNING: The consequences are undefined if an uninitialized element is read befo
     "Set the `i`th value of the `LispArray` `v` to `x`."
     (lisp Unit (v i x)
       (cl:setf (cl:aref v i) x)
-      Unit)))
+      Unit))
+
+  (lisp-toplevel ()
+    (cl:eval-when (:compile-toplevel :load-toplevel)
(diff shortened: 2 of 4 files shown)


stylewarning:
Looks like CCL errors that I can't quite grok.

shirok:
I installed CCL on Ubuntu box and tried, but it passes all tests.  Hmm...

stylewarning:
> I installed CCL on Ubuntu box and tried, but it passes all tests.  Hmm...

Did you try with the different settings that CI is testing?

shirok:
Trying to figure out how CI runs it.

stylewarning:
> Trying to figure out how CI runs it. 

```
        SAFETY: ["0", "3"]
        DISABLE_SPECIALIZATION: ["1", "0"]
        HEURISTIC_INLINING: ["1", "0"]
```

Basically, min/max safety, disable/enable specialization, disable/enable inlining.

stylewarning:
That's what the 0/1's mean in the tests.

shirok:
Update on CCL failure: 

CCL 1.13 doesn't exhibit the failure at all.  CCL 1.12.1 does, _except_ the combination of `DISABLE_SPECIALIZATION=1` and `HEURUSTIC_INLINING=0`.  I guess with that setting this particular optimization in Coalton is suppressed.

The issue is that we create `(LispArray UFix)` instance as Lisp vector of fixnum, but somehow CCL's inlined array accessor `(:INTERNAL ARRAY-ACCESS ARRAY-ACCESS)` expects `CCL::SIMPLE-UNSIGNED-DOUBLEWORD-VECTOR`.

An easy way out is to update CCL in base-image to 1.13.   If we want to avoid hitting this issue on other CL implementations, however, it may be worth seeking other options---maybe selectively turn on/off this particular optimization?

stylewarning:
@shirok I think we should update the base image if it's not too difficult.

stylewarning (approved):

PR #1290, opened 2024-10-04
Add INLINE syntax

digikar99:
The final PR towards adding inlining support in coalton. The earlier ones include:

- [Implement inlining with basic syntactic heuristics #1193](https://github.com/coalton-lang/coalton/pull/1193)
- [Add a constant-propagation pass #1212](https://github.com/coalton-lang/coalton/pull/1212)

This PR adds support for `(inline)` syntax in (i) coalton toplevel (ii) inside define-instance before methods. See the file `tests/inliner-tests.lisp` for a number of examples.

Particular aspects I can think of reviewing in this PR include:

- if someone prefers the name `inline` to `inline-p` at some places
- adding more tests / improving them

But yes, let's address any other worries or issues others spot!


macrologist (changes requested):
Mostly nits

src/parser/toplevel.lisp
```diff
@@ -1488,19 +1540,35 @@ consume all attributes")))
...
+                                                    (parse-error "Malformed inline form"
+                                                                 (note source form "expected list")))))
+                           :for method := (parse-instance-method-definition (cst:first forms) (cst:second form) source)
+                           :collect
+                           (progn
+                             (when inline-p
+                               (setf (instance-method-definition-inline-p method) inline-p))
+                             (setq forms (cst:rest forms))
+                             method))))
+
```
macrologist, line 1561:
```

:for inline-p := (and (should-inline-p forms) 
                     (prog1 (parse-inline (cst:first forms) source)
                         (setf forms (cst:rest forms))))
:for method := (parse-instance-method-definition (cst:first forms) (cst:second form) source)
:when inline-p :do (setf (instance-method-definition-inline-p method) inline-p)
:collect :method 
:do (setf forms (cst:rest forms))

```

nit: this seems easier to follow to me. managing the state of `forms` is a little confusing to follow, I think something like the above makes it clearer.

src/codegen/optimizer.lisp
```diff
@@ -421,8 +424,10 @@ requires direct constructor calls."
              (alexandria:when-let*
                  ((name (node-rator-name node))
                   (code (tc:lookup-code env name :no-error t))
+                  (fun-env-entry (tc:lookup-function env name :no-error t))
+                  (inline-p (when fun-env-entry (tc:function-env-entry-inline-p fun-env-entry)))
                   (_    (and (node-abstraction-p code)
-                             (funcall heuristic code)
+                             (or inline-p (funcall heuristic code))
```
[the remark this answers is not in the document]
digikar99: Would it make sense to rewrite this condition as the following?

```lisp
                  (fun-env-entry (tc:lookup-function env name :no-error t))
                  (_    (and (node-abstraction-p code)
                             (or (tc:function-env-entry-inline-p fun-env-entry)
                                 (funcall heuristic code))
                             (<= (length call-stack)
                                 max-depth)
```

src/parser/toplevel.lisp
```diff
@@ -779,12 +790,25 @@ If the outermost form matches (eval-when (compile-toplevel) ..), evaluate the en
...
+      attributes)
+
      nil)
 
     ((coalton:define)
      (let ((define (parse-define form source))
            monomorphize
-           monomorphize-form)
+           monomorphize-form
+
```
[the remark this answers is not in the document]
digikar99: Is there a preference for the newlines under the `(coalton:declare)` clause below? Should I remove them?

src/parser/toplevel.lisp
digikar99: > nit: this seems easier to follow to me. managing the state of forms is a little confusing to follow, I think something like the above makes it clearer.

Unless I'm missing something, the suggested code still manually manipulates the state of `forms`. But I understand, this part of the code is a bit complicated to follow.

macrologist: that's correct. It's just a nit. Changes are not necessary

src/codegen/optimizer.lisp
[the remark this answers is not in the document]
amorphedstar: Yeah, I think that is a good way to rewrite it so that it will work with the `when-let*`.
The only thing I'm unsure of is if there ever can be a case when there is an entry in the `code-environment` but not in the `function-environment`. (I can't tell easily just from looking at the code, since the code entries and function entries are added at very different places in the code.) If so, then in order to allow heuristic inlining of functions with code env entries but not function env entries, it seems like a separate let statement might be necessary, like
```lisp
                  (_    (and (node-abstraction-p code)
                             (or (alexandria:when-let (fun-env-entry (tc:lookup-function env name :no-error t))
                                   (tc:function-env-entry-inline-p fun-env-entry))
                                 (funcall heuristic code))
                             (<= (length call-stack)
                                 max-depth)
```
or if not, then your suggestion should work exactly as written, or even without the `:no-error t`.

shirok:
This is a general comment, probably for future work.  As the optimizer gets more sophisticated, there'll be more context like `monomorphize-table` and `inline-p-table` to be passed.  At some point we might want to consolidate them into optimizer context, maybe?

src/codegen/optimizer.lisp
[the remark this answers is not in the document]
digikar99: If I remove `:no-error t` in my suggestion, I obtain the error "Internal coalton bug: Unknown function TREE:INCREASING-ORDER" while compiling `library/ord-map.lisp`. So, there indeed seem to be cases with code env entries but not function env entries.

However, I note that your suggestion works even without the `:no-error t`. My hypothesis is there are fun-env-entries only when code is an abstraction node.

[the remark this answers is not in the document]
digikar99: Committed in https://github.com/coalton-lang/coalton/pull/1290/commits/849af1f4816a31fa85e78eb7114ded8a61669d24

tests/inliner-tests.lisp
```diff
@@ -6,3 +6,106 @@
...
+   "(inline)
+      (define (factorial-1 n)
+        (if (== n 0)
+            1
+            (* n (factorial-2 (- n 1)))))
+      (inline)
+      (define (factorial-2 n)
+        (if (== n 0)
+            1
+            (* n (factorial-1 (- n 1)))))"))
```
amorphedstar, line 111: I don't know what style guidelines we have for multi-line strings, but removing some indents here might make it nicer to read:

```suggestion
   "(inline)
    (define (factorial-1 n)
      (if (== n 0)
          1
          (* n (factorial-2 (- n 1)))))
    (inline)
    (define (factorial-2 n)
      (if (== n 0)
          1
          (* n (factorial-1 (- n 1)))))"))
```

tests/inliner-tests.lisp
```diff
@@ -6,3 +6,106 @@
...
+
+(deftest method-inline-error ()
+  (signals coalton-impl/parser:parse-error
+    (check-coalton-types
+     "(define-class (C :a)
+        (m (:a -> :a -> :a)))
+      (define-instance (m Double-Float)
+        (inline m)
+        (define (m x y)
+          (lisp Double-Float (x y) (cl:+ x y))))")))
```
amorphedstar, line 82: You could add another test for duplicate inline attributes to match the tests for `monomorphize` ( https://github.com/search?q=repo%3Acoalton-lang%2Fcoalton+monomorphize+path%3A%2F%5Etests%5C%2Ftest-files%5C%2F%2F&type=code ) but other than that I think this looks pretty good and complete

digikar99: Should the tests be only for the duplicate inline attributes, or for others too (define, parse-toplevel, parse-attribute, define-type)?

amorphedstar: I would think that adding more tests is never a bad thing, especially if you think they would be useful

digikar99: Added in https://github.com/coalton-lang/coalton/pull/1290/commits/7d98dcaabce6a918d28ed2af8394629af059a20e and https://github.com/coalton-lang/coalton/pull/1290/commits/9edb4b3b1167fc461ddfa23126b8e2e5544bb0dc

tests/test-files/define-type.txt
```diff
@@ -265,6 +265,25 @@ error: Invalid target for monomorphize attribute
  4 |  (define-type T)
    |  --------------- when parsing define-type
 
+================================================================================
+111 Invalid target for inline attribute
```
amorphedstar, line 269: I don't know how these special test files work, but maybe someone else would know: does the number before the test case title matter at all? This new test starts with the same number (111) as the existing test below, but I do agree that it makes sense to put it here by the monomorphize parser failure test. Maybe we just increment the counter for each of the later tests accordingly?

digikar99: There are no numbers in `parse-declare.txt`. So may be the numbers are optional?

src/parser/toplevel.lisp
YarinHeffes: I think something like this produces errors that are more aligned with the rest of the parser:
```lisp
(let ((methods (loop :with inline := nil
                           :for forms := (cst:nthrest (if docstring 3 2) form) :then (cst:rest forms)
                           :while (cst:consp forms)
                           :for method-or-attribute := (cst:first forms)
                           :if (eq 'coalton:inline (cst:raw (cst:first method-or-attribute)))
                             :if inline :do (parse-error "Duplicate inline attribute"
                                                         (note source method-or-attribute
                                                               "inline attribute here")
                                                         (note source (source:location-span (attribute-inline-location inline))
                                                               "previous attribute here"))
                               :else :do (setf inline (parse-inline method-or-attribute source))
                           :else
                             :collect (parse-instance-method-definition method-or-attribute (cst:second form) (prog1 inline (setf inline nil)) source))))
        (make-toplevel-define-instance
         :context context
         :pred (parse-predicate unparsed-predicate
                                (source:make-location source
                                                      (util:cst-source-range unparsed-predicate)))
         :docstring docstring
         :methods methods
         :location (form-location source form)
         :head-location (form-location source (cst:second form))
         :compiler-generated nil))
```
where `parse-instance-method-definition` takes an extra argument of type `(or null attribute-inline)`. I don't have a strong opinion about that particular choice; it can be handled as you did with `setf`.

The order of the parser checks (duplicate inline and malformed inline) can be swapped to an order that makes more sense in my opinion, but the code is slightly more difficult to follow:

```lisp
:if (eq 'coalton:inline (cst:raw (cst:first method-or-attribute)))
  :do (setf inline (prog1 (parse-inline method-or-attribute source)
                     (when inline
                       (parse-error "Duplicate inline attribute"
                                    (note source method-or-attribute
                                          "inline attribute here")
                                    (note source (source:location-span (attribute-inline-location inline))
                                          "previous attribute here")))))
```

tests/test-files/define-type.txt
YarinHeffes: I would just go ahead and rename the last two tests to 112 and 113.

tests/test-files/define.txt
```diff
@@ -38,7 +38,16 @@
...
 
 ================================================================================
-5 Define with inline constructor
+5 Define with inline attribute
+================================================================================
+
+(package coalton-unit-tests)
+
+(inline)
+(define x 5)
```
YarinHeffes, line 47: I would add a test with inline declare define and another with declare inline define. I would also copy-paste the duplicate monomorphize error test below as a check on duplicate inline.

YarinHeffes: Also, since we are adding `(inline)` on method definitions which is usage not tested with `monomorphize`, we should definitely have tests in `define-instance.txt`

YarinHeffes (changes requested):
Aside from my main change request in the parser for inlining methods, and for tests in `tests/test-files/define-instance.txt`, this PR looks ready to me. A follow-up PR should include documentation in `docs/intro-to-coalton.md`, and eventually, users should have control over variables like `max-unroll` and `max-depth` when using the attribute. These two points should not hold up this PR.

src/parser/toplevel.lisp
digikar99: I'm liking your version of the `methods` computation.

For this part, I find the following easier to read:

```lisp
:if (eq 'coalton:inline (cst:raw (cst:first method-or-attribute)))
  :do (if inline
          (parse-error "Duplicate inline attribute"
                       (note source method-or-attribute
                             "inline attribute here")
                       (note source (source:location-span (attribute-inline-location inline))
                             "previous attribute here"))
          (setf inline (parse-inline method-or-attribute source)))
```

YarinHeffes: That looks good to me, provided we note that the duplicate inline error will be checked before the malformed inline error. I think it's ultimately arbitrary which error comes first, so you can make the change! I think this PR can be merged soon, and I want it to be.

digikar99: Sorry, it took a while, other things in life! 

Pushed in https://github.com/coalton-lang/coalton/pull/1290/commits/f29eb7e6debe2d0b0bcd7ee57329fac947a1702f - let me know if you want to change anything. I went ahead with `setf`-ing the inline attribute as that keeps the parse-* functions largely consistent with each other. But let me know if you (or anyone!) would prefer to pass the inline attribute to the `parse-instance-method-definition` instead!

tests/test-files/define.txt
digikar99: > I would just go ahead and rename the last two tests to 112 and 113.

Done in https://github.com/coalton-lang/coalton/pull/1290/commits/20af8c03ca0229fc4527080d322165a88841b11f

> I would add a test with inline declare define and another with declare inline define. I would also copy-paste the duplicate monomorphize error test below as a check on duplicate inline.
> .
> 
> Also, since we are adding `(inline)` on method definitions which is usage not tested with `monomorphize`, we should definitely have tests in `define-instance.txt`

Done in https://github.com/coalton-lang/coalton/pull/1290/commits/20af8c03ca0229fc4527080d322165a88841b11f but let me know if you'd like to add more or correct something.

I'm unable to get the tests to pass locally - I get the following unrelated error. I'm assuming it's something on my own end. But if other people can reproduce the error in this PR, I'll look into it to debug!

```lisp
Rerun single test case with:
(coalton-tests:run-test "tests/test-files/define-instance.txt" 100)

expected error (A) and generated error (B):
A: error: Malformed method definition
  --> test:4:2
   |
 3 |  (define-instance (C :a)
   |                   ------ when parsing instance
 4 |    5)
   |    ^ expected list
B: Encountered #<CONCRETE-SYNTAX-TREE:ATOM-CST raw: 5 {100A9DC853}> where a
CONS-CST is required.
```

src/parser/toplevel.lisp
YarinHeffes: You're going to need to do `(and (cst:consp method-or-attribute) ...)` before `(cst:raw (cst:first ...))` in the `:if`. That's causing your error in your comment below.

src/parser/toplevel.lisp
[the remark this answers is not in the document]
YarinHeffes: I still think we should use `inline` because it's not a boolean and it is more consistent with `monomorphize`. It may be worth it to implement the attribute `(optimize :monomorphize :inline)` instead and to store them as an `optimize-attribute` struct, in the same spirit as @shirok's comment above. If that is something we want to do, then it would be better to implement it now, rather than merging this syntax only to change it soon after. Is there a reason not to do this?

src/parser/toplevel.lisp
digikar99: Thanks! So, my dependencies are still in a good condition and not out of date to make the test runs unreliable. Fixed in https://github.com/coalton-lang/coalton/commit/a81dcdb5fc8fef8309637fce2888fd7e980e34c7

src/parser/toplevel.lisp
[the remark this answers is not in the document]
digikar99: I'm okay with the naming change from `inline-p` to `inline`. I assume you are suggesting the `inline` name to be used everywhere instead of `inline-p`? 

About merging `monomorphize` and `inline`: currently, both of them occur in `toplevel-declare` and `toplevel-define`. But only `inline` occurs in `instance-method-definition`. So, I am not sure if they should be merged uniformly? Or perhaps we might also need to maintain a record of "valid optimizations" somewhere (?)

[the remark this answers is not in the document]
YarinHeffes: Not in _every_ case. I think we should use inline-p whenever we are referring to a Boolean and inline in every other case. As for optimize, I recommend you to look at #1297. You’ll need to spend some time reading that to understand how to resolve merge conflicts. Rather than making a consume-inline the same way there is a consume-monomorphize in that PR, I encourage you to handle them together in a function called consume-optimize-attribute and take keyword arguments for monomorphize and inline. It’ll be more concise. There is also an instance there where one of the repr options is disallowed - we would handle the disallowing of monomorphize on methods in the same way. 

Also, you’ll need to look at #1350 to resolve the remaining merge conflicts.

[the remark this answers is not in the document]
YarinHeffes: @digikar99 just to follow up and clarify, the idea to consolidate monomorphize and inline into optimize is just a thought that I have and it would take some work. Maybe we can get a second thought from @amorphedstar or @stylewarning, because otherwise I think the PR is in good shape for a final review after resolving merge conflicts.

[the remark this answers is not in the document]
digikar99: I just attempted a rebase locally and am resolving the conflicts. The monomorphize and inline consolidation seems easier than this resolution (: - but the conflict resolution also seems to be a prerequisite to the monomorphize and inline attribute consolidation. Hoping we get this PR merged in another week before more conflicts take place!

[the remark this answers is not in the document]
digikar99: https://github.com/coalton-lang/coalton/pull/1290/commits/5d2fba98af698696c08af7df78d14311d9dcd2c1 might need a better rewrite. Other than that, the commits are now rebased to 2ad7d0efb7ab8f322a5eee87eef7a3c53f3892f3

src/typechecker/environment.lisp
```diff
@@ -684,11 +686,12 @@
...
-  (constraints         (util:required 'constraints)         :type ty-predicate-list :read-only t)
-  (predicate           (util:required 'predicate)           :type ty-predicate      :read-only t)
-  (codegen-sym         (util:required 'codegen-sym)         :type symbol            :read-only t)
-  (method-codegen-syms (util:required 'method-codegen-syms) :type util:symbol-list  :read-only t)
-  (docstring           (util:required 'docstring)           :type (or null string)  :read-only t))
+  (constraints             (util:required 'constraints)         :type ty-predicate-list :read-only t)
+  (predicate               (util:required 'predicate)           :type ty-predicate      :read-only t)
+  (codegen-sym             (util:required 'codegen-sym)         :type symbol            :read-only t)
+  (method-codegen-syms     (util:required 'method-codegen-syms) :type util:symbol-list  :read-only t)
+  (method-codegen-inline-p (util:required 'method-codegen-inline-p) :type list    :read-only t)
```
YarinHeffes, line 693: nit: align spacing before `:type` and `:read-only`.

src/codegen/monomorphize.lisp
```diff
@@ -414,11 +414,12 @@ propagate dictionaries that have been moved by the hoister."
       (action (:after node-application) #'apply-candidate)
       (action (:after node-direct-application) #'apply-candidate)))))
 
-(defun monomorphize (name manager package resolve-table optimize-node env)
+(defun monomorphize (name manager package resolve-table inline-p-table optimize-node env)
```
YarinHeffes, line 417: I'm not as familiar with the codegen phase yet, but seeing these changes in the `monomorphize` function implies to me that it would make sense to, at some point, consolidate `inline` and `monomorphize` into an umbrella `optimize`. I personally don't think that is necessary before merging this PR, but after the PR is merged, maybe we should file an issue.

YarinHeffes (changes requested):
Looks good to me overall, just a few small changes 👍

src/parser/toplevel.lisp
```diff
@@ -852,9 +868,17 @@ If the attribute is not unique, or a repr attribute is present, signal a parse e
                                 (source:note attribute "monomorphize attribute here")
                                 (source:secondary-note monomorphize "previous attribute here")
                                 (source:secondary-note toplevel-form message)))
-                 (setf monomorphize attribute))))
-    (setf (fill-pointer attributes) 0)
```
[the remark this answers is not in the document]
digikar99: For putting `(setf (fill-pointer attributes) 0)` inside `consume-optimize-attribute`, are you suggesting 

1. We set the fill-pointer only after making sure we have consumed all the attributes? OR,
2. We *necessarily* return *both* monomorphize and inline if both are present?

I'm assuming there can be more than one toplevel attributes active at once. Given the way it currently is, I'm not even happy for `attributes` to be a vector. It should probably be an alist or hash-table. But that's probably for another PR.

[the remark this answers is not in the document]
YarinHeffes: I was thinking something like the following, where we throw a parse error when `inline` or `monomorphize` is found though not requested. What do you think? This is nit, and I don't see a problem with the way you wrote it, but I had this in mind to resolve the attributes with a single call to `consume-optimize-attribute(s?)`.

```lisp
(defun consume-optimize-attribute (attributes toplevel-form message &key (monomorphize-p nil) (inline-p nil))
  "Return the unique monomorphize attribute in ATTRIBUTES, or NIL.
If the attribute is not unique, or a repr attribute is present, signal a parse error."
  (declare (type (member :monomorphize :inline) attribute))
  (let (monomorphize inline)
    (loop :for attribute :across attributes
          :do (etypecase attribute
                (attribute-repr
                 (parse-error "Invalid target for repr attribute"
                              (source:note attribute "repr must be attached to a define-type")
                              (source:secondary-note toplevel-form message)))
                (attribute-monomorphize
                 (unless monomorphize-p
                   (parse-error "Invalid target for monomorphize attribute"
                                (source:note attribute "monomorphize must be attached to a define or declare form")
                                (source:secondary-note toplevel-form message)))
                 (when monomorphize
                   (parse-error "Duplicate monomorphize attribute"
                                (source:note attribute "monomorphize attribute here")
                                (source:secondary-note monomorphize "previous attribute here")
                                (source:secondary-note toplevel-form message)))
                 (setf monomorphize attribute))
                (attribute-inline
                 (unless inline-p
                   (parse-error "Invalid target for inline attribute"
                                (source:note attribute "inline must be attached to a define or declare form")
                                (source:secondary-note toplevel-form message)))
                 (when inline
                   (parse-error "Duplicate inline attribute"
                                (source:note attribute "inline attribute here")
                                (source:secondary-note inline "previous attribute here")
                                (source:secondary-note toplevel-form message)))
                 (setf inline attribute))))
    (setf (fill-pointer attributes) 0)
    (values monomorphize inline)))
```

Then you can change this:
```lisp
    ((coalton:define)
     (let* ((define (parse-define form source))
            (monomorphize (consume-optimize-attribute :monomorphize attributes define "when parsing define"))
            (inline (consume-optimize-attribute :inline attributes define "when parsing define")))
       (setf (toplevel-define-monomorphize define) monomorphize)
       (setf (toplevel-define-inline define) inline)
       (setf (fill-pointer attributes) 0)
       (push define (program-defines program))
       t))
```
to this:
```lisp
    ((coalton:define)
     (let ((define (parse-define form source)))
       (multiple-value-bind (monomorphize inline)
           (consume-optimize-attribute attributes define "when parsing define" :monomorphize-p t :inline-p t)
         (setf (toplevel-define-monomorphize define) monomorphize)
         (setf (toplevel-define-inline define) inline))
       (push define (program-defines program))
       t))
```

src/parser/toplevel.lisp
```diff
@@ -1568,22 +1604,36 @@ consume all attributes")))
...
+                                                  (note source method-or-attribute "inline attribute here")
+                                                  (source:secondary-note (attribute-inline-location inline) "previous attribute here"))
+                                     (setf inline (parse-inline method-or-attribute source)))
+                           :else
+                             :collect (let ((method (parse-instance-method-definition method-or-attribute (cst:second form) source)))
+                                        (setf (instance-method-definition-inline method) inline
+                                              inline nil)
+                                        method)
+                           :finally (when inline
+                                      (parse-error "Unnecessary inline attribute"
```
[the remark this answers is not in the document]
digikar99: Done: https://github.com/coalton-lang/coalton/pull/1290/commits/0e27470c429c734f367e2672a44c479fcb84e240

src/entry.lisp
```diff
@@ -84,9 +86,26 @@
...
 
+                  (loop :for define :in (parser:program-defines program)
+                        :when (parser:toplevel-define-inline define)
+                          :do (setf (gethash (parser:node-variable-name (parser:toplevel-define-name define))
+                                             inline-p-table)
+                                    t))
+
+                  (loop :for declare :in (parser:program-declares program)
+                        :when (parser:toplevel-declare-inline declare)
+                          :do (setf (gethash (parser:identifier-src-name (parser:toplevel-declare-name declare))
```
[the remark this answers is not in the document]
digikar99: Done: https://github.com/coalton-lang/coalton/pull/1290/commits/2727de03bc3c79ff0a461cb486fe0e007daa2855

src/typechecker/environment.lisp
digikar99: Done: https://github.com/coalton-lang/coalton/pull/1290/commits/7229e81c2541bf1ad34ebdc91912d7043343ee53

src/parser/toplevel.lisp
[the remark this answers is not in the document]
digikar99: Got it, thanks for elaborating! 

Though, this does look more complicated that the current code. I'm also unable to figure out the kind of toplevel forms that will trigger `"Invalid target for monomorphize/inline attribute"` since the only way for this error to occur is for someone to call `(consume-optimize-attribute ...)` with `:monomorphize-p nil` or `:inline-p nil`; but that never happens (?)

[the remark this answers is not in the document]
YarinHeffes: That's true! You could set the default values to `t` to make it less verbose, but ultimately this whole chain comes down to a style preference. I don't think you need to implement this change. It can be revisited in the future if it happens to become an issue.

YarinHeffes (approved):

src/codegen/monomorphize.lisp
digikar99: On a second thought, I think I'm leaning towards a separate PR for the consolidation of `inline` and `monomorphize` into the single attribute and table. It's not so much the amount of work, but rather the kind of data structures that should be used and the kind of extensibility one wants from them. Plus, may be putting that here would obscure the commit history (if it's any meaningful currently!) of this PR.

stylewarning (approved):

stylewarning:
I didn't squash commits though I normally would, just to preserve some history. At long last! Thanks!

PR #1291, opened 2024-10-07
Add a couple of variations of `traverse`.

shirok:
These are particularly useful in the optimizing path where traverser wants to have a broader view than a single subtree.

`traverse-with-parent-thunk`: In an optimizing pass, you sometimes neeed to know the parent of the node you're looking at.  With this variation of `traverse`, the action functions receives a thunk after the node argument, which returns the parent node when called inside the function.  An example is `print-node-traversal-order`.

`traverse-with-hook`: This is more general mechanism to intercept calling action procedures.  `traverse-with-parent-hook` is implemented on top of this. Traverse hook is a function that is called whenever `traverse` recurse into subnodes.  It allows to perform operations during traversal, anostic to how the node walker handle recursion for each node class.

### DIFF
--- a/src/codegen/traverse.lisp
+++ b/src/codegen/traverse.lisp
@@ -9,9 +9,12 @@
    #:action
    #:count-applications
    #:count-nodes
+   #:identity-traverse-hook
    #:make-traverse-let-action-skipping-cons-bindings
    #:*traverse*
    #:traverse
+   #:traverse-with-hook
+   #:traverse-with-parent-thunk
    #:traverse-with-binding-list))
 
 (in-package #:coalton-impl/codegen/traverse)
@@ -233,7 +236,11 @@ any node-specific actions that will run after another action which may
 return a node of a different type. For example, if there is an action
 on `:traverse 'node-variable` to substitute variables with arbitrary
 other nodes, then it would be inappropriate to also define an action
-`:after 'node-variable`."
+`:after 'node-variable`.
+
+See `traverse-with-hook` if you need to intercept when actions are
+called recursively.
+"
   (declare (type node        initial-node)
            (type action-list custom-actions)
            (type list        initial-args)
@@ -250,8 +257,54 @@ other nodes, then it would be inappropriate to also define an action
               (fire-action ':traverse (class-name (class-of current-node)) actions args)
               (fire-action ':after    (class-name (class-of current-node)) actions args)
               (fire-action ':after    'node                                actions args))))
-      (let ((*traverse* #'current-traverse))
+      (let ((*traverse* (lambda (node &rest args)
+                          (funcall *traverse-hook*
+                                   (lambda ()
+                                     (apply #'current-traverse node args))
+                                   node))))
         (apply *traverse* initial-node initial-args)))))
+
+(defun identity-traverse-hook (thunk node)
+  "Default traversal hook function, which doesn't do any extra operation.
+This is called when the caller does _not_ use `traverse-with-hook`."
+  (declare (ignore node)) (funcall thunk))
+
+(defvar *traverse-hook* #'identity-traverse-hook
+  "A private dynamic variable used by `traverse-with-hook`.  This is rebound
+dynamically during running `traverse-with-hook`.")
+
+(defun traverse-with-hook (hook node action-list &rest args)
+  "Like `traverse`, but uses `hook` as the traversal hook function.
+The hook function is called whenever `traverse` recurse into the sub
+nodes.  It takes two arguments: `thunk` to recurse traversal, and `node`
+that is about to be recursed.
+
+The `traverse` function is equivalent to this with passing
+`identity-traverse-hook` as the hook.
+
+One of the use cases is to record parent node of the node being traversed.
+See `print-node-parent` in traverse.lisp for the example."
+  (let ((*traverse-hook* hook))
+    (apply #'traverse node action-list args)))
+
+;; Used in traverse-with-parent-node
+(defvar *traverse-parent-node-parent* nil)
+(defvar *traverse-parent-node-current* nil)
+
+(defun traverse-with-parent-thunk (node action-list &rest args)
+  "Like 'traverse', but actions are called with a thunk in the first argument,
+which returns the node's parent node.
+This captures one of the typical use case of `traverse-with-hook`."
+  (labels ((track-parent-node-hook (thunk node)
+             (let ((*traverse-parent-node-parent* *traverse-parent-node-current*)
+                   (*traverse-parent-node-current* node))
+               (funcall thunk))))
+    (apply #'traverse-with-hook
+           #'track-parent-node-hook
+           node
+           action-list
+           (lambda () *traverse-parent-node-parent*)
+           args)))
 
 ;;;
 ;;; Traversals with bound variables
@@ -388,6 +441,27 @@ without any slot information."
       (action (:after node node)
         (decf counter)
         (format t "POST: ~v@{|   ~}~A~%" counter (class-name (class-of node)))
+        (values))))))
+
+(defun print-node-parent (node)
+  "Print visiting node and its parent, using `traverse-with-parent-thunk`."
+  (declare (type node node)
+           (values node &optional))
+  (let ((counter 0))
+    (traverse-with-parent-thunk
+     node
+     (list
+      (action (:before node node parent-thunk)
+        (format t "PRE:  ~v@{|   ~}~A ~A~%" counter
+                (class-name (class-of node))
+                (class-name (class-of (funcall parent-thunk))))
+        (incf counter)
+        (values))
+      (action (:after node node parent-thunk)
+        (decf counter)
+        (format t "POST: ~v@{|   ~}~A ~A~%" counter
+                (class-name (class-of node))
+                (class-name (class-of (funcall parent-thunk))))
         (values))))))
 
 (defun make-traverse-let-action-skipping-cons-bindings ()


src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
               (fire-action ':after    'node                                actions args))))
-      (let ((*traverse* #'current-traverse))
+      (let ((*traverse* (lambda (node &rest args)
+                          (funcall *traverse-hook*
+                                   (lambda ()
+                                     (apply #'current-traverse node args))
+                                   node))))
         (apply *traverse* initial-node initial-args)))))
 
+(defun identity-hook (thunk node)
```
stylewarning, line 262: nit: I suggest renaming to `identity-traverse-hook`

src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
+  "Default traversal hook function, which doesn't do any extra operation."
+  (declare (ignore node)) (funcall thunk))
+
+(defvar *traverse-hook* #'identity-hook)
+
+(defun traverse-with-hook (hook node action-list &rest args)
+  "Like `traverse`, but uses `hook` as the traversal hook function.
+The hook function is called whenever `traverse` recurse into the sub
+nodes.  It takes two arguments: `thunk` to recurse traversal, and `node`
+that is about to be recursed.
```
stylewarning, line 272: Add a paragraph stating that `identity-traverse-hook` is an identity function, or something like that.

src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
+                                   node))))
         (apply *traverse* initial-node initial-args)))))
 
+(defun identity-hook (thunk node)
+  "Default traversal hook function, which doesn't do any extra operation."
+  (declare (ignore node)) (funcall thunk))
+
+(defvar *traverse-hook* #'identity-hook)
+
+(defun traverse-with-hook (hook node action-list &rest args)
```
stylewarning, line 268: Add documentation to the `traverse` function saying something like: "See `traverse-with-hook` for a more customizable variant." or something like that.

src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
+                                   (lambda ()
+                                     (apply #'current-traverse node args))
+                                   node))))
         (apply *traverse* initial-node initial-args)))))
 
+(defun identity-hook (thunk node)
+  "Default traversal hook function, which doesn't do any extra operation."
+  (declare (ignore node)) (funcall thunk))
+
+(defvar *traverse-hook* #'identity-hook)
```
stylewarning, line 266: Add documentation, but be explicit that `traverse-with-hook` should be used.

src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
+The hook function is called whenever `traverse` recurse into the sub
+nodes.  It takes two arguments: `thunk` to recurse traversal, and `node`
+that is about to be recursed.
+
+One of the use cases is to record parent node of the node being traversed.
+See `print-node-parent` in traverse.lisp for the example."
+  (let ((*traverse-hook* hook))
+    (apply #'traverse node action-list args)))
+
+;; Used in traverse-with-parent-ndoe
```
stylewarning, line 279: nit: typo

src/codegen/traverse.lisp
```diff
@@ -250,9 +252,47 @@ other nodes, then it would be inappropriate to also define an action
...
+
+(defun traverse-with-parent-thunk (node action-list &rest args)
+  "Like 'traverse', but actions are called with a thunk in the first argument,
+which returns the node's parent node.
+This captures one of the typical use case of `traverse-with-hook`."
+  (labels ((track-parent-node-hook (thunk node)
+             (let ((*traverse-parent-node-parent* *traverse-parent-node-current*)
+                   (*traverse-parent-node-current* node))
+               (funcall thunk))))
+    (let ((*traverse-hook* #'track-parent-node-hook))
```
stylewarning, line 291: traverse-with-hook can be used here, no?

stylewarning (changes requested):
Mostly small documentation-related comments.

src/codegen/traverse.lisp
shirok: `identity-traverse=hook` isn't exported, and it is only called when a bare `traverse` is used.  I can explain that in the documentatino of  `identity-traverse-hook`.

src/codegen/traverse.lisp
shirok: Right.  That was the original intention but somehow I forgot, I guess.

src/codegen/traverse.lisp
stylewarning: You might consider exporting it as a nice "apropos"-able example.

shirok: That's reasonable.  I'll do so.

src/codegen/traverse.lisp
```diff
@@ -250,9 +256,52 @@ other nodes, then it would be inappropriate to also define an action
...
         (apply *traverse* initial-node initial-args)))))
 
+(defun identity-traverse-hook (thunk node)
+  "Default traversal hook function, which doesn't do any extra operation.
+This is called when the caller does _not_ use `traverse-with-hook`."
+  (declare (ignore node)) (funcall thunk))
+
+(defvar *traverse-hook* #'identity-traverse-hook
+  "A private dynamic variable used by `traverse-with-hook`.  This is rebound
+dyncamically during running `traverse-with-hook`.")
```
stylewarning, line 273: nit: typo

stylewarning (approved):
there is a small typo, but otherwise lgtm

src/codegen/traverse.lisp
```diff
@@ -390,6 +443,27 @@ without any slot information."
...
+                (class-name (class-of node))
+                (class-name (class-of (funcall parent-thunk))))
+        (incf counter)
+        (values))
+      (action (:after node node parent-thunk)
+        (decf counter)
+        (format t "POST: ~v@{|   ~}~A ~A~%" counter
+                (class-name (class-of node))
+                (class-name (class-of (funcall parent-thunk))))
+        (values))))))
```
amorphedstar, line 465: As far as I can tell, identical behavior can be achieved without traversing with the extra hook:
(not actually a code suggestion, just formatting it as one for easy comparison)
```suggestion
(defun print-node-parent (node)
  "Print visiting node and its parent, using `traverse-with-parent-thunk`."
  (declare (type node node)
           (values node &optional))
  (let ((traversal-stack nil))
    (traverse
     node
     (list
      (action (:before node node)
        (format t "PRE:  ~v@{|   ~}~A ~A~%" (length traversal-stack)
                (class-name (class-of node))
                (class-name (class-of (car traversal-stack))))
        (push node traversal-stack)
        (values))
      (action (:after node node)
        (pop traversal-stack)
        (format t "POST: ~v@{|   ~}~A ~A~%" (length traversal-stack)
                (class-name (class-of node))
                (class-name (class-of (car traversal-stack))))
        (values))))))
```
Can you think of any other examples, where some functionality can only be achieved conveniently with `traverse-with-hook`, but not with the original `traverse`?

amorphedstar (commented):
This is a pretty cool idea and it's nice that you were able to do it quite concisely!
There were a few minor typos, but I am mostly curious to see if there is a better example where the extra hook is actually needed.

shirok:
@amorphedstar The original motivation is to get parent node during traversal.  We didn't want a big surgery to `traverse`, though, for we might need other variations in future.  So the idea of hook came.  It is a bit hacky, but extensible.

I don't see other applications immediately, but what that might be needed for optimizations can be:

- Make the path (from the root node to the current node) available
- Gather information of subtrees of a particular type of node.  E.g. specific type of node sets a table in a dynamic variable, and the descendant nodes add info to it.

src/codegen/traverse.lisp
```diff
@@ -250,9 +257,55 @@ other nodes, then it would be inappropriate to also define an action
...
+dynamically during running `traverse-with-hook`.")
+
+(defun traverse-with-hook (hook node action-list &rest args)
+  "Like `traverse`, but uses `hook` as the traversal hook function.
+The hook function is called whenever `traverse` recurse into the sub
+nodes.  It takes two arguments: `thunk` to recurse traversal, and `node`
+that is about to be recursed.
+
+The `traverse` function is equivalent to this with passing
+`identity-traverse-hook` as the hook.
```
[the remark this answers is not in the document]
shirok: Right.  I didn't think the case that `traverse` is called during other `traverse`-family functions, but it is certainly possible.
Is there already a case that a function called from an action recursively calls `traverse`, thus can be affected with this commit? 
Otherwise, I'd just fix this docstring.

src/codegen/traverse.lisp
shirok: You're completely right.   Merely obtaining parent can be achieved by this.
What I really need is a convenient API to access to the parent node, so we may avoid employing `*traverse-hook*` hack.  Let me think on it.

src/codegen/traverse.lisp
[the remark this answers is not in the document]
amorphedstar: Yes, the inliner definitely needs other sub-traversals of the code to be inlined, in order to rename type variables to fresh ones, as well as to perform AST variable substitution.

shirok:
I looked at  https://github.com/coalton-lang/coalton/pull/1291#discussion_r1791160000 and actually what I wanted could be achievable without `*traverse-hook*`.  Let's see if I can make a convenient API around your idea.

shirok:
@amorphedstar Need some input.  I like your approach of keeping `traversal-stack` local, but what I want to achieve is `traverse`-like API that takes an action list, but each action gets the parent node info in the argument.

I'm trying to write `traverse-with-path`, which passes a thunk to return the "path" (`traversal-stack` in your code) to each actions, so that I can use it as follows:

```
(defun print-node-parent (node)
  "Print visiting node and its parent, using `traverse-with-path`."
  (declare (type node node)
           (values node &optional))
  (traverse-with-path
   node
   (list
    (action (:before node node path-thunk)
      (format t "PRE:  ~v@{|   ~}~A ~A~%" (length (funcall path-thunk))
              (class-name (class-of node))
              (class-name (class-of (car (funcall path-thunk)))))
      (values))
    (action (:after node node path-thunk)
      (format t "POST: ~v@{|   ~}~A ~A~%" (length (funcall path-thunk))
              (class-name (class-of node))
              (class-name (class-of (car (funcall path-thunk)))))
      (values)))))
```

But I need to use `(:before node)` and `(:after node)` action in `traverse-with-path` to push/pop the path.  Here's an attempt to do so, which does not work if the caller also passes actions over generic nodes---either one is shadowed.

```
(defun traverse-with-path (node action-list &rest args)
  "Like 'traverse', but actions receives a thunk that reutnrs a reverse
list of the current node's ascendants following the current node.
For example, `car` of the ascendants is the node's immediate parent."
  (declare (type node node)
           (values node &optional))
  (let* ((traversal-path nil)
         (wrapped-action-list
           `(,(action (:before node node _path-thunk)
                (declare (ignore _path-thunk))
                (push node traversal-path)
                node)
              ,@action-list
              ,(action (:after node node _path-thunk)
                 (declare (ignore _path-thunk))
                 (pop traversal-path)
                 node))))
    (apply #'traverse node wrapped-action-list
           (lambda () (cdr traversal-path))
           args)))
```

Any ideas?

shirok:
Found a workaround.  Will submit another PR that supersedes this.

shirok:
Superseded by https://github.com/coalton-lang/coalton/pull/1300.  Closing.

PR #1292, opened 2024-10-08
Correct span labeling

### LINKED ISSUE #1248: Inconsistent span labeling
Several test cases exhibit inconsistent labeling of zero-length error spans, for instance the following, which labels the error as 4:15, while the note points to column 14.

https://github.com/coalton-lang/coalton/blob/67262cc3c1863ccd3b6d12eb296f001a4fa7a7f6/tests/parser-test-files/bad-files/parse-let.1.error#L4,L5

See also 

```
define-class.16.error
define-instance.7.error
parse-and.1.error
parse-attribute.5.error
parse-body.1.error
parse-cond.1.error
parse-cond.4.error
parse-do.4.error
parse-fn.1.error
parse-fn.2.error
parse-for.1.error
parse-for.2.error
parse-for.3.error
parse-for.4.error
parse-for.5.error
parse-if.1.error
parse-if.2.error
parse-if.3.error
parse-if.4.error
parse-let.2.error
parse-let.5.error
parse-lisp.1.error
parse-lisp.2.error
parse-loop.1.error
parse-loop.2.error
parse-match.1.error
parse-match.2.error
parse-or.1.error
parse-the.1.error
parse-the.2.error
parse-unless.1.error
parse-unless.2.error
parse-when.1.error
parse-when.2.error
parse-while-let.1.error
parse-while-let.2.error
parse-while-let.3.error
parse-while-let.4.error
parse-while-let.5.error
parse-while-let.6.error
parse-while-let.7.error
parse-while-let.8.error
parse-while-let.9.error
parse-while.1.error
parse-while.2.error
parse-while.3.error
parse-while.4.error
```

jbouwman:
Update tests with incorrect spans, and add missing numbering for direct reexecution of failed tests.

Add span-relative error message options to cursor, and track last consumed element.

Move source-related generics to coalton-impl/source (source-available-p, source-name, source-stream), and break up source-aware condition reporting function.

Closes #1248


tests/coalton-tests.lisp
```diff
@@ -10,6 +10,8 @@
 (defun %run-tests (name)
   (run-test-file (format nil "tests/test-files/~a" name)))
 
+;; (coalton-tests:run-test "tests/test-files/define-class.txt" 100)
```
Izaakwltn, line 13: Is this just example use code? I'd say either take it out or flesh out the comment with 
```
;; usage example:
;; (coalton-tests...
```

Izaakwltn (approved):
One nit, otherwise lgtm

tests/coalton-tests.lisp
jbouwman: ack, a stray -- zapped. thank you.

PR #1294, opened 2024-10-08
Add `define-type-alias`

YarinHeffes:
This is an implementation of type aliases, including:

```lisp
;; a simple alias
(define-type-alias Coordinate Integer)

;; aliases defined on other aliases
(define-type-alias Point (Tuple Coordinate Coordinate))
(define-type-alias Translation (Point -> Point))

;; parametric aliases
(define-type-alias (UnaryOperator :a) (:a -> :a))
(define-type-alias IntegerUnaryOperator (UnaryOperator Integer))

;; aliases do *not* need to be of kind *
(define-type-alias T Tuple)

;; alias parameters do *not* need to be of kind *
(define-type-alias (IntegerCollection :col) (:col Integer))

(define-type-alias (U :a :b :c) (:b :c -> (UnaryOperator :a)))
```

Error messages include type alias information
```lisp
[Point := (Tuple [Coordinate := Integer] [Coordinate := Integer])]
```

*Notes*

First, parametric type aliases cannot be partially applied.
```lisp
(define-type-alias (T :a :b) (:a :b))
(define-type-alias (U1 :a) (List :a))
(define-type-alias U2 List)

;; this does not compile, because U1 is not completely applied.
(define-type-alias V (T U1 Integer))

;; but this compiles just fine
(define-type-alias V (T U2 Integer))
```

Second, since `the` does not modify types except by applying type-variable substitutions, the line
```lisp
(define-type-alias Coordinate Integer)
(define x (the Coordinate (the Integer 5)))
```
will compile, but it will not store the alias `Coordinate` for the variable `x`, and it will not display in error messages. 

The functions `describe-type-alias` and `describe-type-of` are implemented to display the definition of a type alias and the aliases of a defined symbol, respectively.

The function `set-type-printing-mode` takes an argument `mode` which can be one of `'(:types :aliases :types-and-aliases)` and determines whether to print types, aliases, or both, whenever a type is printed. This setting can be configured before compiling Coalton using `(setf (get ':coalton-config ':type-printing-mode) mode)`.


stylewarning:
I think this PR should be blocked on a complete and tested implementation. I am sympathetic to incomplete library additions, but language features need to be complete and bullet-proof.

Izaakwltn:
I think it would be a good idea to add some documentation in these files:
- `docs/intro-to-coalton.md`
- `docs/coalton-documentation-guide.md`
- maybe in `docs/coalton-lisp-interop.md`, if there are any promises (or lack thereof) about lisp alias behavior.

docs/coalton-documentation-guide.md
```diff
@@ -33,6 +33,16 @@ that contains a filling of various sweet or savoury ingredients."
...
 ```
 
+### `define-alias`
+
+`define-alias` allows for the same docstring style as `define`.
+
+```lisp
+(coalton-toplevel
+
+  (define-alias Index Integer
```
stylewarning, line 43: In retrospect I honestly think we should have followed the pattern of

    (define <thing> <doc> <value>)

but you're right to follow existing behavior

docs/coalton-documentation-guide.md
```diff
@@ -33,6 +33,16 @@ that contains a filling of various sweet or savoury ingredients."
     (Meat String)  "A meat pie with the type of meat."))
 ```
 
+### `define-alias`
+
+`define-alias` allows for the same docstring style as `define`.
+
+```lisp
+(coalton-toplevel
+
```
stylewarning, line 42: no leading empty line

src/faux-macros.lisp
```diff
@@ -33,6 +33,9 @@
 (define-coalton-editor-macro coalton:define-type (name &body definition)
   "Create a new algebraic data type named NAME. (Coalton top-level operator.)")
 
+(define-coalton-editor-macro coalton:define-alias (name &body definition)
+  "Create a new alias named NAME. (Coalton top-level operator.)")
```
stylewarning, line 37: This is specifically a type alias, not an alias for any name.

Maybe we should go as far as rename the operator

    define-type-alias

docs/intro-to-coalton.md
```diff
@@ -272,6 +272,41 @@ Type definitions introduce type constructors. For example, we may construct a so
...
+Outside of a Coalton expression, `describe-type-of` displays the type of a symbol, including its aliases, and returns the type. `describe-alias` displays the alias along with its base type and returns the base type.
+
+```lisp
+COALTON-USER> shifted-coordinate
+#.(TUPLE 1 0)
+
+COALTON-USER> (type-of 'shifted-coordinate)
+(TUPLE INTEGER INTEGER)
+
+COALTON-USER> (describe-type-of 'shifted-coordinate)
```
stylewarning, line 302: I think these functions and their output may need some workshopping.

src/parser/toplevel.lisp
```diff
@@ -265,6 +276,22 @@
 (deftype toplevel-define-type-list ()
   '(satisfies toplevel-define-type-list-p))
 
+(defstruct (toplevel-define-alias
+            (:include toplevel-definition)
+            (:copier :nil))
```
stylewarning, line 281: nil not :nil

src/typechecker/base.lisp
```diff
@@ -60,6 +61,8 @@ This requires a valid PPRINT-VARIABLE-CONTEXT")
      (with-pprint-variable-scope ()
        ,@body)))
 
+(defvar *pprint-aliases* nil)
```
stylewarning, line 64: should have docs if exported

stylewarning (changes requested):
Just some initial thoughts. Not a full review.

stylewarning:
@YarinHeffes Can you update the PR description with an up-to-date account of the PR?

docs/intro-to-coalton.md
YarinHeffes: Do you have any feedback as far as what is good/bad about them?

docs/coalton-documentation-guide.md
YarinHeffes: resolved in [5851da6](https://github.com/coalton-lang/coalton/pull/1294/commits/5851da61b96726e42b26b0e2fc066d3136ae4456)

src/faux-macros.lisp
YarinHeffes: resolved in [57366b7](https://github.com/coalton-lang/coalton/pull/1294/commits/57366b7f79147a4d910d218292688914c7a876a8)

src/parser/toplevel.lisp
YarinHeffes: resolved in [ff0e7f7](https://github.com/coalton-lang/coalton/pull/1294/commits/ff0e7f71131c98c21d973ad71b31bf69b222b51d)

src/typechecker/base.lisp
YarinHeffes: resolved in [9b3beda](https://github.com/coalton-lang/coalton/pull/1294/commits/9b3beda0bbc78b5dd029a0f229c4c08f8203bd7f)

stylewarning (changes requested):

src/parser/toplevel.lisp
```diff
@@ -1107,6 +1142,97 @@ consume all attributes")))
...
+         (parse-error "Malformed type alias definition"
+                      (note source (cst:second form)
+                            "nullary type aliases should not have parentheses")
+                      (help source (cst:second form)
+                            (lambda (existing)
+                              (subseq existing 1 (1- (length existing))))
+                            "remove unnecessary parentheses")))
+
+       ;; (define-type-alias (name type-variables+) ...)
+       (loop :for vars := (cst:rest (cst:second form)) :then (cst:rest vars)
```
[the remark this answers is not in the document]
YarinHeffes: This takes place during the type checking phase, similar to `define-struct` and `define-type`

src/typechecker/define-type.lisp
```diff
@@ -290,6 +312,29 @@
...
+                                 :when (not (member (tc:tyvar-id (partial-type-env-lookup-var partial-env tyvar parsed-type))
+                                                    used-variables))
+                                   :collect tyvar))
+         (number-of-unused-variables (length unused-variables)))
+    (unless (zerop number-of-unused-variables)
+      (tc-error (format nil "Alias type variable~P defined but never used" number-of-unused-variables)
+                (tc-note parsed-type "Type alias ~S defines unused type variable~P ~{:~A~^ ~}"
+                         (parser:identifier-src-name (parser:type-definition-name parsed-type))
+                         number-of-unused-variables
+                         (mapcar (lambda (str) (subseq str 0  (- (length str) (1+ (position #\- (reverse str))))))
```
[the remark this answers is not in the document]
YarinHeffes: addressed in [8619ceb](https://github.com/coalton-lang/coalton/pull/1294/commits/8619ceba30bdc39e007b943a068655ce1fe90dbc)

src/typechecker/define-type.lisp
```diff
@@ -420,6 +465,12 @@
                                                  `(member ,@(mapcar #'tc:constructor-entry-compressed-repr ctors)))
                                                 (t
                                                  name))
+                                :aliased-type (let ((parser-aliased-type
+                                                      (parser:type-definition-aliased-type type)))
+                                                (when parser-aliased-type
```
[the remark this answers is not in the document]
YarinHeffes: addressed in [8619ceb](https://github.com/coalton-lang/coalton/pull/1294/commits/8619ceba30bdc39e007b943a068655ce1fe90dbc)

src/typechecker/types.lisp
```diff
@@ -71,7 +74,8 @@
 ;;; Types
 ;;;
 
-(defstruct (ty (:constructor nil)))
+(defstruct (ty (:constructor nil))
+  (alias nil :type (or null ty-list) :read-only nil))
```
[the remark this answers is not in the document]
YarinHeffes: addressed in [8619ceb](https://github.com/coalton-lang/coalton/pull/1294/commits/8619ceba30bdc39e007b943a068655ce1fe90dbc)

src/typechecker/types.lisp
```diff
@@ -199,6 +207,36 @@
...
+              (tapp-from type2))
+         (ty= (tapp-to type1)
+              (tapp-to type2))))
+
+  (:method ((type1 tgen) (type2 tgen))
+    (equalp (tgen-id type1)
+            (tgen-id type2)))
+
+  (:method (type1 type2)
+    (declare (ignore type1)
```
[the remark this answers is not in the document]
YarinHeffes: addressed in [8619ceb](https://github.com/coalton-lang/coalton/pull/1294/commits/8619ceba30bdc39e007b943a068655ce1fe90dbc)

src/parser/toplevel.lisp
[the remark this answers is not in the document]
Jason94: (https://github.com/coalton-lang/coalton/pull/1294/files#diff-ae01ad44c79e0347430c5826a5c355990288337845de14a0bac46c3d601b25efL79)

src/debug.lisp
```diff
@@ -173,9 +173,24 @@
   "Lookup the type of value SYMBOL in the global environment"
   (tc:lookup-value-type entry:*global-environment* symbol))
 
+(defun coalton:describe-type-of (symbol)
+  "Print the type of value SYMBOL along with its type aliases and return it"
+  (let ((tc:*pprint-type-aliases* t)
+        (type (tc:lookup-value-type entry:*global-environment* symbol)))
```
Jason94, line 179: It looks like these are inconsistent with how it prints aliases that are imported from other packages.

If you :use a package then this will print the alias without the package. If you don't :use a package, but do something like this:
```lisp
  (declare adds (aliases2:Int -> Integer -> Integer))
```
then you will get an error message that does indeed print the package:
```
  --> <macroexpansion>:1:15
   |
 1 |  (COALTON (ADDS "hi" 2))
   |                 ^^^^ Expected type '[ALIASES2:INT := INTEGER]' but got 'STRING'
```

I think, since this is for debugging purposes, it would be helpful to always print the package that a type is from unless it is from the current package or coalton-prelude. Maybe any of the library functions.

---
Here is another example. In this error message, we don't `:use` the package. In the error message, the alias is printed with one : but the type it refers to (from the same package) is printed with two ::'s:

"aliases2"
```lisp
(in-package :cl-user)
(defpackage :aliases2
  (:use
   #:coalton
   #:coalton-prelude)
  (:export
   :Int
   :Ham
   :Swiss
   :Cheese))
(in-package :aliases2)

(coalton-toplevel
  (define-type MoldyMilk Ham Swiss)

  (define-type-alias Cheese MoldyMilk))

(coalton-toplevel
  (define-type-alias Int Integer))
```

aliases (main package)
```lisp
(in-package :cl-user)
(defpackage :aliases
  (:use
   #:coalton
   #:coalton-prelude))
   ;; #:aliases2)) <---- not using the other package
(in-package :aliases)

(coalton-toplevel
  (declare value-of-cheese (aliases2:Cheese -> Integer))
  (define (value-of-cheese chs)
    (match chs
      ((aliases2:Ham) 10)
      ((aliases2:Swiss) 20))))

(coalton (value-of-cheese "not cheese"))
```


```
error: Type mismatch
  --> <macroexpansion>:1:26
   |
 1 |  (COALTON (VALUE-OF-CHEESE "not cheese"))
   |                            ^^^^^^^^^^^^ Expected type '[ALIASES2:CHEESE := ALIASES2::MOLDYMILK]' but got 'STRING'
```

tests/test-files/define-type-alias.txt
```diff
@@ -0,0 +1,142 @@
+================================================================================
+1 Define type alias
```
Jason94, line 2: Should add tests around package boundaries. Permutations that come to mind:

* use an alias for a standard type from another package that has been imported with use
* Reference an alias for a standard type from another package without `:use`

- Both of the above to a type that is from the same package but was not exported.
- Both of the above, but the alias in package 2 is to a type  from a third package which is not `:use`d in the main package

Jason94:
Testing uninterning aliases, if you:

1. Point alias A -> a type, and point alias B -> alias A
2. then unintern A

B will still continue to function. You can use it to define new type signatures - works just fine. The error message you get looks like this:

Note the `#:A` instead of `A`:
```
error: Type mismatch
  --> <macroexpansion>:1:18
   |
 1 |  (COALTON (BAD-ADD "hi" 2))
   |                    ^^^^ Expected type '[B := #:A := INTEGER]' but got 'STRING'
```

This seems like reasonable behavior, so much as anything would be. This (hopefully) won't come up much, but it's worth verifying that it doesn't crash everything if it does.

Here is some code to play around with this:
```lisp
(coalton-toplevel
  ;; (define-type-alias A Integer)
  ;; (define-type-alias B A)

  ;; (declare a-add (A -> A -> Integer))
  ;; (define a-add +)
 )

(coalton-toplevel
  (declare bad-add (B -> B -> Integer))
  (define (bad-add b1 b2) (+ b1 b2)))

(coalton (bad-add 1 2))
(cl:unintern 'A)
```

src/debug.lisp
YarinHeffes: I will take a closer look at the behavior of `describe-type-of`. But regarding your examples, I believe those are all expected behaviors.

First, the use of `:` versus `::` depends on whether is a symbol is exported. If a package `p` exports a symbol `s`, and you do not want to `:use` package `p`, then you can use the symbol with `p:s`. If `p` does not export `s`, then you can still use it with `p::s`. In your example, `aliases2` exports `cheese` but not `moldymilk`.

Second, when package `p` `:uses` package `q`, the symbols from `q` receive the same treatment locally, in `p`, as any other symbol defined in `p`, so my personal opinion is that it is not inconsistent for those types aliases to be printed without the `q:` prefix.

Jason94: That makes sense 👍

tests/test-files/define-type-alias.txt
```diff
@@ -0,0 +1,142 @@
...
+
+(define-type-alias (UnaryOperator :a) (:a -> :a))
+
+================================================================================
+4 Define type alias
+================================================================================
+
+(package coalton-unit-tests)
+
+(define-type-alias (ReverseTranslationRules :a :b) (:b -> :a))
```
Jason94, line 31: I believe I found a bug when there are multiple type variables in the alias. It seems to only detect the first argument passed in. Here is the test code:

```lisp
(coalton-toplevel
  (define-type-alias (MyTuple :a :b) (Tuple :a (Tuple Integer :b)))

  (declare sum-mytup ((MyTuple Integer Integer) -> Integer))
  (define (sum-mytup (Tuple a (Tuple b c))) (+ a b c)))
```
Here is the error that I am getting:
```
  (DEFINE-TYPE-ALIAS #
    ...)
  ...))
error: Incomplete type alias application
  --> <macroexpansion>:4:21
   |
 4 |    (DECLARE SUM-MYTUP ((MYTUPLE INTEGER INTEGER) -> INTEGER))
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Type alias MYTUPLE is applied to 1 argument, but 2 arguments are required.
```

In the test here, it looks like an alias for a type with multiple variables is defined, but not actually used.

Jason94 (changes requested):

Jason94:
I also believe I may have found a bug around higher kinded types, but I'm not sure.

Given this ReaderT type, where ReaderT expects a `*->*` for `:m`, a type-alias defining an alias on the `ST` monad:
```lisp
  (define-type (ReaderT :env :m :value)
    (ReaderT (:env -> :m :value)))

(coalton-toplevel
  (define-struct Configuration
    (overdraft-protection Boolean))

  (define-type-alias (BankState :val)
    (ST Integer :val)))
```
this simple attempt at aliasing a `ReaderT` type:
```lisp
(coalton-toplevel
    (define-type-alias (BankM :val)
    (ReaderT Configuration BankState :val)))
```
gives this error:
```
  (DEFINE-TYPE-ALIAS #
    ...)))
The variable COALTON-IMPL/TYPECHECKER/PARSE-TYPE::TYPE-ALIAS is unbound.
```

IT works if I don't try to "reuse" my alias for the State:
```lisp
(coalton-toplevel
  (define-type-alias (BankM :val)
    (ReaderT Configuration (ST Integer) :val)))
```

I think something here isn't quite right. I think this should work. It's possible that the correct solution is to instead define the alias like this, which would mirror typeclasses:
```lisp
  (define-type-alias BankState
    (ST Integer))
```
but the current implementation of `define-type-alias` seems to require that all of the type parameters be "filled in".

tests/test-files/define-type-alias.txt
YarinHeffes: Thank you for pointing this out, I believe I know what the problem is, and I will try to address it tomorrow.

YarinHeffes:
You bring up a great point. Unfortunately, this was by design. I intended for type aliases to 1. always be fully applied and 2. always alias types of kind `*`. This means that neither:
```lisp
(coalton-toplevel
  (define-type-alias (BankM :val)
    (ReaderT Configuration BankState :val)))
```
nor
```lisp
  (define-type-alias BankState
    (ST Integer))
```
will work. However, the error messages do not make that clear. It would probably be ideal to allow for both of the above examples. I will look into it. Please, let me know if you have any further thoughts about this. I'm open to and welcome feedback and further discussion! Thank you for pointing this out @Jason94.

Jason94:
> You bring up a great point. Unfortunately, this was by design. I intended for type aliases to 1. always be fully applied and 2. always alias types of kind `*`. This means that neither:
> 
> ```commonlisp
> (coalton-toplevel
>   (define-type-alias (BankM :val)
>     (ReaderT Configuration BankState :val)))
> ```
> 
> nor
> 
> ```commonlisp
>   (define-type-alias BankState
>     (ST Integer))
> ```
> 
> will work. However, the error messages do not make that clear. It would probably be ideal to allow for both of the above examples. I will look into it. Please, let me know if you have any further thoughts about this. I'm open to and welcome feedback and further discussion! Thank you for pointing this out @Jason94.

Got it! Yes, I think a clearer error message would be helpful. Something like "Type aliases only supported for '*' types" or something that would be maximally clear to the non-Haskell initiated. In any case, I think this would be great to have someday, but certainly not worth holding up this PR. Maybe we could also add an issue to the repo for a future enhancement?

YarinHeffes:
@Jason94, I gave it some more thought and remembered my rational for disallowing the example:
```lisp
(coalton-toplevel
  (define-type-alias (BankM :val)
    (ReaderT Configuration BankState :val)))
```
namely, when someone uses a type alias, it is short-hand for writing out the type which is aliased, and nothing more. The example you gave is tricky because the kind arity of the `BankState` type alias is the same as the arity of `(ST Integer)`, so it may seem intuitive that you can partially apply `BankState` in this example. But, consider instead
```lisp
(coalton-toplevel
  (define-type-alias (BankState :val) (ST :val Integer))
  (define-type-alias (BankM :val)
    (ReaderT Configuration BankState :val)))
```
In this case, it is not clear with what to replace BankState. I hope this explanation makes sense. 

The second example,
```lisp
  (define-type-alias BankState
    (ST Integer))
```
I think should be implemented and allowed. Let me know if you have more thoughts about either of these examples, especially if you disagree with my point regarding the first example.

Jason94:
> @Jason94, I gave it some more thought and remembered my rational for disallowing the example:
> 
> ```commonlisp
> (coalton-toplevel
>   (define-type-alias (BankM :val)
>     (ReaderT Configuration BankState :val)))
> ```
> 
> namely, when someone uses a type alias, it is short-hand for writing out the type which is aliased, and nothing more. The example you gave is tricky because the kind arity of the `BankState` type alias is the same as the arity of `(ST Integer)`, so it may seem intuitive that you can partially apply `BankState` in this example. But, consider instead
> 
> ```commonlisp
> (coalton-toplevel
>   (define-type-alias (BankState :val) (ST :val Integer))
>   (define-type-alias (BankM :val)
>     (ReaderT Configuration BankState :val)))
> ```
> 
> In this case, it is not clear with what to replace BankState. I hope this explanation makes sense.
> 
> The second example,
> 
> ```commonlisp
>   (define-type-alias BankState
>     (ST Integer))
> ```
> 
> I think should be implemented and allowed. Let me know if you have more thoughts about either of these examples, especially if you disagree with my point regarding the first example.

Yes, you're right. The more I think about it, the more it makes sense to treat aliases as a simple one-to-one swap. So long as the error message on the first example is clear, the second one looks like the best approach.

tests/test-files/define-type-alias.txt
YarinHeffes: This bug should be fixed with my latest commit. I will clean up my code, add detailed comments and documentation, and add test cases. I wanted to note that the function `+` takes two arguments, so if anyone wants to try this bug again, make sure to replace `(+ a b c)` with `(+ a (+ b c)` or something similar.

Jason94 (approved):
Recent changes & fixes are working great! This is compiling now:

```lisp
  (define-struct Configuration
    (overdraft-protection Boolean))

  (define-type-alias BankState
    (ST Integer))

  (define-type-alias BankM
    (EnvT Configuration BankState))
```

docs/intro-to-coalton.md
Jason94: I already approved, but I'm not sure about the idea of showing the aliases in error messages anymore. It's maybe useful at first, but if you start stacking aliases it gets very unwieldy very fast:

```
    error: Type mismatch
      --> <macroexpansion>:4:2
        |
     4  |     (DEFINE (CREATE-ACCOUNT NAME INITIAL-BALANCE)
        |  ___^
     5  | |     (DO
     ...
     9  | |       (ACCOUNT-VALID? ACCOUNT)
     10 | |       (PURE ACCOUNT)))
        | |______________________^ Expected type '([ACCOUNTNAME := STRING] → [BALANCE := INTEGER] → ([BANKM := ((ENVT CONFIGURATION) [BANKSTATEM := (ST [BANKSTATE := (M:MAP [ACCOUNTNAME := STRING] ACCOUNT)])])] (RESULT STRING ACCOUNT)))' but got type '([ACCOUNTNAME := STRING] → [BALANCE := INTEGER] → ([BANKM := ((ENVT CONFIGURATION) [BANKSTATEM := (ST [BANKSTATE := (M:MAP [ACCOUNTNAME := STRING] ACCOUNT)])])] ACCOUNT))'
```

src/typechecker/base.lisp
```diff
@@ -78,9 +78,8 @@ This requires a valid PPRINT-VARIABLE-CONTEXT")
                        (with-pprint-variable-context ()
                          (apply #'format nil format-string format-args))))
 
-(defun tc-note (located format-string &rest format-args)
-  (let ((*pprint-type-aliases* t)) 
-    (apply #'tc-location (source:location located) format-string format-args)))
+(defun tc-note (located format-string &rest format-args)
```
Jason94, line 81: @YarinHeffes Thank you! Two suggestions for possible enhancements:

1. Allow three options: show the original type, show [Alias := Original type], or just show the name of the alias in the error messages. Personally, if I'm looking at code that says one thing then it'd be most helpful to have the error messages say that same thing.
2. Possibly a good candidate for the coalton config? ([From this PR](https://github.com/coalton-lang/coalton/pull/1205/files))

Just dropping some thoughts, though. This is great and I don't want to hold it up by nitpicking.

YarinHeffes: I addressed both points here [cf77681](https://github.com/coalton-lang/coalton/pull/1294/commits/cf7768169747eb1bf116c87d15e1222e5bd40a87)

Jason94: Works great!

YarinHeffes: Thanks for all your feedback!

stylewarning:
@YarinHeffes Can you resolve the conflicts?

stylewarning (approved):

PR #1295, opened 2024-10-09
Rename bound variables

### LINKED ISSUE #1293: Recursive definition failing to compile with inlining enabled
I'm working through current benchmarks and ran across one that fails to compile with inlining enabled:
```
(coalton-toplevel

  (declare fib (Integer -> Integer))
  (define (fib n)
    (when (== n 0)
      (return 0))

    (when (== n 1)
      (return 1))

    (+ (fib (- n 1))
        (fib (- n 2)))))
```
```
  error: 
      (DECLARE FIB ...)
      ...))
    Invalid scc binding group. This should have been detected during typechecking.
```

amorphedstar:
This fixes #1293

The codegen assumes that each variable only appears with one meaning, and in particular can never be bound in a context where it is already bound. When recursively unrolling functions with `node-let`, this was no longer true. This PR adds bound-variable renaming as an option as a part of `apply-ast-substitution`.

### DIFF
--- a/src/codegen/ast-substitutions.lisp
+++ b/src/codegen/ast-substitutions.lisp
@@ -5,6 +5,7 @@
   (:import-from
    #:coalton-impl/codegen/traverse
    #:action
+   #:*traverse*
    #:traverse)
   (:local-nicknames
    (#:parser #:coalton-impl/parser)
@@ -31,21 +32,23 @@
 (deftype ast-substitution-list ()
   '(satisfies ast-substitution-list-p))
 
-(defun apply-ast-substitution (subs node)
+(defun apply-ast-substitution (subs node &optional (rename-bound-variables nil))
   "Substitute variables in the tree of `node` with other nodes specified
 in `subs`. Throw an error if a variable to be substituted is bound in
-a subtree of `node`."
+a subtree of `node`. Also rename all bound variables if `rename-bound-variables`
+is true."
   (declare (type ast-substitution-list subs)
            (type node node)
+           (type boolean rename-bound-variables)
            (values node &optional))
   (traverse
    node
    (list
-    (action (:after node-variable node)
+    (action (:after node-variable node subs)
       (alexandria:when-let
           ((res (find (node-variable-value node) subs :key #'ast-substitution-from)))
         (ast-substitution-to res)))
-    (action (:after node-lisp node)
+    (action (:after node-lisp node subs)
       (multiple-value-bind (let-bindings lisp-var-bindings)
           (loop :for (lisp-var . coalton-var) :in (node-lisp-vars node)
                 :for new-var := (gensym (symbol-name coalton-var))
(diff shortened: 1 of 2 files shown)


src/codegen/ast-substitutions.lisp
```diff
@@ -70,14 +73,34 @@ a subtree of `node`."
                :type (node-type node)
                :bindings let-bindings
                :subexpr new-lisp-node)))))
-    (action (:before node-direct-application node)
+    (action (:before node-direct-application node subs)
       (when (find (node-direct-application-rator node) subs :key #'ast-substitution-from)
```
YarinHeffes, line 77: I believe the issue is localized to the `node-direct-application` pass. The safest way to handle this would be to add an additional argument `rename-bound-variables` so that you can be sure that you can track variable renaming separately from the other substitutions. A solution that seems to work, at least well enough to pass the tests, is to replace this `when` with the following

```lisp
 (if (and rename-bound-variables
         (node-variable-p (ast-substitution-to res)))
    (make-node-direct-application
     :type (node-type node)
     :rator-type (node-direct-application-rator-type node)
     :rator (node-variable-value (ast-substitution-to res))
     :rands (node-direct-application-rands node))
    (util:coalton-bug
     "Failure to apply ast substitution on variable ~A to node-direct-application"
     (node-direct-application-rator node)))
```
This seems to be the only issue. Once it is addressed, this PR is ready to merge in my opinion. @amorphedstar let me know if you would want me to work on this, although I would definitely appreciate your input either way.

YarinHeffes (changes requested):
Looks good overall - just needs to address the issue with handling `direct-node-application`.

src/codegen/ast-substitutions.lisp
amorphedstar: Yeah, that looks about right. I think the thing that concerned me is that the kind of rewriting used for inlining here is different from the most general AST substitution which will replace `node-variable` with arbitrary AST nodes.
For the current use case in inlining, it just rewrites variable names in the function body, and now it also recursively renames variables bound in `let` blocks, but this is just about the names, and really doesn't do anything with AST nodes.
So I don't know if it makes sense to use the same
```lisp
(defstruct ast-substitution
  (from (util:required 'from) :type parser:identifier :read-only t)
  (to   (util:required 'to)   :type node              :read-only t))
```
for both cases.
Maybe it would be better to separate this to a more specialized variable renaming function, with a more specialized substitution structure like this:
```lisp
(defstruct identifier-renaming
  (from (util:required 'from) :type parser:identifier :read-only t)
  (to   (util:required 'to)   :type parser:identifier :read-only t))
```
Do you think that sounds like a better structure, or is it nicer to have the similar functionality together in one function as it currently is?


One other thing I think would be good to know: @stylewarning mentioned that there are some subtleties for correctness in inlining.
The function arguments obviously need to be renamed in the body, but we found that we also needed to rename type variables so that they aren't unified over different calls, and now bound variables also need to be renamed.
I wonder if there is a good reference listing exactly what is necessary, so we can have more confidence that we aren't missing something else still.

YarinHeffes: I think the structure `identifier-renaming` is the most correct thing to do, even if the code will be a bit more verbose.

YarinHeffes:
Closing with #1363.

PR #1296, opened 2024-10-09
Remove location parameter from check-duplicates

jbouwman:
source:location is used to report duplicate locations, and doesn't need to be provided as a parameter.


Izaakwltn (approved):
lgtm

PR #1298, opened 2024-10-09
Add `Into :a String => Into (Cell :a) String` instance

Izaakwltn:
This adds an instance of `Into :a String => Into (Cell :a) String`

### DIFF
--- a/library/cell.lisp
+++ b/library/cell.lisp
@@ -144,6 +144,10 @@
   (define-instance (Into (Cell :a) :a)
     (define into read))
 
+  (define-instance (Into :a String => Into (Cell :a) String)
+    (define (into c)
+      (into (read c))))
+
   (define-instance (Default :a => Default (Cell :a))
     (define (default) (new (default)))))
 


library/string.lisp
```diff
@@ -195,6 +195,16 @@ does not have that suffix."
       (lisp String (lst)
         (cl:coerce lst 'cl:string))))
 
+  (define-instance (Into (List String) String)
```
stylewarning, line 198: is this the right canonical implementation?

Izaakwltn: changed to `mconcat` and moved it fully into `Coalton` land

stylewarning: Yes but why should the `String` of a list of strings be their concatenation? It sounds like a loose interpretation of what Into should mean.

Izaakwltn: I get what you mean now, I may leave the cell definition then and scrap these two

stylewarning (approved):

PR #1299, opened 2024-10-09
Add profiling to coalton-library/system

Izaakwltn:
This adds a profiling interface (besides just Time) for Coalton.

### DIFF
--- a/library/system.lisp
+++ b/library/system.lisp
@@ -8,6 +8,12 @@
    #:time
    #:sleep)
   (:export
+   #:get-real-time
+   #:get-run-time
+   #+sbcl #:get-bytes-consed
+   #:Profile
+   #:capture-profile)
+  (:export
 
    #:LispCondition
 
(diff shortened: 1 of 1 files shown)


jbouwman (approved):
looks good to me

library/system.lisp
```diff
@@ -8,6 +8,12 @@
    #:time
    #:sleep)
   (:export
+   #:get-real-time
+   #:get-run-time
+   #+sbcl #:get-bytes-consed
```
stylewarning, line 13: this should be exported by all implementations but maybe return None

library/system.lisp
```diff
@@ -62,6 +68,64 @@ While the result will always contain microseconds, some implementations may retu
...
+    (lisp UFix ()
+      (cl:get-internal-run-time)))
+
+  (declare get-real-time (Unit -> UFix))
+  (define (get-real-time)
+    "Gets the real-time."
+    (lisp UFix ()
+      (cl:get-internal-real-time)))
+
+  #+sbcl
```
stylewarning, line 89: please do not do this in the standard library

stylewarning (commented):
The conditionals were a mistake.

library/system.lisp
Izaakwltn: won't happen again

PR #1300, opened 2024-10-10
Add `traverse-with-path`

shirok:
It is a variation of `traverse`, except the actions receives a `path-thunk` argument after the visiting node.  When called within the action, `path-thunk` returns a list of ascendant nodes of the visiting node, i.e. its car is the immediate parent, and its the last element is the root node on which `traverse-with-path` is called.

This is useful for AST walker that needs wider context surrounding the node of concern.

This PR supersedes https://github.com/coalton-lang/coalton/pull/1291.  This one avoid relying on special variables.

### DIFF
--- a/src/codegen/traverse.lisp
+++ b/src/codegen/traverse.lisp
@@ -12,6 +12,7 @@
    #:make-traverse-let-action-skipping-cons-bindings
    #:*traverse*
    #:traverse
+   #:traverse-with-path
    #:traverse-with-binding-list))
 
 (in-package #:coalton-impl/codegen/traverse)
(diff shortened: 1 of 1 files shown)


src/codegen/traverse.lisp
```diff
@@ -253,6 +254,60 @@ other nodes, then it would be inappropriate to also define an action
...
+                                     (apply (action-function action) node path-thunk args))))
+                     (:after
+                      (make-action when 'node
+                                   (lambda (node path-thunk &rest args)
+                                     (let ((r (apply (action-function action)
+                                                     node
+                                                     path-thunk
+                                                     args)))
+                                       (pop traversal-path)
+                                       r)))))
```
amorphedstar, line 283: I think you could also just `prog1` this

src/codegen/traverse.lisp
```diff
@@ -253,6 +254,60 @@ other nodes, then it would be inappropriate to also define an action
...
+                   (make-action when 'node
+                                (lambda (node &rest _rest)
+                                  (declare (ignore rest))
+                                  (funcall op node))))))
```
amorphedstar, line 287: Wait, what is `op`? Is this supposed to be a copy of the `ecase` above, just with the `(apply (action-function action) node path-thunk args)` removed?

amorphedstar (commented):
This looks pretty good. I wonder if there is a more elegant way to handle multiple wrapping actions. The current traversal code treats an `action-list` almost like an `alist` so that defaults can be overridden, but this PR makes it clear that it could also be nice to use an `action-list` like an actual sequence of commands.

src/codegen/traverse.lisp
shirok: right.

src/codegen/traverse.lisp
shirok: Good catch.  Remnant of the previous version.

src/codegen/traverse.lisp
```diff
@@ -253,6 +254,66 @@ other nodes, then it would be inappropriate to also define an action
       (let ((*traverse* #'current-traverse))
         (apply *traverse* initial-node initial-args)))))
 
+(defun traverse-with-path (node action-list &rest args)
```
YarinHeffes, line 257: Nit: consider changing `node -> initial-node`, `action-list -> custom-actions`, and `args -> initial-args` to match `traverse`.

stylewarning (approved):

PR #1302, opened 2024-10-11
Remove conditionals from profiling

Izaakwltn:
There are still a few interior conditionals but nothing that affects the outward interface.

### DIFF
--- a/library/system.lisp
+++ b/library/system.lisp
@@ -7,14 +7,19 @@
    (#:math #:coalton-library/math))
   (:export
    #:gc
-   #:time
    #:sleep)
   (:export
    #:get-real-time
-   #:get-run-time
-   #+sbcl #:get-bytes-consed
-   #:Profile
-   #:capture-profile)
+   #:internal-time-units-per-second
+   #:time-units->seconds
+   #:time-units->rounded-microseconds
+   #:monotonic-bytes-consed
+
+   #:time
+   #:space
+
+   #:MeteredResult
+   #:spacetime)
   (:export
 
    #:LispCondition
@@ -48,21 +53,6 @@
       (trivial-garbage:gc :full cl:t)
       Unit))
 
-  (declare time ((Unit -> :a) -> (Tuple :a Integer)))
-  (define (time f)
-    "Run the thunk `f` and return a tuple containing its value along with the run time in microseconds.
-
-While the result will always contain microseconds, some implementations may return a value rounded to less precision (e.g., rounded to the nearest second or millisecond)."
-    (let start = (lisp Integer () (cl:get-internal-run-time)))
-    (let value = (f))
-    (let end   = (lisp Integer () (cl:get-internal-run-time)))
-    (Tuple value
-           (lisp Integer (start end)
-             (cl:values
-              (cl:round
-               (cl:* 1000000 (cl:- end start))
-               cl:internal-time-units-per-second)))))
-
   (declare sleep ((math:Rational :a) => :a -> Unit))
   (define (sleep n)
     "Sleep for `n` seconds, where `n` can be of any type with an instance of `Rational`.
@@ -81,57 +71,105 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
 
 (coalton-toplevel
 
-  (declare get-run-time (Unit -> UFix))
+  (declare get-run-time (Unit -> Integer))
   (define (get-run-time)
-    "Gets the run-time."
-    (lisp UFix ()
+    "Gets the run-time in internal time units. This is implementation specific: it may measure real time, run time, CPU cycles, or some other quantity.
+
+The difference between two successive calls to this function represents quantity accumulated during that period of time.
+
+This function is not exported as its output is too implementation specific."
+    (lisp Integer ()
       (cl:get-internal-run-time)))
 
-  (declare get-real-time (Unit -> UFix))
+  (declare get-real-time (Unit -> Integer))
   (define (get-real-time)
-    "Gets the real-time."
-    (lisp UFix ()
+    "Gets the real-time in internal time units. The difference between two successive calls to this function represents the time that has elapsed."
+    (lisp Integer ()
       (cl:get-internal-real-time)))
 
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
-  (define (get-bytes-consed)
-    "Gets the number of bytes consed (only implemented for SBCL"
-    (lisp UFix ()
-      (sb-ext:get-bytes-consed)))
+  (declare internal-time-units-per-second Integer)
+  (define internal-time-units-per-second
+    "The number of internal time units per second. This is implementation specific."
+    (lisp Integer ()
+      cl:internal-time-units-per-second))
 
-  (define-struct (Profile :a)
-    "A profile of a run function."
-    (output
-     "The output of the function" :a)
-    (run-time
-     "The run time of the run" UFix)
-    (real-time
-     "The real time of the run" UFix)
+  (declare time-units->seconds (Integer -> Fraction))
+  (define (time-units->seconds t)
+    "Converts internal time units into `Fraction` seconds."
+    (math:exact/ t internal-time-units-per-second))
+
+  (declare time-units->rounded-microseconds (Integer -> Integer))
+  (define (time-units->rounded-microseconds t)
+    "Converts internal time units into an integer number of rounded microseconds."
+    (math:round/ (* 1000000 t)
+                 internal-time-units-per-second))
+
+  (declare monotonic-bytes-consed (Unit -> (Optional Integer)))
+  (define (monotonic-bytes-consed)
+    "Returns the number of bytes consed since some unspecified point in time.
+
+The difference between two successive calls to this function represents the number of bytes consed in that period of time."
     #+sbcl
-    (bytes-consed
-     "The number of bytes consed during the run." UFix))
+    (Some (lisp Integer ()
+            (sb-ext:get-bytes-consed)))
+    #-sbcl
+    None)
 
-  (declare capture-profile ((Unit -> :a) -> (Profile :a)))
-  (define (capture-profile f)
-    "Runs a function, recording profile information and returning a Profile object."
+  ;;;
+  ;;; Function instrumentation
+  ;;;
+
+  (declare time ((Unit -> :a) -> (Tuple :a Integer)))
+  (define (time f)
+    "Run the thunk `f` and return a tuple containing its value along with the run time in microseconds.
+
+While the result will always contain microseconds, some implementations may return a value rounded to less precision (e.g., rounded to the nearest second or millisecond)."
+    (let start = (get-real-time))
+    (let value = (f))
+    (let end   = (get-real-time))
+    (Tuple value (time-units->rounded-microseconds (- end start))))
+
+  (declare space ((Unit -> :a) -> (Tuple :a (Optional Integer))))
+  (define (space f)
+    "Run the thunk `f` and return a tuple containing its value along with the approximate number of bytes consed during the course of executing f.
+
+The amount of space used may be peculiar to the implementation, such as rounding to certain page boundaries.
+
+A garbage collection will be forced prior to invoking `f`."
     (gc)
-    (let (#+sbcl
-          (start-bytes-consed (get-bytes-consed))
-          (start-run-time     (get-run-time))
-          (start-real-time    (get-real-time))
-          (value              (f))
-          #+sbcl
-          (end-bytes-consed   (get-bytes-consed))
-          (end-run-time       (get-run-time))
-          (end-real-time      (get-real-time)))
-      (Profile
-       value
-       (- end-run-time start-run-time)
-       (- end-real-time start-real-time)
-       #+sbcl
-       (- end-bytes-consed start-bytes-consed)))))
+    (let start = (monotonic-bytes-consed))
+    (let value = (f))
+    (let end   = (monotonic-bytes-consed))
+    (Tuple value (- end start)))
+
+  (define-struct (MeteredResult :a)
+    "Function output with space and timing metedata."
+    (result
+     "The result of the function." :a)
+    (time-elapsed
+     "The real time elapsed running the function (in internal time units)." Integer)
+    (bytes-consed
+     "The number of bytes consed during the run." (Optional Integer)))
+
+  (declare spacetime ((Unit -> :a) -> (MeteredResult :a)))
+  (define (spacetime f)
+    "Runs a function, gathering space and timing information and returning a `MeteredResults` object.
+
+Garbage collection will be performed before profiling is performed."
+    (gc)
+    ;; The order of these bindings ensures that slight inaccuracy of
+    ;; the measurements is shared across both bytes consed and
+    ;; elapsed time.
+    (let start-bytes-consed = (monotonic-bytes-consed))
+    (let start-real-time    = (get-real-time))
+    (let value              = (f))
+    (let end-bytes-consed   = (monotonic-bytes-consed))
+    (let end-real-time      = (get-real-time))
+    (MeteredResult
+     value
+     (- end-real-time start-real-time)
+     (- end-bytes-consed start-bytes-consed))))
+
 
 ;;;
 ;;; Gathering System information


library/system.lisp
```diff
@@ -124,7 +124,9 @@ While the result will always contain microseconds, some implementations may retu
        (- end-run-time start-run-time)
        (- end-real-time start-real-time)
        #+sbcl
-       (- end-bytes-consed start-bytes-consed)))))
```
stylewarning, line 127: should be able to remove the conditionals here and above.

library/system.lisp
```diff
@@ -86,13 +86,14 @@ While the result will always contain microseconds, some implementations may retu
...
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
+  (declare get-bytes-consed (Unit -> (Optional UFix)))
   (define (get-bytes-consed)
-    "Gets the number of bytes consed (only implemented for SBCL"
-    (lisp UFix ()
-      (sb-ext:get-bytes-consed)))
+    "Gets the number of bytes consed (only implemented in `SBCL`). "
+    (lisp (Optional UFix) ()
```
stylewarning, line 92: Put the optional outside of the Lisp call

library/system.lisp
```diff
@@ -86,13 +86,14 @@ While the result will always contain microseconds, some implementations may retu
     (lisp UFix ()
       (cl:get-internal-real-time)))
 
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
+  (declare get-bytes-consed (Unit -> (Optional UFix)))
   (define (get-bytes-consed)
```
stylewarning, line 90: maybe we should have this be `monotonic-bytes-consed` and the docs should be something like

> Return the number of bytes consed since some unspecified point in time. The difference between two successive calls to this function represents the number of bytes consed in that period of time.

library/system.lisp
```diff
@@ -86,13 +86,14 @@ While the result will always contain microseconds, some implementations may retu
     (lisp UFix ()
       (cl:get-internal-real-time)))
 
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
+  (declare get-bytes-consed (Unit -> (Optional UFix)))
```
stylewarning, line 89: Should be `Integer` since SBCL may return that.

library/system.lisp
```diff
@@ -102,9 +103,8 @@ While the result will always contain microseconds, some implementations may retu
      "The run time of the run" UFix)
     (real-time
      "The real time of the run" UFix)
-    #+sbcl
     (bytes-consed
-     "The number of bytes consed during the run." UFix))
+     "The number of bytes consed during the run (will be None except in `SBCL`)." (Optional UFix)))
```
stylewarning, line 107: Don't specify SBCL as an exception.

library/system.lisp
```diff
@@ -102,9 +103,8 @@ While the result will always contain microseconds, some implementations may retu
      "The run time of the run" UFix)
```
stylewarning, line 103: What are the units? Is UFix appropriate?

library/system.lisp
```diff
@@ -102,9 +103,8 @@ While the result will always contain microseconds, some implementations may retu
...
     (real-time
      "The real time of the run" UFix)
-    #+sbcl
     (bytes-consed
-     "The number of bytes consed during the run." UFix))
+     "The number of bytes consed during the run (will be None except in `SBCL`)." (Optional UFix)))
 
   (declare capture-profile ((Unit -> :a) -> (Profile :a)))
   (define (capture-profile f)
```
stylewarning, line 110: A cooler and more whimsical name would be `spacetime`. :)

stylewarning (changes requested):

library/system.lisp
Izaakwltn: SBCL says it might return a [consed bignum](https://github.com/sbcl/sbcl/blob/a1cecc85c4bd176c2ff3838ccc1e4c510156cbef/src/code/gc.lisp#L60C23-L60C36), but I'm garbage collecting beforehand so it should avoid numbers that big. Changing it to Integer though

library/system.lisp
Izaakwltn: They are microseconds, I think UFix makes sense?

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
 
   (declare get-run-time (Unit -> UFix))
   (define (get-run-time)
-    "Gets the run-time."
+    "Gets the run-time in microseconds."
     (lisp UFix ()
       (cl:get-internal-run-time)))
```
stylewarning, line 81: this is not guaranteed to be microseconds. you just use a library or `internal-time-units-per-second`

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
...
+          (end-bytes-consed   (monotonic-bytes-consed))
           (end-run-time       (get-run-time))
           (end-real-time      (get-real-time)))
       (Profile
        value
        (- end-run-time start-run-time)
        (- end-real-time start-real-time)
-       #+sbcl
-       (- end-bytes-consed start-bytes-consed)))))
+       (match (Tuple end-bytes-consed start-bytes-consed)
```
stylewarning, line 125: Monad `do` notation is another option :)

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
...
     (bytes-consed
-     "The number of bytes consed during the run." UFix))
+     "The number of bytes consed during the run." (Optional Integer)))
 
-  (declare capture-profile ((Unit -> :a) -> (Profile :a)))
-  (define (capture-profile f)
-    "Runs a function, recording profile information and returning a Profile object."
+  (declare spacetime ((Unit -> :a) -> (Profile :a)))
+  (define (spacetime f)
+    "Runs a function, recording space and timing information and returning a Profile object."
```
stylewarning, line 112: backticks around Profile

note that a GC will happen before the function is run

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
 
   (declare get-run-time (Unit -> UFix))
```
stylewarning, line 77: CLHS says this returns an integer

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
...
   (define (get-run-time)
-    "Gets the run-time."
+    "Gets the run-time in microseconds."
     (lisp UFix ()
       (cl:get-internal-run-time)))
 
   (declare get-real-time (Unit -> UFix))
```
stylewarning, line 83: this is the equivalent of `monotonic-bytes-consed` but for time. maybe parity with that API?

problem here is that "real time" or "run time" don't really have meaning unless one is intimately familiar with the Lisp standard.

library/system.lisp
```diff
@@ -76,55 +76,56 @@ While the result will always contain microseconds, some implementations may retu
...
   (define (get-real-time)
-    "Gets the real-time."
+    "Gets the real-time in microseconds."
```
stylewarning, line 85: same comment as above; this is not necessarily microseconds.

library/system.lisp
stylewarning, line 83: Integer

stylewarning (changes requested):

library/system.lisp
```diff
@@ -46,21 +48,6 @@
...
-    (let end   = (lisp Integer () (cl:get-internal-run-time)))
-    (Tuple value
-           (lisp Integer (start end)
-             (cl:values
-              (cl:round
-               (cl:* 1000000 (cl:- end start))
-               cl:internal-time-units-per-second)))))
-
   (declare sleep (Integer -> Unit))
   (define (sleep n)
```
stylewarning, line 52: Maybe file an issue about this. CL:SLEEP can take any non-negative real as an argument.

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
 
 (coalton-toplevel
 
-  (declare get-run-time (Unit -> UFix))
+  (declare get-run-time (Unit -> Integer))
   (define (get-run-time)
-    "Gets the run-time."
-    (lisp UFix ()
+    "Gets the run-time in `internal time units`. This is implementation specific: it may measure real time, run time, CPU cycles, or some other quantity. The difference between two successive calls to this function represents quantity accumulated during that period of time."
```
stylewarning, line 66: Don't use a code block for "internal time units".

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+    (lisp Integer ()
       (cl:get-internal-run-time)))
 
-  (declare get-real-time (Unit -> UFix))
+  (declare get-real-time (Unit -> Integer))
   (define (get-real-time)
-    "Gets the real-time."
-    (lisp UFix ()
+    "Gets the real-time in `internal time units`. The difference between two successive calls to this function represents the time that has elapsed."
```
stylewarning, line 72: Don't use a code block.

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
 
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
-  (define (get-bytes-consed)
-    "Gets the number of bytes consed (only implemented for SBCL"
-    (lisp UFix ()
-      (sb-ext:get-bytes-consed)))
+  (declare internal-time-units-per-second (Unit -> Integer))
+  (define (internal-time-units-per-second)
```
stylewarning, line 77: Maybe this should just be a variable and not a function, since its value is constant.

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+    "The number of `internal time units` per second. This is implementation specific."
+    (lisp Integer ()
+      cl:internal-time-units-per-second))
+
+  (declare time-in-seconds (Integer -> Double-Float))
+  (define (time-in-seconds t)
```
stylewarning, line 83: `time-units->seconds`

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+    "Converts `internal time units` into seconds."
+    (real:inexact/ t (internal-time-units-per-second)))
+
+  (declare time-in-microseconds (Integer -> Integer))
+  (define (time-in-microseconds t)
```
stylewarning, line 88: time-units->rounded-microseconds maybe?

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+    "Converts `internal time units` into microseconds."
```
stylewarning, line 89: no code block

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
-      (sb-ext:get-bytes-consed)))
+  (declare internal-time-units-per-second (Unit -> Integer))
+  (define (internal-time-units-per-second)
+    "The number of `internal time units` per second. This is implementation specific."
+    (lisp Integer ()
+      cl:internal-time-units-per-second))
+
+  (declare time-in-seconds (Integer -> Double-Float))
+  (define (time-in-seconds t)
+    "Converts `internal time units` into seconds."
```
stylewarning, line 84: no code block

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+  (declare time-in-microseconds (Integer -> Integer))
+  (define (time-in-microseconds t)
+    "Converts `internal time units` into microseconds."
+    (real:round/ (* 1000000 t)
+                 (internal-time-units-per-second)))
+
+  (declare monotonic-bytes-consed (Unit -> (Optional Integer)))
+  (define (monotonic-bytes-consed)
+    "Returns the number of bytes consed since some unspecified point in time.
+The difference between two successive calls to this function represents the number of bytes consed in that period of time."
```
stylewarning, line 96: newline for new paragraph

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+    #-sbcl
+    None)
 
   (define-struct (Profile :a)
-    "A profile of a run function."
+    "A profile of a function, containing space and timing information.."
     (output
-     "The output of the function" :a)
+     "The output of the function." :a)
     (run-time
```
stylewarning, line 107: Maybe just store time units, and write a couple accessors for getting seconds etc.

It's strange to me to store inexact information in a profile capture.

library/system.lisp
```diff
@@ -74,57 +61,86 @@ While the result will always contain microseconds, some implementations may retu
...
+     "The real time of the run in seconds" Double-Float)
     (bytes-consed
-     "The number of bytes consed during the run." UFix))
+     "The number of bytes consed during the run." (Optional Integer)))
 
-  (declare capture-profile ((Unit -> :a) -> (Profile :a)))
-  (define (capture-profile f)
-    "Runs a function, recording profile information and returning a Profile object."
+  (declare time ((Unit -> :a) -> (Tuple :a Integer)))
+  (define (time f)
```
stylewarning, line 115: make a `space` function too?

stylewarning (changes requested):
mostly nits and polish

library/system.lisp
Izaakwltn: [filed](https://github.com/coalton-lang/coalton/issues/1304)

library/system.lisp
```diff
@@ -3,16 +3,23 @@
...
    #:coalton-library/builtin
    #:coalton-library/classes)
+  (:local-nicknames
+   (#:real #:coalton-library/math/real))
   (:export
    #:gc
-   #:time
    #:sleep)
   (:export
+   #:get-un-time
```
stylewarning, line 12: typo

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
       (cl:get-internal-real-time)))
 
-  #+sbcl
-  (declare get-bytes-consed (Unit -> UFix))
-  #+sbcl
-  (define (get-bytes-consed)
-    "Gets the number of bytes consed (only implemented for SBCL"
-    (lisp UFix ()
-      (sb-ext:get-bytes-consed)))
+  ;; (declare *internal-time-units-per-second* (Unit -> Integer))
```
stylewarning, line 83: A proper declare can be written. Definitely no comment here though.

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+  (define  *internal-time-units-per-second*
```
stylewarning, line 84: No earmuffs. It's not a dynamic variable. It can just be the name unadorned.

stylewarning, line 84: I think we should export this.

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+      cl:internal-time-units-per-second))
+
+  (declare time-units->seconds (Integer -> Double-Float))
+  (define (time-units->seconds t)
+    "Converts internal time units into seconds."
+    (real:inexact/ t *internal-time-units-per-second*))
+
+  (declare time-units->rounded-microseconds (Integer -> Integer))
+  (define (time-units->rounded-microseconds t)
+    "Converts internal time units into microseconds."
```
stylewarning, line 96: "...into an integer number of microseconds"

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+
+The difference between two successive calls to this function represents the number of bytes consed in that period of time."
+    #+sbcl
+    (Some (lisp Integer ()
+            (sb-ext:get-bytes-consed)))
+    #-sbcl
+    None)
+
+  ;;
+  ;;
```
stylewarning, line 112: Write a comment here. Maybe "function instrumentation"

Three semicolons per ANSI spec.

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+    (let start = (get-run-time))
+    (let value = (f))
+    (let end   = (get-run-time))
+    (Tuple value (time-units->rounded-microseconds (- end start))))
+
+  (declare space ((Unit -> :a) -> (Tuple :a (Optional Integer))))
+  (define (space f)
+    "Run the thunk `f` and return a tuple containing its value along with the number of bytes consed.
+
+Whether space information is availabel is implementation specific."
```
stylewarning, line 129: typo: available

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+
+While the result will always contain microseconds, some implementations may return a value rounded to less precision (e.g., rounded to the nearest second or millisecond)."
+    (let start = (get-run-time))
+    (let value = (f))
+    (let end   = (get-run-time))
+    (Tuple value (time-units->rounded-microseconds (- end start))))
+
+  (declare space ((Unit -> :a) -> (Tuple :a (Optional Integer))))
+  (define (space f)
+    "Run the thunk `f` and return a tuple containing its value along with the number of bytes consed.
```
stylewarning, line 127: "...approximate number of bytes consed during the course of executing `f`. The amount of space used may be peculiar to the implementation, such as rounding to certain page boundaries."

library/system.lisp
stylewarning, line 129: add a paragraph: "A garbage collection will be forced prior to invoking `f`."

library/system.lisp
```diff
@@ -74,57 +66,110 @@ While the result will always contain microseconds, some implementations may retu
...
+
+Whether space information is availabel is implementation specific."
+    (gc)
+    (let start = (monotonic-bytes-consed))
+    (let value = (f))
+    (let end   = (monotonic-bytes-consed))
+    (Tuple value (- end start)))
+
+  ;;
+  ;;
```
stylewarning, line 137: no need for this section separator in my opinion, but if you'd like it, add a title of some sort.

three semicolons per ANSI spec too

stylewarning (changes requested):
some more bits and bobs

library/system.lisp
stylewarning: We might even just want to return a fraction. Why approximate it for them?

stylewarning:
@Izaakwltn Do you have thoughts on returning a rational from the ->seconds call?

Izaakwltn:
> @Izaakwltn Do you have thoughts on returning a rational from the ->seconds call?

I didn't quite know what you meant but now it makes sense- 

I think it would work well, though I don't believe there is currently a direct way to convert a fraction to various float types (besides Big-Float).

I can make an issue for an interface converting Fractions to inexact floats. Perhaps it should be a method in the Rational class .

Izaakwltn:
> > @Izaakwltn Do you have thoughts on returning a rational from the ->seconds call?
> 
> I didn't quite know what you meant but now it makes sense-
> 
> I think it would work well, though I don't believe there is currently a direct way to convert a fraction to various float types (besides Big-Float).
> 
> I can make an issue for an interface converting Fractions to inexact floats. Perhaps it should be a method in the Rational class .

#1310

library/system.lisp
```diff
@@ -81,57 +72,108 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+  (declare internal-time-units-per-second Integer)
+  (define  internal-time-units-per-second
+    "The number of internal time units per second. This is implementation specific."
+    (lisp Integer ()
+      cl:internal-time-units-per-second))
+
+  (declare time-units->seconds (Integer -> Fraction))
+  (define (time-units->seconds t)
+    "Converts internal time units into `Fraction` seconds."
+    (math:mkfraction t internal-time-units-per-second))
```
stylewarning, line 98: there's gotta be a better way than mkfraction :|

library/system.lisp
```diff
@@ -81,57 +72,108 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+  ;;;
+  ;;; Function instrumentation
+  ;;;
+
+  (declare time ((Unit -> :a) -> (Tuple :a Integer)))
+  (define (time f)
+    "Run the thunk `f` and return a tuple containing its value along with the run time in microseconds.
+
+While the result will always contain microseconds, some implementations may return a value rounded to less precision (e.g., rounded to the nearest second or millisecond)."
+    (let start = (get-run-time))
```
stylewarning, line 126: can we use real time instead? i think wall clock time is more useful than some unspecified gunk

library/system.lisp
```diff
@@ -81,57 +72,108 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+The amount of space used may be peculiar to the implementation, such as rounding to certain page boundaries.
+
+A garbage collection will be forced prior to invoking `f`."
+    (gc)
+    (let start = (monotonic-bytes-consed))
+    (let value = (f))
+    (let end   = (monotonic-bytes-consed))
+    (Tuple value (- end start)))
 
   (define-struct (Profile :a)
```
stylewarning, line 144: we might not want to call this a "Profile" but maybe a MeteredResult or something... just an idea.

Profile might be better for actual profiling in the future.

library/system.lisp
```diff
@@ -81,57 +72,108 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
-    "Runs a function, recording profile information and returning a Profile object."
+  (declare spacetime ((Unit -> :a) -> (Profile :a)))
+  (define (spacetime f)
+    "Runs a function, profiling space and timing information and returning a `Profile` object.
+
+Garbage collection will happen before profiling is performed."
     (gc)
-    (let (#+sbcl
-          (start-bytes-consed (get-bytes-consed))
+    (let ((start-bytes-consed (monotonic-bytes-consed))
```
stylewarning, line 161: let bindings aren't necessarily sequential. Use the short let instead

library/system.lisp
```diff
@@ -81,57 +72,108 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
           (end-run-time       (get-run-time))
           (end-real-time      (get-real-time)))
       (Profile
        value
        (- end-run-time start-run-time)
        (- end-real-time start-real-time)
-       #+sbcl
-       (- end-bytes-consed start-bytes-consed)))))
+       (do
+        (a <- end-bytes-consed)
```
stylewarning, line 173: Optional has Num I think, so maybe we don't even need monads

stylewarning (changes requested):

library/system.lisp
Izaakwltn: I tried using `/` and it errored on compile:

```
  read-error: 
    
      error: Type mismatch
      --> system.lisp:24:2
        |
     24 |     (define (time-units->seconds t)
        |  ___^
     25 | |     "Converts internal time units into `Fraction` seconds."
     26 | |     (math:/ t internal-time-units-per-second))
        | |______________________________________________^ Expected type '(INTEGER → FRACTION)' but got type '(INTEGER → INTEGER)'
```

stylewarning: general/

/ is only when the input and output types match

stylewarning (changes requested):
see comment about general/

library/system.lisp
stylewarning: or exact/

Izaakwltn:
> see comment about general/ or exact/

switched to `exact/`

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+The amount of space used may be peculiar to the implementation, such as rounding to certain page boundaries.
+
+A garbage collection will be forced prior to invoking `f`."
+    (gc)
+    (let start = (monotonic-bytes-consed))
+    (let value = (f))
+    (let end   = (monotonic-bytes-consed))
+    (Tuple value (- end start)))
+
+  (define-struct (MeteredResults :a)
```
stylewarning, line 145: MeteredResult

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+    "Function output with space and timing metedata."
     (output
```
stylewarning, line 147: let's rename this to "result"

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
-     "The output of the function" :a)
-    (run-time
-     "The run time of the run" UFix)
+     "The output of the function." :a)
     (real-time
```
stylewarning, line 149: time-elapsed

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
-     "The real time of the run" UFix)
-    #+sbcl
+     "The real time of the run in internal time units." Integer)
```
stylewarning, line 150: "The real time elapsed during the ..."

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
+     "The number of bytes consed during the run." (Optional Integer)))
 
-  (declare capture-profile ((Unit -> :a) -> (Profile :a)))
-  (define (capture-profile f)
-    "Runs a function, recording profile information and returning a Profile object."
+  (declare spacetime ((Unit -> :a) -> (MeteredResults :a)))
+  (define (spacetime f)
+    "Runs a function, gathering space and timing information and returning a `MeteredResults` object.
+
+Garbage collection will happen before profiling is performed."
```
stylewarning, line 158: "...will be performed..."

library/system.lisp
```diff
@@ -81,57 +71,102 @@ Sleep uses type class `Rational`'s `best-approx` instead of `Real`'s `real-appro
...
     (bytes-consed
-     "The number of bytes consed during the run." UFix))
+     "The number of bytes consed during the run." (Optional Integer)))
 
-  (declare capture-profile ((Unit -> :a) -> (Profile :a)))
-  (define (capture-profile f)
-    "Runs a function, recording profile information and returning a Profile object."
+  (declare spacetime ((Unit -> :a) -> (MeteredResults :a)))
+  (define (spacetime f)
+    "Runs a function, gathering space and timing information and returning a `MeteredResults` object.
```
stylewarning, line 156: make sure to rename "MeteredResult" here too

stylewarning (changes requested):
Last and final review, I promise. Just some naming and a few other nits.

Izaakwltn:
Changes changed

library/system.lisp
```diff
@@ -7,14 +7,19 @@
...
-   #:capture-profile)
+   #:internal-time-units-per-second
+   #:time-units->seconds
+   #:time-units->rounded-microseconds
+   #:monotonic-bytes-consed
+
+   #:time
+   #:space
+
+   #:MeteredResults
```
stylewarning, line 21: change the export

stylewarning (changes requested):

stylewarning (approved):

PR #1305, opened 2024-10-14
Changing sleep to accept (Rational :a) types

### LINKED ISSUE #1304: Sleep should be able to take any nonnegative real
As seen in the hyperspec, `cl:sleep` [can take any nonnegative real](http://clhs.lisp.se/Body/f_sleep.htm)

Currently sleep only accepts an Integer seconds.

Izaakwltn:
Fixes #1304 

I chose to use type class `Rational`'s `best-approx` instead of `Real`'s `real-approx` because it handles the approximation without arbitrary precision. 

This decision only blocks `sleep` from being called with `CReal` values.

If `Real` is preferred, we just need to pick a reasonable precision threshold.

### DIFF
--- a/library/system.lisp
+++ b/library/system.lisp
@@ -3,6 +3,8 @@
    #:coalton
    #:coalton-library/builtin
    #:coalton-library/classes)
+  (:local-nicknames
+   (#:math #:coalton-library/math))
   (:export
    #:gc
    #:time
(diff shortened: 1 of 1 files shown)


library/system.lisp
```diff
@@ -61,12 +63,15 @@ While the result will always contain microseconds, some implementations may retu
                (cl:* 1000000 (cl:- end start))
                cl:internal-time-units-per-second)))))
 
-  (declare sleep (Integer -> Unit))
+  (declare sleep ((math:Rational :a) => :a -> Unit))
   (define (sleep n)
     "Sleep for `n` seconds."
```
stylewarning, line 68: Add documentation about what n can be in a follow on paragraph, per your PR description. That was very useful design information!

library/system.lisp
```diff
@@ -61,12 +63,15 @@ While the result will always contain microseconds, some implementations may retu
                (cl:* 1000000 (cl:- end start))
                cl:internal-time-units-per-second)))))
 
-  (declare sleep (Integer -> Unit))
+  (declare sleep ((math:Rational :a) => :a -> Unit))
```
stylewarning, line 66: This will probably cons more than people care for, and so we might add some specializations for floats and the like. Maybe file an issue for that? It would be a good first issue for somebody.

library/system.lisp
```diff
@@ -61,12 +63,15 @@ While the result will always contain microseconds, some implementations may retu
...
   (define (sleep n)
     "Sleep for `n` seconds."
-    (lisp Unit (n)
-      (cl:sleep n)
-      Unit)))
+    (if (math:negative? n)
+        (error "must be a positive number.")
```
stylewarning, line 70: "nonnegative", not "positive"

maybe write "sleep must take ..."?

stylewarning (changes requested):

library/system.lisp
Izaakwltn: #1307

stylewarning (approved):

Issue #1198, opened 2024-08-08
into-iter bug


stylewarning:
```
COALTON-USER> (coalton-codegen
                (define (f)
                  (iter:into-iter iter:empty)))
```

gives

```
Internal coalton bug: Scheme ∀ :A. (:A → (ITERATOR #T121243)) should not have any free type variables.

If you are seeing this, please file an issue on Github.
   [Condition of type COALTON-IMPL/UTIL:COALTON-BUG]

Restarts:
 0: [RETRY] Retry SLIME REPL evaluation request.
 1: [*ABORT] Return to SLIME's top level.
 2: [ABORT] abort thread (#<THREAD "repl-thread" RUNNING {1001378003}>)

Backtrace:
  0: (COALTON-IMPL/UTIL:COALTON-BUG "Scheme ~S should not have any free type variables." ∀ :A. (:A → (ITERATOR #T121243)))
      Locals:
        ARGS = (∀ :A. (:A → (ITERATOR #T121243)))
        REASON = "Scheme ~S should not have any free type variables."
  1: (COALTON-IMPL/TYPECHECKER/TOPLEVEL:TOPLEVEL-DEFINE (#S(COALTON-IMPL/PARSER/TOPLEVEL:TOPLEVEL-DEFINE :NAME #S(COALTON-IMPL/PARSER/EXPRESSION:NODE-VARIABLE :SOURCE # :NAME F) :PARAMS (#1=_) :ORIG-PARAMS..
      Locals:
        COALTON-IMPL/TYPECHECKER/DEFINE::DECLARES = COMMON-LISP:NIL
        COALTON-IMPL/TYPECHECKER/DEFINE::DEFINES = (#S(COALTON-IMPL/PARSER/TOPLEVEL:TOPLEVEL-DEFINE ..))
        COALTON-IMPL/TYPECHECKER/DEFINE::ENV = #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {1023B93AE3}>
        COALTON-IMPL/TYPECHECKER/DEFINE::FILE = #S(SOURCE-ERROR/ERROR:FILE :STREAM #<SB-IMPL::STRING-INPUT-STREAM {D68EE93}> :NAME "COALTON-TOPLEVEL (NIL)")
```

stylewarning:
Fix merged, needs a test case.

PR #1314, opened 2024-10-17
Fixing `into-iter` bug

YarinHeffes:
Addresses issue #1198


Izaakwltn:
A test or two might be good so that this is tracked

stylewarning (approved):
Can you follow up with a test case added to the unit tests?

PR #1319, opened 2024-10-22
Added bits functions

Izaakwltn:
Added a few functions to `coalton-library/bits`.

`dpb` and `ldb` mimic their Common Lisp counterparts, following the same argument order.

`print-bits` is a handy debugging tool for printing bit types to a specified number of bits.

### DIFF
--- a/library/bits.lisp
+++ b/library/bits.lisp
@@ -15,7 +15,9 @@
    #:or
    #:xor
    #:not
-   #:shift))
+   #:shift
+   #:dpb
+   #:ldb))
 
 (in-package #:coalton-library/bits)
 
@@ -36,7 +38,19 @@
     (not   "The bitwise logical `not` of two integers"
            (:int -> :int))
     (shift "The arithmetic left-shift of an integer by an integer number of bits"
-           (Integer -> :int -> :int))))
+           (Integer -> :int -> :int)))
+
+  (declare dpb (Bits :a => :a -> UFix -> UFix -> :a -> :a))
+  (define (dpb newbyte size position bitstring)
+    "Deposits a byte `newbyte` of size `size` into a bitstring `bitstring` at a position `position`."
+    (lisp :a (newbyte bitstring size position)
+      (cl:dpb newbyte (cl:byte size position) bitstring)))
+
+  (declare ldb (Bits :a => UFix -> UFix -> :a -> :a))
+  (define (ldb size position bitstring)
+    "Deposits a byte of size `size` into a bitstring at a position `position`."
+    (lisp :a (bitstring size position)
+      (cl:ldb (cl:byte size position) bitstring))))
 
 #+sb-package-locks
 (sb-ext:lock-package "COALTON-LIBRARY/BITS")


stylewarning:
Please keep the original argument order where the bytespec comes first. I actually think that's better for currying. 

I would also get rid of the bytespecs entirely. Just pass them as curried args.

Izaakwltn:
> Please keep the original argument order where the bytespec comes first. I actually think that's better for currying.
> 
> I would also get rid of the bytespecs entirely. Just pass them as curried args.

Changed the order, removed Bytespec

library/bits.lisp
```diff
@@ -36,7 +40,25 @@
     (not   "The bitwise logical `not` of two integers"
            (:int -> :int))
     (shift "The arithmetic left-shift of an integer by an integer number of bits"
-           (Integer -> :int -> :int))))
+           (Integer -> :int -> :int)))
+
+  (declare dpb (Bits :a => UFix -> UFix -> UFix -> :a -> :a))
```
stylewarning, line 45: should the type of newbyte be :a ?

library/bits.lisp
```diff
@@ -36,7 +40,25 @@
...
+  (define (ldb size position bitstring)
+    "Deposits a byte of size `size` into a bitstring at a position `position`."
+    (lisp :a (bitstring size position)
+      (cl:ldb (cl:byte size position) bitstring)))
+
+  (declare print-bits (Bits :a => UFix -> :a -> String))
+  (define (print-bits width x)
+    "Prints an integer to a binary string, preserving leading zeros to a given width."
+    (lisp String (x width)
+               (cl:format cl:nil (cl:format cl:nil "~~~a,'0b" width) x))))
```
stylewarning, line 61: indentation

stylewarning, line 61: you don't need a double format. use

    ~v,'0B

library/bits.lisp
```diff
@@ -36,7 +40,25 @@
...
+      (cl:dpb newbyte (cl:byte size position) bitstring)))
+
+  (declare ldb (Bits :a => UFix -> UFix -> :a -> :a))
+  (define (ldb size position bitstring)
+    "Deposits a byte of size `size` into a bitstring at a position `position`."
+    (lisp :a (bitstring size position)
+      (cl:ldb (cl:byte size position) bitstring)))
+
+  (declare print-bits (Bits :a => UFix -> :a -> String))
+  (define (print-bits width x)
```
stylewarning, line 58: i expect "print" to be side-effectful. Maybe "bits->string"?

stylewarning (changes requested):

library/bits.lisp
```diff
@@ -36,7 +39,25 @@
...
+      (cl:dpb newbyte (cl:byte size position) bitstring)))
+
+  (declare ldb (Bits :a => UFix -> UFix -> :a -> :a))
+  (define (ldb size position bitstring)
+    "Deposits a byte of size `size` into a bitstring at a position `position`."
+    (lisp :a (bitstring size position)
+      (cl:ldb (cl:byte size position) bitstring)))
+
+  (declare bits->string (Bits :a => UFix -> :a -> String))
+  (define (bits->string width x)
```
stylewarning, line 57: since x is a type class constrained type, it's not guaranteed to be a compatible Lisp value with FORMAT. you need to convert it to an integer for this to work, or extract the bits one by one 

i suggest either fixing this or removing it

stylewarning (changes requested):

library/bits.lisp
Izaakwltn: Sounds good, I removed it. 

It would fit better in a potential future Coalton formatting library anyways.

stylewarning (approved):

PR #1326, opened 2024-11-20
Added string formatting library

Izaakwltn:
Here is a rough draft for a string formatting library.

I'm happy to change any of the names or syntax.

The library centers around a macro `format`, which takes any number of strings and concatenates them into one string.

The rest of the library provides tools for formatting different types into strings in different ways.

A few examples:
```
COALTON-LIBRARY/FORMAT> (coalton (format "Numbers:" \n (format-with-delimiter eng \n (iter:up-to 5))))
"Numbers:
zero
one
two
three
four
"
```
```
COALTON-LIBRARY/FORMAT> (coalton (format "the number " (ticked 20) " is not the same as " (ticked 21)))
"the number `20` is not the same as `21`"
```
```
COALTON-LIBRARY/FORMAT> (coalton-toplevel
                          (declare print-bin (Integer -> Unit))
                          (define (print-bin x)
                            (coalton-library/functions:trace (format "The integer " (into x) " is printed in binary as " (bin x)))))
; No value
COALTON-LIBRARY/FORMAT> (coalton (print-bin 3))
The integer 3 is printed in binary as 11
```

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -76,6 +76,7 @@
                (:file "vector")
                (:file "char")
                (:file "string")
+               (:file "format")
                (:file "slice")
                (:file "hashtable")
                (:file "queue")
--- a/examples/quil-coalton/src/package.lisp
+++ b/examples/quil-coalton/src/package.lisp
@@ -6,7 +6,8 @@
   (:use #:coalton
         #:coalton-prelude)
   (:local-nicknames
-   (#:string #:coalton-library/string))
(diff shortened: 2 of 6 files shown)


Izaakwltn:
Closing because this probably doesn't belong in the standard library, moved it to https://github.com/Izaakwltn/gertrude

stylewarning:
You should note it in the Coalton README in a new section about Coalton examples elsewhere.

PR #1327, opened 2024-12-03
Add from-optional function

Jason94:
Adds a small utility function to `optional.lisp` to get the value from an option or a default value if `None`. Shamelessly stolen from `fromMaybe` from Haskell.


library/optional.lisp
```diff
@@ -23,6 +24,13 @@
   ;; Optional
   ;;
 
+  (declare from-optional (:a -> (Optional :a) -> :a))
+  (define (from-optional def opt)
```
Izaakwltn, line 28: Personally I think maybe the variable should be 'default' for clarity

library/optional.lisp
```diff
@@ -23,6 +24,13 @@
...
+    "Get the value of OPT, or return DEF if OPT is None."
+    (match opt
```
Izaakwltn, line 30: This function has some overlap with 'unwrap-or-else', though I think that this may be a better syntax if you're not applying anything to the unwrapped element.

Izaakwltn: It may be worth abstracting this so it could work for Result too, maybe make a function 'unwrap-or-default' that work for anything satisfying the 'unwrappable' class.

Izaakwltn: Actually that abstraction already exists:

https://github.com/coalton-lang/coalton/blob/19965ebef94d7c6459988d16366eb576ded80eb0/library/classes.lisp#L321

Jason94: Ah! Drat - I was afraid this was already covered by something I couldn't find. This PR probably adds no value then.

Jason94: Yeah, I'm going to go ahead and close. Thanks!

PR #1333, opened 2024-12-26
Threads API in Standard Library

garlic0x1:
I have added a wrapper over `bordeaux-threads` to the standard library, it is only loaded if this reader conditional is satisfied: `#+(or sb-thread threads openmcl-native-threads)`, which tests for thread capabilities in SBCL, ECL, and CCL.

This might not be in scope for the standard library but I think it will be useful.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -38,7 +38,9 @@
   :depends-on (#:coalton-compiler
                #:coalton/hashtable-shim
                #:trivial-garbage
-               #:alexandria)
+               #:alexandria
+               #+(or sb-thread threads openmcl-native-threads)
+               #:bordeaux-threads)
   :pathname "library/"
   :serial t
   :components ((:file "set-float-traps")
@@ -86,6 +88,8 @@
                (:file "seq")
                (:file "system")
                (:file "file")
+               #+(or sb-thread threads openmcl-native-threads)
+               (:file "threads")
                (:file "prelude")))
 
 (cl:when (cl:member (uiop:getenv "COALTON_PORTABLE_BIGFLOAT") '("1" "true" "t") :test #'cl:equalp)
@@ -265,4 +269,6 @@
                (:file "looping-native-tests")
                (:file "monomorphizer-tests")
                (:file "inliner-tests")
-               (:file "file-tests")))
+               (:file "file-tests")
+               #+(or sb-thread threads openmcl-native-threads)
+               (:file "thread-tests")))
--- a/library/threads.lisp
+++ b/library/threads.lisp
@@ -1,0 +1,257 @@
+(coalton-library/utils:defstdlib-package #:coalton-library/threads
+  (:use
+   #:coalton
+   #:coalton-library/system
+   #:coalton-library/classes)
+  (:export
+   ;; Threads
+   #:Thread
+   #:spawn
+   #:make-thread
+   #:current-thread
+   #:all-threads
+   #:join
+   #:interrupt
+   #:destroy
+   #:alive?
+   ;; Locks
+   #:Lock
+   #:RecursiveLock
+   #:with-lock-held
+   #:with-recursive-lock-held
+   #:make-lock
+   #:make-recursive-lock
+   #:acquire-lock
+   #:acquire-recursive-lock
+   #:release-lock
+   #:release-recursive-lock
+   ;; Condition Variables
+   #:ConditionVariable
+   #:make-cv
+   #:wait-cv
+   #:notify-cv
+   #:broadcast-cv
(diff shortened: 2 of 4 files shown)


garlic0x1:
known deficiencies:
- Some options like names and timeouts aren't yet exposed as I'm not sure the idiomatic way to compensate for lack of kwargs.
- Only tested on SBCL and CCL.

possible style issues:
- Names are slightly garlicized.
- `with-lock-held` should probably be changed, other `with` macros in the stdlib are actually `call-with` macros so I am unsure of the naming convention.

stylewarning:
@garlic0x1 What do you think about having a separate ASDF subsystem for this? Not 100% sure, but maybe would be a good idea for threadless use cases.

coalton.asd
```diff
@@ -38,7 +38,9 @@
   :depends-on (#:coalton-compiler
                #:coalton/hashtable-shim
                #:trivial-garbage
-               #:alexandria)
+               #:alexandria
+               #+(or sb-thread threads openmcl-native-threads)
```
stylewarning, line 42: these should use the ASDF feature gates (ASDF has special syntax for this and your conditionals below)

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+;;; This module implements bindings to bordeaux-threads.
+;;; It will only be loaded if thread capabilities are 
+;;; detected in `cl:*features*'.
+;;;
+
+;;---------;;
+;; Threads ;;
+;;---------;;
+
+(cl:defmacro spawn (cl:&body body)
```
stylewarning, line 66: My sense is to not include this, but open for debate.

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+  `(make-thread (fn () ,@body)))
+
+(coalton-toplevel
+  (repr :native bt2:thread)
+  (define-type (Thread :a))
```
stylewarning, line 71: docs

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+
+  (define-instance (Eq (Thread :a))
+    (define (== a b)
+      (lisp Boolean (a b) (to-boolean (cl:equal a b)))))
```
stylewarning, line 75: maybe cl:eq?

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+  (declare make-thread ((Unit -> :a) -> Thread :a))
+  (define (make-thread thunk)
+    "Creates and returns a thread, which will call the function
+`thunk' with no arguments: when `thunk' returns, the thread terminates."
+    (lisp (Thread :a) (thunk)
+      (bt2:make-thread
+       (cl:lambda () (call-coalton-function thunk Unit)))))
+
+  (declare current-thread (Unit -> Thread :a))
+  (define (current-thread)
```
stylewarning, line 86: If the current thread is a Thread Integer and the environment unifies :a with Float, won't we have a type error?

I think this is a fundamental issue with this API that can creep in with this function and all below.

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+If the count is zero, blocks until `sem' can be decremented."
+    (lisp Unit (sem)
+      (bt2:wait-on-semaphore sem)
+      Unit)))
+
+;;---------;;
+;; Atomics ;;
+;;---------;;
+
+(coalton-toplevel
```
stylewarning, line 286: I'm not sure the API below should have conditionals in it. Maybe atomic values shouldn't be included in the thread API for now.

library/threads.lisp
```diff
@@ -0,0 +1,322 @@
...
+  (define-type AtomicInteger)
+
+  (declare make-atomic (#+32-bit U32 #+64-bit U64 -> AtomicInteger))
+  (define (make-atomic value)
+    "Creates an `AtomicInteger' with initial value `value'."
+    (lisp AtomicInteger (value)
+      (bt2:make-atomic-integer :value value)))
+
+  (declare atomic-cmp-and-swap (AtomicInteger -> #+32-bit U32 #+64-bit U64 -> #+32-bit U32 #+64-bit U64 -> Boolean))
+  (define (atomic-cmp-and-swap atomic old new)
```
stylewarning, line 297: either spell the name out or use "cas"

stylewarning (changes requested):
I don't think this is a type-safe API.

garlic0x1:
I will make a lower-level LispThread type that Thread can be turned into for equality tests and have current-thread/all-threads return that.  Should make it properly typesafe.

PR #1334, opened 2024-12-27
coalton:coalton implicit progn

garlic0x1:
Implicit progn in `coalton:coalton` seems like a good thing to have, we added it to `coalton:lisp` and it will make REPL usage nicer.

Before:
```lisp
COALTON-USER> (coalton 1 2 3)
;;=>
error: Malformed coalton expression
  --> <macroexpansion>:1:11
   |
 1 |  (COALTON 1 2 3)
   |             ^ unexpected form
   [Condition of type COALTON-IMPL/PARSER/BASE:PARSE-ERROR]
```

After:

```lisp
COALTON-USER> (coalton 1 2 3)
3 (2 bits, #x3, #o3, #b11)
```


YarinHeffes (approved):
Looks good to me 👍

PR #1337, opened 2024-12-27
[Draft] Collections API Rework and Feature Enhancements

Jason94:
This draft PR is a start to implementing the proposal [in this discussion](https://github.com/coalton-lang/coalton/discussions/1329). The code mostly matches the API described in the proposal, with small differences as development arose.

Not mentioned in the discussion, but I've also started on a macro-based test suite to verify that all collections conform to the Collections contract.

### Example Code
The following snippet demonstrates usage of the new API. Note the creation of functions generic over `LinearCollection`s, calling the same functions directly on Lists and Vectors, and that the entire collections API is able to be imported directly through coalton-prelude.

```lisp
(coalton-toplevel
  (declare push-end-tens (LinearCollection :m => :m UFix -> :m UFix))
  (define (push-end-tens coll)
    (let result = (cel:new coll))
    (for x in (iter:range-increasing 10 10 91)
      (cel:write! result (push-end x (cel:read result))))
    (cel:read result))
  
  (declare push-end-tens! (MutableLinearCollection :m => :m UFix -> :m UFix))
  (define (push-end-tens! coll)
    (for x in (iter:range-increasing 10 10 91)
      (push-end! x coll))
    coll))

(coalton-toplevel
  (define vec (the (Vector UFix) (new-from 4 (fn (x) x))))
  (define lst (the (List UFix)   (new-from 4 (fn (x) x)))))

(coalton
  (progn
   (push 100 vec)
   (push 100 lst)
   (head vec)
   (head lst)))

(coalton
  (progn
    (traceobject "Push tens onto vec" (push-end-tens vec))      ; -> Push tens onto vec: #(0 1 2 3 10 20 30 40 50 60 70 80 90)
    (traceobject "Push tens onto list" (push-end-tens lst))     ; -> Push tens onto list: (0 1 2 3 10 20 30 40 50 60 70 80 90)
    (traceobject "Original vec" vec)                            ; -> Original vec: #(0 1 2 3)
    (traceobject "Original list" lst)                           ; -> Original list: (0 1 2 3)
    (traceobject "Push! tens onto vec" (push-end-tens! vec))    ; -> Push! tens onto vec: #(0 1 2 3 10 20 30 40 50 60 70 80 90)
    (traceobject "Original vec" vec)))                          ; -> Original vec: #(0 1 2 3 10 20 30 40 50 60 70 80 90)

;; Doesn't compile:
; (coalton
;   (push-end-tens! lst))
```

### Progress

- [x] Collection typeclass
- [ ] KeyedCollection typeclass
- [x] LinearCollection typeclass
- [x] Extend List behavior to satisfy typeclasses
- [x] Extend Vector behavior to satisfy typeclasses
- [ ] Extend OrdMap
- [ ] Extend HashTable
- [ ] Extend Seq
- [ ] Create an immutable Set collection
- [ ] (Partial) Create a collections testing suite
- [ ] (Partial) Create a collections doc page and link to it from the main page

### Open Questions

I'm posting here to get general feedback on the API as I continue, and also to request input on a few specific questions.

**1. Start with List & Vector, or go all in?**

OrdMap and Seq will probably require substantial work to get them to feature parity with the other Collections, since the underlying data structures are much more complicated. (AT least, if done efficiently) HashTable might be easier. List and Vector are basically already done. We could release the Collections API just covering List and Vector initially, and then release PR's to bring the others up to parity afterwards. Or we could wait to release them all at once.

**2. How much effort to put into backwards compatibility?**

The new API is a big departure in behavior, function naming, and packaging. It will certainly break old code. It will probably require broad but safe refactors (mostly renaming functions) for any existing Coalton code. The discussion proposed a deprecation warning scheme. The current PR doesn't go that far, but does at least preserve the old packages as duplicate files in the meantime.


library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    ;; Create new collections
+    (new-collection
+     "Create a new, empty collection."
+     (Unit -> :m))
+    (new-repeat
+     "Create a new collection, attempting to add `elt` `n` times."
+     (UFix -> :a -> :m))
+    (new-from
+     "Create a new collection by appling a function over the range [0, n)."
+     (UFix -> (UFix -> :a) -> :m))
```
YarinHeffes, line 98: This is defined as `tabulate` in #1500 which may be merged soon.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+
+Could be mutable or immutable. Methods are not allowed to modify the
+underlying collection. If you need mutability as part of the contract,
+probably to prevent defensive copying, use one of the Mutable
+collection typeclasses."
+    ;; Create new collections
+    (new-collection
+     "Create a new, empty collection."
+     (Unit -> :m))
+    (new-repeat
```
YarinHeffes, line 93: `new-repeat` can be defined in terms of `new-from`, e.g., `(new-from n (const x))`.

`remove-elt` can be defined in terms of  `filter`.

`contains-elt` can be defined in terms of `contains-where`

We can add a `count` defined in terms of `count-where`.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+     "Create a new collection with all distinct elements."
+     (Eq :a => :m -> :m))
+    (remove-elt
+     "Remove all occurrences of `elt` from the collection."
+     (Eq :a => :a -> :m -> :m))
+    ;; Query the collection
+    (empty?
+     "Check if the collection contains no elements."
+     (:m -> Boolean))
+    (length
```
YarinHeffes, line 116: Perhaps `size` makes more sense here because these are not specifically linear?

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    (count-where
+     "The number of elements satisfying the predicate."
+     ((:a -> Boolean) -> :m -> UFix))
+    ;; Manipulate at the element level
+    (add
+     "Add an element to the collection. For linear collections, should add to
+the front or back, depending on which is natural for the underlying data structure."
+     (:a -> :m -> :m)))
+
+  (define-class (Collection :m :a => ImmutableCollection :m :a)
```
YarinHeffes, line 134: I think that either this class isn't necessary and can be removed, or you might consider moving all of the changing functions from `Collection` into this class, e.g., `filter` and `add` but not `empty?` and `length`.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    "An immutable collection.")
+
+  (define-class (Collection :m :a => MutableCollection :m :a (:m -> :a))
+    "A mutable collection."
+    (copy
+     "Create a shallow copy of the collection."
+     (:m -> :m))
+    (add!
```
YarinHeffes, line 142: Do you plan to add the rest of the functions (e.g., `filter`) in their mutable forms here? I might have missed it if that was mentioned in the description.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    (new-collection
+     "Create a new, empty collection."
+     (Unit -> :m))
+    (new-repeat
+     "Create a new collection, attempting to add `elt` `n` times."
+     (UFix -> :a -> :m))
+    (new-from
+     "Create a new collection by appling a function over the range [0, n)."
+     (UFix -> (UFix -> :a) -> :m))
+    (new-convert
```
YarinHeffes, line 99: Is it possible to define this in terms of the other collections methods?

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+;; (coalton-toplevel
+;;   (define-class (KeyedCollection :m :k :v (:m -> :k :v)))
+;;   )
+
+(coalton-toplevel
+  ;; NOTE: In all cases, LinearCollection methods should return collections that don't
+  ;; share mutable state with the original.
+  (define-class (LinearCollection :m)
+    ;; Get elements from the collection
+    (head
```
YarinHeffes, line 156: These (`head`, `last`, `tail`, `init`) can be defined in terms of their # versions. Actually, if you define `take#` and `drop#`, then you can define all of these methods from `head` through `drop` in terms of `take#` and `drop#`.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+     "Return all except the first element of the collection."
+     (:m :a -> :m :a))
+    (take
+     "Return the first `n` elements of the collection."
+     (UFix -> :m :a -> :m :a))
+    (drop
+     "Return all except the first `n` elements of the collection."
+     (UFix -> :m :a -> :m :a))
+    ;; Query the collection
+    (index-elt
```
YarinHeffes, line 178: More cases here and below where one method can be defined by another.

`index-elt` by `index-elt#`

`indices-elt` by `indices-where`

etc.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+     "Split into collections on each element satisfying `pred`. Elements satisfying `pred` will be excluded from the resulting collections."
+     ((:a -> Boolean) -> :m :a -> List (:m :a)))
+    ;; Manipulate at the collection level
+    (reverse
+     "Return the collection with elements reversed."
+     (:m :a -> :m :a))
+    (sort
+     "Return a sorted collection of orderable elements."
+     (Ord :a => :m :a -> :m :a))
+    (sort-with
```
YarinHeffes, line 219: I think `sort-by` is more common. What do you think?

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+     "Return the collection with an element added to the front."
+     (:a -> :m :a -> :m :a))
+    (push-end
+     "Return the collection with an element added to the end."
+     (:a -> :m :a -> :m :a))
+    (insert-at
+     "Return the collection with an element inserted at an index."
+     (UFix -> :a -> :m :a -> :m :a)))
+
+  (define-class (LinearCollection :m => ImmutableLinearCollection :m))
```
YarinHeffes, line 239: Same comment as with Collections and ImmutableCollections.

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+
+;; TODO: Will tackle KeyedCollection in another PR.
+;; (coalton-toplevel
+;;   (define-class (KeyedCollection :m :k :v (:m -> :k :v)))
+;;   )
+
+(coalton-toplevel
+  ;; NOTE: In all cases, LinearCollection methods should return collections that don't
+  ;; share mutable state with the original.
+  (define-class (LinearCollection :m)
```
YarinHeffes, line 154: Why did you decide to define Collections using a functional dependency but not LinearCollection?

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    (push-end
+     "Return the collection with an element added to the end."
+     (:a -> :m :a -> :m :a))
+    (insert-at
+     "Return the collection with an element inserted at an index."
+     (UFix -> :a -> :m :a -> :m :a)))
+
+  (define-class (LinearCollection :m => ImmutableLinearCollection :m))
+
+  (define-class (LinearCollection :m => MutableLinearCollection :m)
```
YarinHeffes, line 241: If you used fundeps here, then the door is opened for ideas like

`(define-instance (MutableLinearCollection (Cell (List :a)) :a))`

library/collections/classes.lisp
```diff
@@ -0,0 +1,274 @@
...
+    (reverse!
+     "Reverse the collection in place. The collection is returned for convenience."
+     (:m :a -> :m :a))
+    (sort!
+     "Sort a collection of orderable elements in place. The collection is returned for convenience."
+     (Ord :a => :m :a -> :m :a))
+    (sort-with!
+     "Sort the collection in place under the given ordering. The collection is returned for convenience."
+     ((:a -> :a -> Ord) -> :m :a -> :m :a))
+    (push!
```
YarinHeffes, line 251: `push!` and `push-end!` can be defined by `insert-at!` and `length`.

YarinHeffes (commented):
I've only looked at the `classes` so far, but I wanted to share my thoughts, most of which have to do with reducing redundancy among the methods. Let me know what you think :)

library/collections/classes.lisp
Jason94: Yeah that's a good point. If we did that, would we want to define a separate length function that does the exact same thing for LinearCollections? Obviously that would be redundant, but theres a *lot* of code out there that uses "length". This is already a "rip off the bandaid" PR, so I'm open to that.

YarinHeffes: I think it's fine to have `size` in Collections and `length` in LinearCollections.

library/collections/classes.lisp
Jason94: Oh, I see what you're saying - thanks! Yes, I agree. And because this is all going to be invisible to users of the API, if we ever add a new collection that wants a specialization for performance, we can always move it _back_ into the typeclass.

library/collections/classes.lisp
Jason94: I think it's a _fund_amental problem with functional dependencies. We have typeclass:

```lisp
(define-class (Collection :m :a (:m -> :a))
  ...)
```

We could try to define `zip` as:
```lisp
  (zip
    (:m -> :m -> :m (Tuple :a :b)))
```
but the return type won't work, because `:m` isn't `List`, it's `List :a`. The kind of `:m` inside the class definition is `*`, not `* -> *`, so you can never write `:m (Tuple :a :b)`. There's also problems with trying to "unpack" the `:b` type from the second parameter, but that isn't so bad because you can just use an iterator:

```lisp
(zip
  (IntoIterator :n :b => :m -> :n -> :m (Tuple :a :b)))
```.

So the second arg is fixed, but the return type still fails to compile.

Still, now that I'm thinking about your suggestion of moving some of these functions out of the class definitions, it would probably be easy to define the `zip` functions in terms of _other_ functions that can be defined with fundeps. Then it would be possible, to redefine `LinearCollection` with functional dependencies. Actually, I do like that approach - what do you think?

library/collections/classes.lisp
Jason94: Got it. Yeah, this could definitely be implemented in terms of `tabulate` when then gets merged. it's an interesting design decision how the collections library should work with those sorts of typeclases. These classes are sort of supposed to be a user-focused "one stop shop" for all things collections, whereas typeclasses like `Semigroup` are small, atomic building blocks that are more meant for libraries. There are some places where I chose to re-expose some functionality, like how `empty?` is duplicative of `Monoid`. But I didn't re-expose `<>` from `Semigroup` or `map` from `Functor`.

Maybe that was just because I don't love the `mepty` name from Coalton's Monoid class, haha.

library/collections/classes.lisp
YarinHeffes: Well, zip-with calls uses the "with" to generate the elements of the result, like Lisp's `mapcar` (it _is_ `mapcar`), but sort-by uses the function like Lisp's `:key`. It might even make more sense to use `sort-where`. But sort-by is common in other languages. What do you think?

library/collections/classes.lisp
YarinHeffes: I like that approach too – I think we're on the same page on this now.

library/collections/classes.lisp
Jason94: Oh wow, this is really nice. Or even: `(define-instance (MutableLinearCollection :m :a => MutableLinearCollection (Cell :m) :a)`

library/collections/classes.lisp
Jason94: I'm actually just going to take the `zip` functions out. I started working on them, and I was pretty much just wrapping the `FromIterator:zip` function.

Jason94:
@YarinHeffes & @stylewarning I just learned about `specialize` a few days ago. It seems like we could use that to trim down the typeclasses here? There are some functions, like `head` that I wanted to leave in the typeclass to allow it to be overridable for performance reasons. Would it be possible to move it out of the typeclass as a normal function that uses an index-based lookup, and then specialize it for `List`?

Issue #1340, opened 2025-01-02
Typeclass Definition "Forgets" constraints on functionally dependent type parameter


Jason94:
I have two typeclasses with functionally dependent type parameters. The simplest case, with no type constraints compiles:

```lisp
  (define-class (Collection_ :m :a (:m -> :a))
    (new-collection_
     (Unit -> :m))
    (add_
     (:a -> :m -> :m)))
  
  (define-class (EqCollection_ :m :a (:m -> :a))
    (contains-elt?_
     (:a -> :m -> Boolean))))

  (define-instance (Collection_ (List :a) :a)
    (define new-collection_ (const Nil))
    (define add_ Cons))
  
  (define-instance (Eq :a => EqCollection_ (List :a) :a)
    (define (contains-elt?_ elt lst)
      (some? (l:find (== elt) lst))))
```

as does the case where there is a type constraint on the "outer" type parameter:

```lisp
  (define-class ((Collection_ :m :a) => EqCollection_ :m :a (:m -> :a))
    (contains-elt?_
     (:a -> :m -> Boolean))))
```

However, the case where there is constraint on the `:a` itself fails to compile:

```lisp
  (define-class (Eq :a => EqCollection_ :m :a (:m -> :a))
    (contains-elt?_
     (:a -> :m -> Boolean))))
  
  (define-instance (Eq :a => EqCollection_ (List :a) :a)
    (define (contains-elt?_ elt lst)
      (some? (l:find (== elt) lst))))
==>
error: Instance missing context
--> <macroexpansion>:7:19
|
7 | (DEFINE-INSTANCE (EQ :A => EQCOLLECTION_ (LIST :A) :A)
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ No instance for EQ :A
```
Something is happening here where it's not "seeing" that in the instance, `:a` is an `Eq`. I think the problem is with the instance definition, not the typeclass definition, because it correctly doesn't compile _without_ the `Eq` typeclass either:

```lisp
  (define-class (Eq :a => EqCollection_ :m :a (:m -> :a))
    (contains-elt?_
     (:a -> :m -> Boolean))))
  
  (define-instance (EqCollection_ (List :a) :a)
    (define (contains-elt?_ _ _)
      False))
==>
  error: Instance missing context
  --> <macroexpansion>:7:19
   |
[...]

PR #1342, opened 2025-01-03
Fix `define-instance` context bug (1340)

YarinHeffes:
Fixing the following bug:

```lisp
(coalton-toplevel
  (define-class (C :col :elem))
  (define-class (Eq :elem => EqC :col :elem))
  (define-instance (C (List :elem) :elem))
  (define-instance (Eq :elem => EqC (List :elem) :elem)))
```

This block previously would not compile, because the typechecker checked that `:elem` satisfies `Eq :elem` before checking the predicates provided in the `define-instance`. See #1340.


Jason94:
Fixed the issue I was having working on collections, so looks good to me 👍

stylewarning (approved):

PR #1343, opened 2025-01-05
 Allow additional constraints in typeclass methods 

Jason94:
This pull request allows typeclass methods to provide additional constraints on the typeclass variables within individual method definitions.

The motivating use case is developing the collections library, to allow for a hierarchy of collection types so that Dictionary/Map like types are collections of tuples and so that linear collections like Vectors/Lists have the same interface as maps of UFix's.

The following is currently compiling ([on this branch that fixes a bug](https://github.com/coalton-lang/coalton/pull/1342)):
```lisp
  (define-class (Collection :m :a (:m -> :a)))

  (define-class (Collection :m (Tuple :k :v) => KeyedCollection :m :k :v (:m -> :k :v)))

  (define-class (KeyedCollection :m UFix :a => LinearCollection_ :m :a (:m -> :a))))
```
However, some collection functions rely on an `Eq` instance for the contained type. Under Coalton's current limitations on method constraints, supporting these functions requires creating an additional typeclass:
```lisp
  (define-class ((Eq :a) (Collection :m :a) => EqCollection :m :a (:m -> :a)))
```

This PR follows a discussion on the best way to approach this problem in the Discord.

---

The actual change is small. This was essentially already supported, most of what needed to be done was removing checks against this use case.

See the tests in the commit for examples of what will/won't compile.

### DIFF
--- a/src/typechecker/define-class.lisp
+++ b/src/typechecker/define-class.lisp
@@ -391,19 +391,6 @@
                                    (tc-note method
                                             "the method is ambiguous"))
 
-                         ;; Ensure that the type variables in each
-                         ;; pred are not a subset of the class
-                         ;; variables.
-                   :do (loop :for pred :in (parser:qualified-ty-predicates ty)
-                             :for tyvars := (remove-duplicates
-                                             (mapcar #'parser:tyvar-name (parser:collect-type-variables pred))
-                                             :test #'eq)
-
-                             :when (subsetp tyvars var-names)
-                               :do (tc-error "Invalid method predicate"
-                                             (tc-note pred
-                                                      "method predicate contains only class variables")))
-
                    :do (loop :for tyvar :in new-tyvars
                              :do (partial-type-env-add-var env tyvar))
 
--- a/tests/type-inference-tests.lisp
+++ b/tests/type-inference-tests.lisp
@@ -236,7 +236,25 @@
       (define test-addl id))
 
     (declare test-bare-testtype (TestType -> TestType))
-    (define test-bare-testtype test-bare)"))
+    (define test-bare-testtype test-bare)")
+
+  ;; Check that it works with functional dependencies
+  (check-coalton-types
+   "(define-type (Box :a)
+      (Box :a))
+
+    (define-class (ClassA :m :a (:m -> :a))
+      (get-a (:m -> :a)))
+
+    (define-class (ClassB :m :a (:m -> :a))
+      (convert (ClassA :n :a => :n -> :m)))
+
+    (define-instance (ClassA (Box :a) :a)
+      (define (get-a (Box a)) a))
+
+    (define-instance (ClassB (Box :a) :a)
+      (define (convert bx)
+        (Box (get-a bx))))"))
 
 (deftest test-typeclass-flexible-instances ()
   (check-coalton-types
--- a/tests/type-inference-tests.lisp
+++ b/tests/type-inference-tests.lisp
@@ -192,6 +192,51 @@
    "(define-class (Test :a)
       (test (Eq :b => :a -> :b)))"))
 
+(deftest test-typeclass-additional-constraints ()
+
+  ;; Check that typeclass methods can provide additional constraints
+  (check-coalton-types
+   "(define-class (Test :a)
+      (test-bare (:a -> :a))
+      (test-addl (Eq :a => :a -> :a)))
+
+    ;; Can define on types without Eq
+    (define-type TestType A B)
+
+    (define-instance (Test TestType)
+      (define test-bare id)
+      (define test-addl id))
+
+    ;; Can define generic function that uses Eq if constrained
+    (declare test-generic ((Eq :a) (Test :a) => :a -> :a))
+    (define test-generic test-addl)
+
+    ;; Can use test-addl on TestType if Eq defined
+    (define-instance (Eq TestType)
+      (define (== x y)
+        (match (Tuple x y)
+          ((Tuple (A) (A)) True)
+          ((Tuple (B) (B)) True)
+          (_ False))))
+
+    (declare test-addl-testtype (TestType -> TestType))
+    (define test-addl-testtype test-addl)")
+
+  ;; Check that methods not requiring additional constraints can be used
+  (check-coalton-types
+   "(define-class (Test :a)
+      (test-bare (:a -> :a))
+      (test-addl (Eq :a => :a -> :a)))
+
+    ;; Can define on types without Eq
+    (define-type TestType A B)
+
+    (define-instance (Test TestType)
+      (define test-bare id)
+      (define test-addl id))
+
+    (declare test-bare-testtype (TestType -> TestType))
+    (define test-bare-testtype test-bare)"))
 
 (deftest test-typeclass-flexible-instances ()
   (check-coalton-types
--- a/src/codegen/translate-instance.lisp
+++ b/src/codegen/translate-instance.lisp
@@ -46,6 +46,10 @@
            (loop :for pred :in (tc:toplevel-define-instance-context instance)
                  :collect (pred-type pred env)))
 
+         (subs (tc:predicate-match (tc:ty-class-predicate class) pred))
+
+         (superclass-preds (tc:apply-substitution subs (mapcar #'car (tc:ty-class-superclass-dict class))))
+
          (method-definitions
            (loop :for method :in (tc:ty-class-unqualified-methods class)
                  :for method-name := (tc:ty-class-method-name method)
@@ -57,7 +61,7 @@
          (unqualified-method-definitions
            (loop :for method :in (tc:ty-class-unqualified-methods class)
                  :for method-name := (tc:ty-class-method-name method)
-                 :for method-type := (tc:ty-class-method-type method)
+                 :for method-type := (tc:apply-substitution subs (tc:ty-class-method-type method))
                  :for binding := (gethash method-name (tc:toplevel-define-instance-methods instance))
                  :collect (translate-toplevel (tc:attach-explicit-binding-type binding (tc:fresh-inst method-type))
                                               env
@@ -65,10 +69,6 @@
                                               :extra-context ctx)))
 
          (method-ty (mapcar #'node-type unqualified-method-definitions))
-
-         (subs (tc:predicate-match (tc:ty-class-predicate class) pred))
-
-         (superclass-preds (tc:apply-substitution subs (mapcar #'car (tc:ty-class-superclass-dict class))))
 
          (superclass-ty
            (loop :for pred :in superclass-preds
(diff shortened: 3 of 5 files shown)


stylewarning (approved):

stylewarning:
@Jason94 Can you investigate the CCL errors? Allegro errors can be ignored.

Jason94:
> @Jason94 Can you investigate the CCL errors? Allegro errors can be ignored.

CCL & SBCL tests passing now :+1:

stylewarning:
@eliaslfox Do you see any issue removing this extra check?

src/typechecker/define-class.lisp
```diff
@@ -402,19 +402,6 @@
                                    (tc-note method
                                             "the method is ambiguous"))
 
-                         ;; Ensure that the type variables in each
```
Jason94, line 405: Found a regression. Previously this was typechecking, and now it's failing to unify. Fix and add a test case.

```lisp
  (define-class (Collection :m :a (:m -> :a))
    (new-convert
     "Convert a collection of another type. If converting a LinearCollection to a LinearCollection, should preserve order."
     ((Collection :f :a) (itr:IntoIterator :f :a) => :f -> :m)))

    (define-instance (Collection (List :a) :a)
      (define (new-convert coll)
        (itr:collect! (itr:into-iter coll))))
```

stylewarning (changes requested):
changing my review status until a little more testing is done

src/typechecker/define-class.lisp
Jason94: This seems to be the minimal code needed to reproduce it the error. I'm not sure if it is because of the functional dependencies or having multiple typeclass parameters:

Compiles:
```lisp
  (define-type TestType X Y)

  (define-class (A :a)
    (id-a (:a -> :a)))

  (define-class (B :b)
    (a-to-b (A :a => :a -> :b -> :b)))

  (define-instance (A TestType)
    (define id-a id))

  (define-instance (B TestType)
    (define (a-to-b _ b) b))
```
Doesn't compile:
```lisp
  (define-type (Box :a)
    (Box :a))

  (define-class (ClassA :m :a (:m -> :a))
    (get-a (:m -> :a)))

  (define-class (ClassB :m :a (:m -> :a))
    (convert (ClassA :n :a => :n -> :m)))

  (define-instance (ClassA (Box :a) :a)
    (define (get-a (Box a)) a))

  (define-instance (ClassB (Box :a) :a)
    (define (convert bx)
      (Box (get-a bx))))
==>
Compile-time error:
  (DEFINE-TYPE #
    ...)
  ...))
Unknown instance for predicate CLASSA #T23431 #T23442
```

stylewarning:
@Jason94 I just rebased this; sorry I should have asked first. Is this PR ready for (re-)review?

stylewarning:
@Jason94 I see some additional work has happened. Is this ready for re-review?

src/typechecker/define-class.lisp
Jason94: @stylewarning I added a failing test with the above, so it's now properly failing. As I said, I'll have some more time to take try to figure this out over the next few weeks. But I hit a wall last time, so if you have any thoughts on where to go from here, I'd appreciate it.

The code that I changed is in the typechecker, but the code that is causing the problem is all the way over in the codegen portion of the compiler.

Here is a snippet that's more analogous to the fundep version, which still compiles:
```lisp
(coalton-toplevel
  (define-type MyType
    A B)

  (define-class (Incable :a)
    ;; (inc (:a -> :a))
    (is-inced (Eq :a => :a -> Boolean))))

(coalton-toplevel
  (define-instance (Eq MyType)
    (define (== a b)
      (match (Tuple a b)
        ((Tuple (A) (A)) True)
        ((Tuple (B) (B)) True)
        (_ False)))))

(coalton-toplevel
  (define-instance (Incable MyType)
    (define (is-inced myt)
      (== myt myt))))
```
Here, `typechecker/environment:lookup-class-instance` gets called with a class over the concrete `MYTYPE`:
```
  18 |           |--- coalton-impl/typechecker/unify:predicate-match
     |           |    > EQ #T130
     |           |    > EQ #T23524
     |           |    < (#T130 +-> #T23524 #T130 +-> #T23524)
  19 |           `--- coalton-impl/typechecker/environment:lookup-class-instance
     |              | > #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {1008ECAB53}>
     |              | > EQ MYTYPE
```
But, when I try to compile the snippet from the above comment with the fundeps that doesn't compile, somehow the `unify:predicate-match` function is calling `lookup-class-instance` with a class over a TVAR, instead of a concrete type:
```
  20                |--- coalton-impl/typechecker/unify:predicate-match
                    |    > CLASSA #T23077
                    |    > CLASSA #T23081
                    |    < (#T23077 +-> #T23081 #T23077 +-> #T23081)
  21                `--- coalton-impl/typechecker/environment:lookup-class-instance
                       | > #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {10086DCCA3}>
                       | > CLASSA #T23567
```.

So that seems to be the problem, as far as I can tell. Like I said, I'll keep digging, but if you can think of some things that could be going on, that would help! I'm happy to share more from the copious notes of stacktraces that I saved in January, if it'd help.

Comment (stylewarning) [reply src/typechecker/define-class.lisp:394]: cc @YarinHeffes

Comment (Jason94) [reply src/typechecker/define-class.lisp:394]: Looking into it some more, `translate-instance:translate-instance` produces a qualified and unqualified version of the method definition:
```lisp
         (method-definitions
                 [...]
                 :collect (cons codegen-sym (translate-toplevel binding env method-name))))

         (unqualified-method-definitions
           (loop :for method :in (tc:ty-class-unqualified-methods class)
                 [...]
                 :collect (translate-toplevel (tc:attach-explicit-binding-type binding (tc:fresh-inst method-type))
                                              env
                                              method-name
                                              :extra-context ctx)))
```

the `attach-explicit-binding-type` attaches the new TVAR created by `tc:fresh-inst` to the _name_ of the method, but not the _body_:
```lisp
(defgeneric attach-explicit-binding-type (binding explicit-type)
  [...]

  (:method ((binding instance-method-definition) explicit-type)
    [...]
    (make-instance-method-definition
     :name (make-node-variable
            :name (node-variable-name (instance-method-definition-name binding))
            :type explicit-type
            :location (source:location (instance-method-definition-name binding)))
     :params (instance-method-definition-params binding)
     :body (instance-method-definition-body binding)
     :location (source:location binding))))
```
So when it tries to translate the expression, it creates the predicate context by looking in the name but not the body, which creates a mismatch, so it isn't able to figure it out:
```lisp
   2 - coalton-impl/codegen/translate-expression:translate-expression
     | > #S(COALTON-IMPL/TYPECHECKER/EXPRESSION:NODE-BODY :NODES COMMON-LISP:NIL :LAST-NODE #S(COALTON-IMPL/TYPECHECKER/EXPRESSION:NODE-VARIABLE :TYPE EQ #T28883 ⇒ ((LIST #T28883) → (LIST #T28883)) :LOCATION #S(COALTON-IMPL/SOURCE:LOCATION :SOURCE #<COALTON-IMPL/SOURCE::SOURCE-STRING {100C4F0543}> :SPAN (95 . 112)) :NAME REMOVE-DUPLICATES))
     | > ((EQ #T28660 . #:DICT4902))

```
Note how in the context list, we have `EQ #T28660` but in the actual function we now have `EQ #T28883 ⇒ ((LIST #T28883) → (LIST #T28883)`.

This leads to the eventual error:
```lisp
Unknown instance for predicate EQ #T28883

Backtrace:
 0: (COALTON-IMPL/TYPECHECKER/ENVIRONMENT:LOOKUP-CLASS-INSTANCE #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {100C4EF413}> EQ #T28883 :NO-ERROR COMMON-LISP:NIL)
 1: (COALTON-IMPL/CODEGEN/RESOLVE-INSTANCE:RESOLVE-STATIC-DICT EQ #T28883 ((EQ #T28660 . #:DICT4902)) #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {100C4EF413}>)
 2: (COALTON-IMPL/CODEGEN/RESOLVE-INSTANCE:RESOLVE-DICT EQ #T28883 ((EQ #T28660 . #:DICT4902)) #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {100C4EF413}>)
 3: (COALTON-IMPL/CODEGEN/TRANSLATE-EXPRESSION::APPLY-DICTS #S(COALTON-IMPL/TYPECHECKER/EXPRESSION:NODE-VARIABLE :TYPE EQ #T28883 ⇒ ((LIST #T28883) → (LIST #T28883)) :LOCATION #S(COALTON-IMPL/SOURCE:LOCAT..
 4: ((:METHOD COALTON-IMPL/CODEGEN/TRANSLATE-EXPRESSION:TRANSLATE-EXPRESSION (COALTON-IMPL/TYPECHECKER/EXPRESSION:NODE-VARIABLE COMMON-LISP:T COMMON-LISP:T)) #S(COALTON-IMPL/TYPECHECKER/EXPRESSION:NODE-VA..
```

Jason94:
> @Jason94 I see some additional work has happened. Is this ready for re-review?

Finally ready for re-review!

tests/type-inference-tests.lisp
```diff
@@ -192,6 +192,69 @@
...
+  ;; Check that it works with functional dependencies
+  (check-coalton-types
+   "(define-type (Box :a)
+      (Box :a))
+
+    (define-class (ClassA :m :a (:m -> :a))
+      (get-a (:m -> :a)))
+
+    (define-class (ClassB :m :a (:m -> :a))
+      (convert (ClassA :n :a => :n -> :m)))
```
YarinHeffes, line 250: This is a good test to have, but I just wanted to make a comment that this example works without this PR, because the set of variables `{ :n :a }` is not a subset of the class variables `{ :m :a }`.

YarinHeffes (approved):
I spent some time thinking about the merits of keeping things the way they are vs. merging this PR, and ultimately this PR lets the programmer be more expressive and I support it.

I think defining a separate type class like `EqCollection` is fine, but I don't think there's any reason not to allow the functionality of this PR for the programmer to express a method of `Collection` like `contains?`/`member` instead.

👍

stylewarning (approved):
It has not been noted yet that this change is effectively blessed by [_Type classes: an exploration of the design space_ by Jones, Jones, & Meijer](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=11998e37d8dc39a7207e91dd12699ec0667cae89). In Section 4.10, they pose the decision "what limitations, if any, are there on contexts in class-member type signatures?" Their answer is pretty straightforward:

> In Haskell, [a constraint on a class-member is] illegal, because it further constrains the class type variable [of the class]. There seems to be no technical reason for this restriction. It is simply a nuisance to the Haskell specification, implementation, and (occasionally) programmer.

So it seems this design change is harmless and ordained as "A-OK". 

We should consider other aspects from that paper.

PR #1344, opened 2025-01-07
Persistent hash tree

jbouwman:
Define a persistent hash tree for managing environment definitions.

This could potentially underlie some of Coalton API collections defined in #1337


stylewarning:
Closing (but not deleting) due to being stale.

PR #1352, opened 2025-01-10
Removed unused Num function.

YarinHeffes:
Removed unused function `handle-unsigned-overflow` in `num.lisp`.


jbouwman (approved):
👍🏽 fewer innocuous warnings

PR #1359, opened 2025-01-16
Removed suffix from float string representations.

YarinHeffes:
Removed `d0` suffix from appearing in string representations of double floats, and updated single-float casting for consistency.


stylewarning:
Why?

YarinHeffes:
When I originally wrote this, I was using `(Into :T String)` for printing. I suppose this is superseded by a more specialized way to format strings, in which case I can see how it would be preferable to include the `d0` suffix.

PR #1360, opened 2025-01-16
Removed re-evaluation of last coalton form

YarinHeffes:
Previously, the last form in a `coalton` expression with more than one form would be evaluated twice. This PR fixes that.

### DIFF
--- a/src/parser/expression.lisp
+++ b/src/parser/expression.lisp
@@ -1047,7 +1047,7 @@ Rebound to NIL parsing an anonymous FN.")
   (let ((nodes (mapcar (lambda (form) (parse-expression form source)) forms)))
     (make-node-progn
      :body (make-node-body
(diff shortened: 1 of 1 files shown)


src/parser/expression.lisp
```diff
@@ -1047,7 +1047,7 @@ Rebound to NIL parsing an anonymous FN.")
   (let ((nodes (mapcar (lambda (form) (parse-expression form source)) forms)))
     (make-node-progn
      :body (make-node-body
-            :nodes nodes
+            :nodes (nreverse (rest (reverse nodes)))
```
stylewarning, line 1050: use `butlast`

stylewarning (changes requested):

stylewarning (approved):

PR #1363, opened 2025-01-17
Rename bound variables when inlining

YarinHeffes:
This is a continuation of #1295. See also #1293.

This PR fixes the CI test failure in #1295 by separating the bound variable substitutions from the call-site-given AST substitutions and treating them separately.


stylewarning (approved):

amorphedstar:
For correctness in more cases, we probably want to add stuff into new-subs whenever a variable is bound (not just in a let), like I think in bind and while-let and maybe match branches and something else.

PR #1372, opened 2025-02-10
Add explicit inline attributes

YarinHeffes:
Adding explicit `(inline)` attributes to standard library.

- [ ] big-float/*
- [ ] computable-reals/*
- [x] math/*
- [ ] monad/*
- [ ] bits
- [ ] boolean
- [x] builtin
- [ ] cell
- [ ] char
- [ ] classes
- [ ] file
- [x] functions
- [x] hash
- [ ] hashtable
- [ ] iterator
- [ ] lisparray
- [ ] list
- [ ] optional
- [ ] ord-map
- [ ] ord-tree
- [ ] primitive-types
- [ ] queue
- [ ] random-access
- [ ] result
- [ ] seq
- [ ] slice
- [ ] string
- [ ] system
- [ ] tuple
- [ ] types
- [ ] vector


stylewarning:
I don't think every single function/file should have inlines. For example, big float arithmetic almost surely doesn't.

PR #1375, opened 2025-02-10
Add inlining for boolean functions

YarinHeffes:
Adds `(inline)` attributes in `library/builtin.lisp` and `library/boolean.lisp`.
Also adds type declarations for `not` and `xor`.

### DIFF
--- a/library/boolean.lisp
+++ b/library/boolean.lisp
@@ -25,7 +25,6 @@
         (cl:eq x y))))
 
   (define-instance (Ord Boolean)
-    (inline)
     (define (<=> x y)
       (match x
         ((True)
--- a/library/boolean.lisp
+++ b/library/boolean.lisp
@@ -19,11 +19,13 @@
   ;;
 
   (define-instance (Eq Boolean)
+    (inline)
     (define (== x y)
-        (lisp Boolean (x y)
-              (cl:eq x y))))
+      (lisp Boolean (x y)
+        (cl:eq x y))))
 
   (define-instance (Ord Boolean)
+    (inline)
     (define (<=> x y)
       (match x
         ((True)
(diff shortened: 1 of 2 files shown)


library/boolean.lisp
```diff
@@ -19,11 +19,13 @@
...
   (define-instance (Eq Boolean)
+    (inline)
     (define (== x y)
-        (lisp Boolean (x y)
-              (cl:eq x y))))
+      (lisp Boolean (x y)
+        (cl:eq x y))))
 
   (define-instance (Ord Boolean)
+    (inline)
```
stylewarning, line 28: don't inline

library/builtin.lisp
```diff
@@ -26,39 +26,48 @@
      (cl:error ,datum ,@arguments)))
 
 (coalton-toplevel
+  (inline)
```
stylewarning, line 29: don't inline for debugging

library/builtin.lisp
```diff
@@ -26,39 +26,48 @@
...
-  (define xor
+  (inline)
+  (declare xor (Boolean -> Boolean -> Boolean))
+  (define (xor x y)
     "Synonym for `boolean-xor`."
-    boolean-xor)
+    (boolean-xor x y))
 
+  (inline)
   (declare boolean-not (Boolean -> Boolean))
```
stylewarning, line 47: these functions and below may be better as being implemented with LISP forms since booleans are normal Lisp booleans and Lisp will optimize better 

for example, cl:not instead of pattern matching

cl:and and cl:or for their respective boolean-* functions

stylewarning (changes requested):

library/builtin.lisp
YarinHeffes: https://github.com/coalton-lang/coalton/pull/1375/commits/2b8a076ea3252795c260b74b0fdfe3fb8d7b738b

PR #1376, opened 2025-02-10
Add inline attributes for Cell

YarinHeffes:
Adds `(inline)` attributes for `library/cell.lisp`.


stylewarning (changes requested):
this PR should be limited to just `Cell` in my opinion

stylewarning (approved):

PR #1377, opened 2025-02-10
Add inline attributes for Vector and Slice

YarinHeffes:
Adds inlining to `library/vector.lisp` and `library/slice.lisp`.


stylewarning:
@YarinHeffes Can we break `LispArray` into one PR and `RandomAccess` (for everything but `LispArray`) into another PR?

stylewarning (approved):

PR #1392, opened 2025-02-19
Add inline attributes for library/functions.lisp

YarinHeffes:
Adds inline attributes to simple functions.

Question: should `id` and `const` be redefined to return `fn`s,  like `compose` already does? This would break any code that uses these functions as fully applied like `(id x)`, but I can't think of a case where that arises in practice anyway.

### DIFF
--- a/library/functions.lisp
+++ b/library/functions.lisp
@@ -97,17 +97,11 @@
     "The same as `fold` but with the argument order swapped to match `cl:reduce`"
     (fold (flip f) y xs))
 
-
-  ;; We don't write (COMPOSE F G X) even though it's OK so that the
-  ;; most common case of using compose---as a binary function---is
-  ;; considered to be "saturated".
   (inline)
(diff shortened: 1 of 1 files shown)


library/functions.lisp
```diff
@@ -97,6 +101,7 @@
   ;; We don't write (COMPOSE F G X) even though it's OK so that the
   ;; most common case of using compose---as a binary function---is
   ;; considered to be "saturated".
+  (inline)
   (declare compose ((:b -> :c) -> (:a -> :b) -> (:a -> :c)))
   (define (compose f g)
```
stylewarning, line 106: Let's change this to f g x

stylewarning (approved):

library/functions.lisp
YarinHeffes: https://github.com/coalton-lang/coalton/pull/1392/commits/41b11199d35c7622aea7ba1232a529ee87612d9a

stylewarning (approved):

PR #1397, opened 2025-03-04
Fix code generation to optimize mutual tail-calls.

### LINKED ISSUE #1341: Coalton fails to eliminate mutual tail calls
SBCL is able to eliminate mutual tail calls in the classic worlds-worst even?/odd? algorithm:

```lisp
(in-package :cl-user)

(declaim (optimize (speed 3) (safety 0) (debug 0)))

(defparameter *n-small* 100)
(defparameter *n-big* 100000000)

(defun odd? (n)
  (declare (fixnum n))
  (if (zerop n)
      nil
      (even? (1- n))))

(defun even? (n)
  (declare (fixnum n))
  (if (zerop n)
      t
      (odd? (1- n))))

(even? *n-big*) ;; ==> Returns T and does not blow the stack
```
However, the same program in Coalton seems to not eliminate the tail calls and quickly leads to a stack overflow:
```lisp
(cl:in-package :cl-user)
(defpackage :stack-test
  (:use
   #:coalton
   #:coalton-prelude))
(in-package :stack-test)

(cl:declaim #.coalton-impl/settings:*coalton-optimize*)
(cl:declaim (cl:optimize (cl:speed 3) (cl:safety 0) (cl:debug 0)))

(coalton-toplevel
  (define n-big 100000000)

  (declare isodd? (Integer -> Boolean))
  (define (isodd? n)
    (if (== n 0)
        False
        (iseven? (1- n))))

  (declare iseven? (Integer -> Boolean))
  (define (iseven? n)
    (if (== n 0)
        True
        (isodd? (1- n)))))

(coalton (isodd? n-big))
```
throwing:
```
Control stack exhausted (no more space for function call frames).
This is probably due to heavily nested or infinitely recursive function
calls, or a tail call that SBCL cannot or has not optimized away.

PROCEED WITH CAUTION.
   [Condition of type SB-KERNEL::CONTROL-STACK-EXHAUSTED]

Restarts:
 0: [*ABORT] Return to SLY's top level.
 1: [ABORT] abort thread (#<THREAD tid=13032 "slynk-worker" RUNNING {11044843A3}>)

Backtrace:
 0: (ISEVEN? #<unavailable argument>)
 1: (ISODD? #<unavailable argument>)
 2: (ISEVEN? #<unavailable argument>)
 3: (ISODD? #<unavailable argument>)
 4: (ISEVEN? #<unavailable argument>)
```
Verifying that I did have release mode on:
```lisp
CL-USER> (uiop:featurep ':coalton-release)
T
```
Probably not related to the stack overflow, but I did get these 45 compiler messages when compiling the Coalton program. I was surprised, because I thought that the Integer type annotation would be enough to give SBCL the equivalent fixnum declaration:
```
; note: unable to
;   open-code FLOAT to RATIONAL comparison
; due to type uncertainty:
;   The first argument is a COMMON-LISP:NUMBER, not a (COMMON-LISP:COMPLEX
;                                                      COMMON-LISP:SINGLE-FLOAT).
; 
; note: unable to
;   open-code FLOAT to RATIONAL comparison
; due to type uncertainty:
;   The first argument is a COMMON-LISP:NUMBER, not a (COMMON-LISP:COMPLEX
;                                                      COMMON-LISP:DOUBLE-FLOAT).
; 
; note: unable to open code because: The operands might not be the same type.
; 
; compilation unit finished
;   printed 45 notes
```
Testing the CL version without the type declarations, I got similar notes (although fewer of them), but it still did not overflow the stack.

fukamachi:
SBCL eliminates mutual tail-calls only when all type declarations appear before function definitions.
However, Coalton currently generates type declarations within each function body.

This commit changes to place type declarations before function definitions.

fix #1341


src/codegen/program.lisp
```diff
@@ -209,7 +208,19 @@ Example:
   (append
    ;; Predeclare symbol macros
    (loop :for (name . node) :in bindings
-         :collect `(global-lexical:define-global-lexical ,name ,(tc:lisp-type (node-type node) env)))
+         :append
+         `((global-lexical:define-global-lexical ,name ,(tc:lisp-type (node-type node) env))
+           ,@(when (and (node-abstraction-p node)
```
stylewarning, line 213: Is there a reason to not put this logic below, just above the `compile-function` call?

Also, please add a comment above the declaration saying why we do an external declaim over an internal declare.

src/codegen/program.lisp
```diff
@@ -199,7 +198,7 @@ Example:
            (type node-abstraction node)
            (type tc:environment env))
   `(defun ,name ,(node-abstraction-vars node)
-     ,(function-declarations node env)
+     (declare (ignorable ,@(node-abstraction-vars node)))
```
stylewarning, line 201: We have a Coalton rule where if the name starts with `_` then it is an ignored variable.

Was the old behavior to emit `ignorable` for all variables?

stylewarning (changes requested):
I am "requesting changes" but one of my requests may not be correct.

src/codegen/program.lisp
fukamachi: This is to make sure that all `declaim`s appear before function definitions.
It is the point of this change to optimize tail-calls with SBCL.

For example, the following code won't be optimized and will raise call stack exhausted with a large number.

```common-lisp
(declaim isodd? ...)
(defun isodd? (n)
  ;; Calls iseven?
  )

(declaim iseven? ...)
(defun iseven? (n)
  ;; Calls isodd?
  )
```

However, it works with this.

```common-lisp
;; Definitions first
(declaim isodd? ...)
(declaim iseven? ...)

(defun isodd? (n)
  ;; Calls iseven?
  )
(defun iseven? (n)
  ;; Calls isodd?
  )
```

I'll add a comment.

src/codegen/program.lisp
fukamachi: From the result of `coalton-codegen`, yes.

src/codegen/program.lisp
fukamachi: I added a comment.
I moved the code into an individual loop rather than sharing with symbol macros.

src/codegen/program.lisp
```diff
@@ -211,6 +210,22 @@ Example:
    (loop :for (name . node) :in bindings
          :collect `(global-lexical:define-global-lexical ,name ,(tc:lisp-type (node-type node) env)))
 
+   ;; Function declarations
+   ;; They must appear before function definitions to optimize mutual tail-calls.
```
stylewarning, line 214: "... in SBCL."

stylewarning (approved):

PR #1401, opened 2025-03-12
Add `rec` (named let)

fukamachi:
Added "named let", which is commonly used in the Scheme world.

```common-lisp
(rec fact ((n 5))
  (if (== n 1)
      1
      (* n (fact (1- n)))))
```

This is equivalent to:

```common-lisp
(let ((n 5)
      (fact (fn (n)
              (if (== n 1)
                  1
                  (* n (fact (1- n)))))))
  (fact n))
```

ref. https://docs.racket-lang.org/guide/let.html#%28part._.Named_let%29

## TODO

- [x] Add `rec`
- [x] Add tests for normal cases
- [x] Add tests for error cases


shirok:
The 'init' expressions needs to be evaluated in the outer scope of named let.   The expansion of linked Racket page is subtly incorrect, unless they changed the semantics from Scheme.

```
;; Scheme
(let ((f 3))
  (let f ((n f)) n))
 => 3
```

This is because `f` as the initial value of `n` must refer to the outer `f`.

Using `fact` example above, the correct Coalton expansion should be:

```
((let ((fact (fn (n) ...)))
   fact)
 5)
```

Check out the definition of `let` in https://standards.scheme.org/corrected-r7rs/r7rs-Z-H-9.html#TAG:__tex2page_sec_7.3

shirok:
Considering that Coalton's `let` is Scheme's `letrec`, however, it may be an option to change the semantics and evaluate init exprs within the named let scope.

The reason it's natural in Scheme is that you can say `(let ((f 3)) (let ((f f)) f))`.  The init expressions of `let` is evaluated in the outer scope.

fukamachi:
Good catch. Apparently, I tend to forget Coalton is Lisp-1.

There might be a way to make it suitable for Coalton programmers, but I feel the example `(let ((f 3)) (let f ((n f)) n))` should return 3.

stylewarning:
@fukamachi Can you add tests for malformed syntax? I want to make sure that syntax errors in the name, or bindings, etc. are properly printed. Example malformed syntaxes:

```
(let 1 ((x 0)) x) ; bad name

(let f ((1 0)) 0) ; bad binding name

(let f (hi) 0) ; bad binding
```

stuff like this.

Another issue I see is that we can't declare the type of `f` unfortunately. Not sure if we should fix it in this PR though, since it would take a lot more heavy lifting.

YarinHeffes:
> Another issue I see is that we can't declare the type of `f` unfortunately. Not sure if we should fix it in this PR though, since it would take a lot more heavy lifting.

Just thinking out loud... we could allow
```lisp
(let f (Integer -> Integer)
    ((a ...)
     (b ...))
  (do-something-with a b))
```
to expand to
```lisp
(let ((a ...)
      (b ...))
  (let ((declare f (Integer -> Integer))
        (f (fn (a b) (do-something-with a b))))
    (f a b)))
```
which doesn't _seem_ on the face of it to add much work to this PR, the added work I imagine to be primarily more unit tests.

stylewarning:
@YarinHeffes that's a good idea. Maybe we hybridize a couple ideas. 

my only grievance now is calling this whole thing "let". How about "rec"?

```
(rec [<type>] <name>
     ([<binding>...])
  <body>...)
```

This would be similar to THE or LISP forms in that the type comes first. Except here it's optional. 
p
I think the pros and cons are somewhat obvious.

@shirok @fukamachi @YarinHeffes

shirok:
> my only grievance now is calling this whole thing "let". How about "rec"?

I'm for giving it a separate name other than `let`.  Overloading `let` has confused newcomers to Scheme.  Besides, Coaton's `let` is Scheme's `letrec`, so that's another reason not to overload the name, regarding different scope rules.

Btw, Scheme already has `rec` (https://srfi.schemers.org/srfi-31/srfi-31.html) but with different semantics.

eitaroai:
Although I don't come up with a good name, I agree with not overusing `let`.
`rec` sounds good to me as it seems it's introduced with the same motivation.

fukamachi:
Although I don't come up with a good name, I agree with not overusing `let`.
`rec` sounds good to me as it seems it's introduced with the same motivation.

examples/thih/src/thih.lisp
```diff
@@ -528,15 +528,15 @@
...
-    (let ((rec (fn (rs xs)
-                 (match xs
-                   ((Nil)
-                    rs)
-                   ((Cons p ps)
-                    (if (entail ce (append rs ps) p)
-                        (rec rs ps)
-                        (rec (Cons p rs) ps)))))))
-      (rec Nil xs)))
+    (rec f ((rs Nil)
```
YarinHeffes, line 531: Now that I see this change here, I am wondering: should we be concerned about deprecating old code that uses `rec` as a variable name? Personally, I am not as concerned about the alternative of overloading `let`, but I understand the merits of differentiating `named-let` with a new name anyway. I also wonder if explicitly using `named-let` instead of `let` or `rec` is better, if it's not too verbose.

stylewarning: I think it's too verbose and idiosyncratic to Scheme.

src/parser/expression.lisp
```diff
@@ -224,6 +225,8 @@ Rebound to NIL parsing an anonymous FN.")
 ;;;;
 ;;;; node-let := "(" "let" "(" (node-let-binding | node-let-declare)+ ")" body ")"
 ;;;;
+;;;; node-rec := "(" "rec" [qualified-ty] identifier "(" (node-let-binding | node-let-declare)+ ")" body ")"
```
YarinHeffes, line 228: Nit, since we are not defining a node `node-rec`.
```suggestion
;;;; node-let := "(" "let" "(" (node-let-binding | node-let-declare)+ ")" body ")"
;;;;           | "(" "rec" [qualified-ty] identifier "(" (node-let-binding | node-let-declare)+ ")" body ")"
```

stylewarning: (side note: if the error messages are poor we may need to make a node-rec)

fukamachi: I applied it. I'm not sure if it's okay that it says `let` in error messages like "Malformed let binding", though.

src/parser/expression.lisp
```diff
@@ -223,6 +224,7 @@ Rebound to NIL parsing an anonymous FN.")
 ;;;; node-let-declare := "(" "declare" identifier qualified-ty ")"
 ;;;;
 ;;;; node-let := "(" "let" "(" (node-let-binding | node-let-declare)+ ")" body ")"
+;;;;           | "(" "rec" [qualified-ty] identifier "(" (node-let-binding | node-let-declare)+ ")" body ")"
```
stylewarning, line 227: this should be a separate production in the grammar

tests/test-files/parse-rec.txt
```diff
@@ -0,0 +1,158 @@
...
+================================================================================
+
+(package coalton-unit-tests)
+
+(define x
+  (rec f (hi) 0))
+
+--------------------------------------------------------------------------------
+
+error: Malformed let binding
```
stylewarning, line 120: malformed "rec" binding?

tests/test-files/parse-rec.txt
```diff
@@ -0,0 +1,158 @@
...
+(define x
+  (rec f ((hi)) 0))
+
+--------------------------------------------------------------------------------
+
+error: Malformed let binding
+  --> test:4:13
+   |
+ 4 |    (rec f ((hi)) 0))
+   |               ^ let bindings must have a value
```
stylewarning, line 141: "rec bindings"?

stylewarning (commented):
some initial comments on wording and comments

stylewarning:
We also need to add something to the getting started document.

src/parser/expression.lisp
YarinHeffes: For context, it was my suggestiong to change this because we are making a `node-let` and don't actually have a `node-rec`.

src/parser/expression.lisp
```diff
@@ -667,6 +669,112 @@ Rebound to NIL parsing an anonymous FN.")
...
+          (eq 'coalton:rec (cst:raw (cst:first form))))
+
+     ;; (rec)
+     (unless (cst:consp (cst:rest form))
+       (parse-error "Malformed rec"
+                    (note-end source (cst:first form)
+                              "expected function name")))
+
+     (multiple-value-bind (type name rest)
+         (cond
```
YarinHeffes, line 682: This `cond` results in a bad error message for a case like this, where the programmer writes a non-function type.

```lisp
(rec UFix f ((...
```
Consider something like this: 
```suggestion
         (case (length (cst:rest form))
           ((0 1 2) ... parse-error ...)
           ((3) ... parse function name ...)
           ((4) ... parse type and function name ...)
           (otherwise ... parse-error ...))
```

src/parser/expression.lisp
YarinHeffes: For context, it was my suggestion to combine what @fukamachi originally wrote as 
```lisp
;;;; node-rec := ...
```
into the grammar for `node-let` because we are constructing a `node-let` with `make-node-let`. Of course, this can be reverted.

YarinHeffes (changes requested):
In addition my one change requested below, we will also need an entry for `rec` here:
https://github.com/coalton-lang/coalton/blob/ebf096e29254d14c26c862a2ac723be65c0c50fe/src/faux-macros.lisp#L80-L89 
Otherwise, this PR looks good to me. My opinion is that documentation [here](https://github.com/coalton-lang/coalton/blob/main/docs/intro-to-coalton.md) can be added in a separate PR.

src/parser/expression.lisp
stylewarning: What if it's a type alias @YarinHeffes ?

YarinHeffes: That's exactly the sort of case that I meant to address by my comment. Of course, I'm realizing that the `case` is not the right solution, since the body is not restricted to a single form. So, instead, the solution is to check the `third` form, rather than the `second`. This is because, in a properly formed `rec`, the third form will be an atom _if and only if_ the `rec` is of the form
```lisp
(rec qual-ty name bindings &body body)
```
So the test can be something like
```lisp
(cond
  ((cst:atom (cst:third form))
   (values (cst:second form) (cst:third form) (cst:nthrest 3 form)))
  (t
   (values nil (cst:second form) (cst:nthrest 2 form))))
```
And then check for proper formation of each component below. Note that we already check for the case `(rec)` before the `cond`.

YarinHeffes (approved):
`named-while-let` next? :) edit: `while-rec` 😎 
Also, there should be some announcement, maybe on the discord, for anyone who has existing functions named `rec` to make the necessary changes.

stylewarning:
> `named-while-let` next? :) edit: `while-rec` 😎 Also, there should be some announcement, maybe on the discord, for anyone who has existing functions named `rec` to make the necessary changes.

I will announce when merged.

src/faux-macros.lisp
```diff
@@ -84,6 +84,9 @@
 (define-coalton-editor-macro coalton:let (bindings &body form)
   "A lexical LET binding.")
 
+(define-coalton-editor-macro coalton:rec (name bindings &body body)
```
YarinHeffes, line 87: After spending a bit more time with this, I think the following syntax may be favorable.
```lisp
(rec name bindings body)
(rec (name qual-ty) bindings body)
```
as opposed to what we have now:
```lisp
(rec name bindings body)
(rec qual-ty name bindings body)
```

YarinHeffes (changes requested):
Minor syntax change request before merging (apologies for going back on this).

src/faux-macros.lisp
stylewarning: I agree with this.

src/parser/expression.lisp
```diff
@@ -667,6 +670,120 @@ Rebound to NIL parsing an anonymous FN.")
...
+            (values (cst:second form)
+                    nil
+                    (cst:third form)
+                    (cst:nthrest 3 form)))
+           ;; (rec (name qual-ty) bindings ...)
+           (t
+            (unless (cst:proper-list-p (cst:second form))
+              (parse-error "Malformed rec"
+                           (note source (cst:second form)
+                                 "expected list")))
```
YarinHeffes, line 704: Maybe `"unexpected empty list"`?

src/parser/expression.lisp
```diff
@@ -667,6 +670,120 @@ Rebound to NIL parsing an anonymous FN.")
...
+              ((cst:null (cst:rest (cst:second form)))
+               (values (cst:first (cst:second form))
+                       nil
+                       (cst:third form)
+                       (cst:nthrest 3 form)))
+              (t
+               (unless (cst:null (cst:nthrest 2 (cst:second form)))
+                 (parse-error "Malformed rec"
+                              (note-end source (cst:second (cst:second form))
+                                        "extra arguments after type declaration")))
```
YarinHeffes, line 715: Or, maybe for consistency, `"unexpected trailing form(s)"` and refer to `(cst:nthrest 2 (cst:second form))`?

src/parser/expression.lisp
YarinHeffes, line 228: Update this with the new syntax `(identifier | "(" identifier qualified-ty ")")`.

YarinHeffes (changes requested):
Just a few nits before merging. Make sure the tests are updated if you decide to change the error messages.

src/parser/expression.lisp
fukamachi: I didn't mean for an empty list, but a dotted list.
I'll add both cases to the test.

stylewarning (approved):

Issue #1402, opened 2025-03-13
Warnings when recompiling Coalton


YarinHeffes:
The following warnings appear when recompiling Coalton
```
; in: %DEFINE-INTEGER-ROUNDINGS I8
;     (LAMBDA ()
;       (LET* ((#:G523 (LOAD-TIME-VALUE #))
;              (#:NEW522 (COALTON-LIBRARY/MATH/REAL::CLASS/QUANTIZABLE # # #)))
;         (MULTIPLE-VALUE-BIND (#:NEW522)
;             (THE COALTON-LIBRARY/MATH/REAL::CLASS/QUANTIZABLE (VALUES #:NEW522))
;           (FUNCALL #'(SETF #) #:NEW522 #:G523))))
; 
; caught COMMON-LISP:STYLE-WARNING:
;   Call to COALTON-LIBRARY/MATH/NUM::%HANDLE-8BIT-OVERFLOW could not be inlined
;   because its source code was not saved. A global INLINE or SB-EXT:MAYBE-INLINE
;   proclamation must be in effect to save function definitions for inlining.
; in: %DEFINE-INTEGER-ROUNDINGS I32
;     (LAMBDA ()
;       (LET* ((#:G559 (LOAD-TIME-VALUE #))
;              (#:NEW558 (COALTON-LIBRARY/MATH/REAL::CLASS/QUANTIZABLE # # #)))
;         (MULTIPLE-VALUE-BIND (#:NEW558)
;             (THE COALTON-LIBRARY/MATH/REAL::CLASS/QUANTIZABLE (VALUES #:NEW558))
;           (FUNCALL #'(SETF #) #:NEW558 #:G559))))
; 
; caught COMMON-LISP:STYLE-WARNING:
;   Call to COALTON-LIBRARY/MATH/NUM::%HANDLE-32BIT-OVERFLOW could not be inlined
;   because its source code was not saved. A global INLINE or SB-EXT:MAYBE-INLINE
;   proclamation must be in effect to save function definitions for inlining.
; in: %DEFINE-INTEGER-ROUNDINGS I64
;     (LAMBDA ()
;       (LET* ((#:G595 (LOAD-TIME-VALUE #))
;              (#:NEW594 (COALTON-LIBRARY/MATH/REAL::CLASS/QUANTIZABLE # # #)))
;         (MULTIPLE-VALUE-BIND (#:NEW594)
[...]

PR #1406, opened 2025-03-17
Change overflow handler inlined function definition to macro

YarinHeffes:
We previously used a macro that defined an inlined function for handling signed integer overflows. This PR changes that to a macro that expands into the body of that function, fixing #1402. Note that this PR is not _ideal_ in the sense that the original code _should_ have worked and this PR does not address the cause of the original failure, but is instead a work-around.

### DIFF
--- a/library/math/num.lisp
+++ b/library/math/num.lisp
@@ -143,15 +143,17 @@
 
   (lisp-toplevel ()
 
-    (cl:declaim (cl:inline %unsigned->signed))
-    (cl:defun %unsigned->signed (bits x)
-      ;; This is the two's complement conversion of X (interpreted as BITS
-      ;; bits) to a signed integer (as a Lisp object).
-      (cl:-
-       (cl:ldb (cl:byte (cl:1- bits) 0) x)
-       (cl:dpb 0 (cl:byte (cl:1- bits) 0) x)))
-
     (cl:eval-when (:compile-toplevel :load-toplevel)
+      (cl:defmacro %unsigned->signed (bits expr)
+        ;; This is the two's complement conversion of X (interpreted as BITS
+        ;; bits) to a signed integer (as a Lisp object).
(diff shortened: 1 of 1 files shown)


YarinHeffes:
Not sufficient:
```lisp
COALTON-USER> (coalton (the I8 (fromInt 300)))
; in: COALTON (THE I8 (FROMINT 300))
;     (LAMBDA ()
;       (LET ((#:X-1707600 300))
;         (DECLARE (IGNORABLE #:X-1707600))
;         (THE (VALUES (SIGNED-BYTE 8) &OPTIONAL)
;              (LET (#)
;                (VALUES #)))))
; 
; caught COMMON-LISP:STYLE-WARNING:
;   Call to COALTON-LIBRARY/MATH/NUM::%UNSIGNED->SIGNED could not be inlined
;   because its source code was not saved. A global INLINE or SB-EXT:MAYBE-INLINE
;   proclamation must be in effect to save function definitions for inlining.
; 
; compilation unit finished
;   caught 1 STYLE-WARNING condition
; Evaluation aborted on #<COMMON-LISP:SIMPLE-ERROR "Signed value overflowed 8 bits." {7011094A83}>.
```

YarinHeffes:
Changed `%unsigned->signed` to be a macro as well. This is still a temporary fix but I suppose it still is an improvement to the extent that some additional constants are pre-computed.

stylewarning:
I am going to investigate this a bit before we merge anything.

YarinHeffes:
> I am going to investigate this a bit before we merge anything.

Sounds good, you might like to start here:

https://github.com/coalton-lang/coalton/blob/main/src/parser/toplevel.lisp#L822

I have a hunch that the call to `eval` might need to be replaced with something else, though I'm not sure with what.

YarinHeffes:
superseded by #1409

Issue #1408, opened 2025-03-18
LISP-TOPLEVEL doesn't behave well with file compiler semantics


stylewarning:
`lisp-toplevel` is intended to allow toplevel Lisp code to be freely mixed with toplevel Coalton code. A principal use-case is to define Lisp functions or macros to be used in Coalton code, toplevel or otherwise.

One specific desired pattern is something schematically like this:

```
(coalton-toplevel
  (lisp-toplevel ()
    (cl:declaim (cl:inline f))
    (cl:defun f (x) x)
    (cl:defmacro def (name)
      `(define (,name x) (f x))))
  (def foo))
```

However, this doesn't work because the macro `def` is not available to *the Coalton compiler*. In ordinary Lisp, the file compiler arranges for `defmacro` to be available to the rest of the file, however in Coalton, we are more eager to look at definitions and macroexpand before invoking the file compiler.

(It is worth noting that the Coalton compiler is typically invoked at *read time*: we define a macro character on `#\(` and proceed to do Coalton compilation then, expanding into Lisp code. This adds to the complexity of the behavior. Things like `eval-when` cannot be used to tell the Coalton compiler about things like macros.)

This led to awkward fixes/workarounds, as remarked by @YarinHeffes, such as the Coalton parser checking if it's parsing `eval-when`, looking for a `:compile-toplevel`, and [consequently `eval`ing its contents](https://github.com/coalton-lang/coalton/blob/main/src/parser/toplevel.lisp#L822). This is a sort of emulation of `eval-when` for Coalton's purposes, but leads to a variety of bugs. One such bug that plays in the read-time nature of Coalton compilation and the aforementioned workaround is code like

```
'(coalton-codegen
    (lisp-toplevel ()
      (eval-when (:compile-toplevel)
        (very-bad-and-dangerous-thing))))
```

getting evaluated despite the quote.

I haven't figured out if it's related or not, but this also caused issues with things persisting in the FASLs, especially around Lisp-defined functions and declaiming them as `inline`. Some defined functions with `defun` wouldn't persist altogether. This led to issue #1402, whose fix involves removing or working around `lisp-toplevel`.

Solving the matter of this issue would involve:

- Review how Coalton code gets compiled and macroexpanded, starting at the read macro for `#\(`.
- Reviewing the current implementation of `lisp-toplevel`, from syntax to codegen and everything in between.
[...]

PR #1409, opened 2025-03-18
move Num-related macros to its own file

stylewarning:
These macros were moved due to the variety of issues present in the LISP-TOPLEVEL form. See issue #1408 and #1402 for details.

This commit also makes `define-sxhash-hasher` a proper Coalton macro.

### DIFF
--- a/library/hash.lisp
+++ b/library/hash.lisp
@@ -2,8 +2,9 @@
   (:use
    #:coalton
    #:coalton-library/classes)
+  (:import-from #:coalton-library/math/hash-defining-macros
+                #:define-sxhash-hasher)
   (:export
-   #:Hash
    #:combine-hashes
    #:combine-hashes-order-independent
    #:define-sxhash-hasher))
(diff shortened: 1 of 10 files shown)


stylewarning:
This is proposed to supersede #1406.

stylewarning:
Fixes #1402.

library/hash.lisp
```diff
@@ -15,6 +15,14 @@
 #+coalton-release
 (cl:declaim #.coalton-impl/settings:*coalton-optimize-library*)
 
+;;; Utility for defining Hash instances.
+(cl:defmacro define-sxhash-hasher (type)
+  "Define an instance of Hash for TYPE using CL:SXHASH."
+  `(define-instance (Hash ,type)
+     (define (hash item)
```
YarinHeffes, line 22: Perhaps, might as well take the opportunity to add `(inline)` here?

library/hash.lisp
```diff
@@ -86,16 +94,14 @@ The hash function must satisfy the invariant that `(== left right)` implies `(==
...
-     (define-instance (Hash ,type)
-       (define (hash item)
-         (lisp Hash (item)
-           (cl:sxhash item))))))
+        0)))
 
-(define-sxhash-hasher Hash)
+  ;; Don't use DEFINE-SXHASH-HASHER here because it is not yet known
+  ;; to the Coalton compiler. See related issue #1408.
+  (define-instance (Hash Hash)
```
YarinHeffes, line 101: There should be no problem using `(define-sxhash-hasher Hash)` here, if you wrap the `defmacro` with `eval-when`, I tested and it compiles. This is because the macro isn't defined in a `lisp-toplevel`.

YarinHeffes (approved):
Overall this looks good to me. I don't know that we need to move all of the macros to a separate file, since it seems that the issues were localized to the two overflow-handler related definitions, which could have been moved to the top of `num.lisp`. This fixes the issue, and there's the understanding that we plan to fix `lisp-toplevel` and perhaps move things back to the way they were. Aside from the two nits, this is good to go.

YarinHeffes:
That makes sense, thank you for clarifying 👍

library/hash.lisp
stylewarning: My consternation with using EVAL-WHEN around DEFMACRO is that the Common Lisp standard already demands that the macro be available to the rest of the file. 

> If a defmacro form appears as a top level form, the compiler must store the macro definition at compile time, so that occurrences of the macro later on in the file can be expanded correctly. Users must ensure that the body of the macro can be evaluated at compile time if it is referenced within the file being compiled.

Our issue is that we have broken expected semantics of file compilation with how we compile Coalton, requiring extra EVAL-WHENs so the Coalton compiler knows about it.

My personal opinion is that we shouldn't proliferate workarounds due to Coalton's strange file compilation semantics. In this case, I'd rather do what I did with num.lisp and make a new file for this macro alone; it just seemed like overkill. But this example is so small that maybe my "shouldn't proliferate" opinion is too extreme, and adding an EVAL-WHEN with a comment is harmless.

library/hash.lisp
stylewarning: This is a good idea.

library/hash.lisp
stylewarning: @YarinHeffes I tried wrapping it in `eval-when` but it didn't work. It will only work if it's inside of a `lisp-toplevel`.

stylewarning: It also didn't work in a `lisp-toplevel` for me.

PR #1421, opened 2025-03-20
Add inlining to coalton-library/math/integral

YarinHeffes:
Add inlining to `coalton-library/math/integral`

### DIFF
--- a/library/math/integral.lisp
+++ b/library/math/integral.lisp
@@ -211,6 +211,7 @@ are floored and truncated division, respectively."
            (odd? (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-ODD?")))
            (gcd (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-GCD")))
            (^ (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-^")))
+           (^^ (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-^^")))
            (lcm (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-LCM")))
            (isqrt (cl:intern (cl:concatenate 'cl:string (cl:symbol-name type) "-ISQRT"))))
     `(cl:progn 
@@ -219,7 +220,7 @@ are floored and truncated division, respectively."
        (coalton-toplevel
          (define-instance (Integral ,type)
            (inline)
-           (define toInteger into))
+           (define (toInteger x) (into x)))
 
          (specialize even? ,even? (,type -> Boolean))
          (inline)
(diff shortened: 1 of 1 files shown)


stylewarning (changes requested):
Blanket inlining isn't the right thing to do here. Consider that any of these functions might work on very large integers.

What would be better is to specialize these functions and inline those specializations.

library/math/integral.lisp
```diff
@@ -210,47 +217,54 @@ are floored and truncated division, respectively."
...
-          (lisp Boolean (n) (to-boolean (cl:oddp n))))
-
-        (specialize ^ ,^ (,type -> ,type -> ,type))
-        (declare ,^ (,type -> ,type -> ,type))
-        (define (,^ base power)
-          ,(cl:if signed
-                  `(if (< power 0)
+         (define-instance (Integral ,type)
+           (inline)
+           (define toInteger into))
```
stylewarning, line 222: can we write this in function form? `(tointeger x)` etc.

library/math/integral.lisp
```diff
@@ -210,47 +217,54 @@ are floored and truncated division, respectively."
...
+           (lisp Boolean (n) (to-boolean (cl:evenp n))))
+
+         (specialize odd? ,odd? (,type -> Boolean))
+         (inline)
+         (declare ,odd? (,type -> Boolean))
+         (define (,odd? n)
+           (lisp Boolean (n) (to-boolean (cl:oddp n))))
+
+         (specialize ^ ,^ (,type -> ,type -> ,type))
+         (inline)
```
stylewarning, line 237: only inline if unsigned, I think

library/math/integral.lisp
```diff
@@ -210,47 +217,54 @@ are floored and truncated division, respectively."
...
+
+         (specialize lcm ,lcm (,type -> ,type -> ,type))
+         (inline)
+         (declare ,lcm (,type -> ,type -> ,type))
+         (define (,lcm a b)
+           ;; Allow Coalton to handle fixnum overflow
+           (fromInt (lisp Integer (a b) (cl:lcm a b))))
+
+         (specialize isqrt ,isqrt (,type -> ,type))
+         (inline)
```
stylewarning, line 260: only inline if unsigned I think

library/math/integral.lisp
```diff
@@ -271,13 +285,15 @@ are floored and truncated division, respectively."
 
     `(coalton-toplevel
        (specialize ^ ,^ (,type -> Integer -> ,type))
+       (inline)
```
stylewarning, line 288: in general i'm not keen to inline things that expand into an `if` with an error

stylewarning (changes requested):

library/math/integral.lisp
YarinHeffes: Fixed, let me know if you're satisfied with the code-style of the fix

library/math/integral.lisp
YarinHeffes: Fixed

stylewarning (approved):

Issue #1414, opened 2025-03-19
list:union is broken


stylewarning:
it's checking the same condition twice in an OR: `(or (member x ys) (member x ys))`

PR #1423, opened 2025-03-20
fix list:union conditional

bernardoras:
Fixes issue #1414.<br><br>In line 343, the if expression checks, for each xs item, only if it is a member of ys, when building the accumulator of xs items for later including ys remaining unique items to the union.<br>That causes repeated items in xs that are not in ys to be ignored.<br><br>Fix: check if each xs item is also already not in the current accumulator.<br><br>Added two assertions to cover lists with repeated items (the second of which fails) https://github.com/coalton-lang/coalton/blob/44d9e2facb571c3fea5cf4b7caac16d08575181a/library/list.lisp#L332-L346

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -341,7 +341,7 @@
           ((Nil) acc)
           ((Cons x xs)
            (if (or (member x ys)
-                   (member x ys))
+                   (member x acc))
                (f xs acc)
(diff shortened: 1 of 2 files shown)


stylewarning (approved):
Thanks!

Issue #1390, opened 2025-02-13
Some specialized class instances are not inlined properly.


YarinHeffes:
As far as I can tell, the specialized methods `==` and `+` are implemented according to the same pattern for complex floats. However only the specialized `==` is inlined as expected.

```lisp
COALTON-USER> (coalton-toplevel
                (define x (lisp (Complex Double-Float) () #C(1.0d0 2.0d0)))
                (define y (lisp (Complex Double-Float) () #C(3.0d0 3.0d0))))
; No value
COALTON-USER> (coalton-codegen-types
                (monomorphize) (define z (+ x y))
                (monomorphize) (define p (== x y)))
(COMMON-LISP:PROGN
 (COMMON-LISP:DECLAIM
  (SB-EXT:MUFFLE-CONDITIONS SB-KERNEL:REDEFINITION-WARNING))
 (COALTON-IMPL/GLOBAL-LEXICAL:DEFINE-GLOBAL-LEXICAL P COMMON-LISP:BOOLEAN)
 (COMMON-LISP:SETF P
                     (COMMON-LISP:LET ((#:B-11129489 Y))
                       (COMMON-LISP:DECLARE
                        (COMMON-LISP:IGNORABLE #:B-11129489))
                       (COMMON-LISP:LET ((#:A-11119488 X))
                         (COMMON-LISP:DECLARE
                          (COMMON-LISP:IGNORABLE #:A-11119488))
                         (COMMON-LISP:LET ((#:A-11529493 #:A-11119488))
                           (COMMON-LISP:DECLARE
                            (COMMON-LISP:IGNORABLE #:A-11529493))
                           (COMMON-LISP:LET ((#:B-11539494 #:B-11129489))
                             (COMMON-LISP:DECLARE
                              (COMMON-LISP:IGNORABLE #:B-11539494))
                             (COMMON-LISP:THE
                              (COMMON-LISP:VALUES COMMON-LISP:BOOLEAN
                                                  COMMON-LISP:&OPTIONAL)
                              (COMMON-LISP:LET ((COALTON-LIBRARY/MATH/COMPLEX::B
                                                 #:B-11539494)
                                                (COALTON-LIBRARY/MATH/COMPLEX::A
                                                 #:A-11529493))
                                (COMMON-LISP:DECLARE
                                 (COMMON-LISP:TYPE
                                  (COMMON-LISP:COMPLEX
                                   COMMON-LISP:DOUBLE-FLOAT)
[...]

PR #1433, opened 2025-03-24
Brute-force propagate `let` types upward to resolve specialization issue

stylewarning:
(authored by @garlic0x1)

### DIFF
--- a/src/codegen/optimizer.lisp
+++ b/src/codegen/optimizer.lisp
@@ -715,7 +715,12 @@ NODE in the environment ENV."
 
                         (specialization (tc:lookup-specialization-by-type env rator-name rator-type :no-error t)))
                    (unless specialization
+                     (print "NO SPECIALIZATION FOUND")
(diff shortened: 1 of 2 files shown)


src/codegen/optimizer.lisp
```diff
@@ -193,6 +193,12 @@
       (action (:before node-let) #'add-local-funs)
       (action (:before node-bind) #'add-bind-fun)))))
 
+(defun propagate-let-type (node)
```
YarinHeffes, line 196: Add comment above indicating that this is a temporary, brute-force solution to #1390 and that the type of the let nodes should be handled earlier in the codegen process.

YarinHeffes (changes requested):
This does seem to fix #1390. I think we can merge this PR and close #1390, then open a new issue to resolve the let types earlier in the codegen.

YarinHeffes (approved):

PR #1435, opened 2025-03-24
Modify native Remainder instances to include floats

YarinHeffes:
Enables, for example,
```lisp
> (coalton (mod 2.0 1.7))
0.3
```


library/math/integral.lisp
```diff
@@ -185,26 +185,26 @@ are floored and truncated division, respectively."
...
        (inline)
-       (define (div a n)
-         (lisp ,type (a n)
-           (cl:nth-value 0 (cl:floor a n))))
-       (inline)
        (define (divMod a n)
          (lisp (Tuple ,type ,type) (a n)
-           (cl:multiple-value-call 'Tuple (cl:floor a n)))))))
+           (cl:multiple-value-bind (d m) (cl:floor a n)
+             (Tuple (coalton (the ,type (fromInt (lisp Integer () d)))) m)))))))
```
stylewarning, line 207: Nested COALTON calls seem to break the Coalton compiler sometimes it seems.

YarinHeffes: Addressed here https://github.com/coalton-lang/coalton/pull/1435/commits/05c6c184611881710ee7df532bc77dca2d19125d

PR #1447, opened 2025-04-26
Limit Inlining in Recursive Functions

garlic0x1:


src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+
+(defparameter *inliner-max-unroll* 4
+  "Limit depth to unroll recursive functions to.")
+(defparameter *inliner-max-depth*  16
+  "Limit total inliner stack depth.")
+(defparameter *inliner-heuristic*  'gentle-heuristic
+  "Heuristic to determine if a function should be inlined
+even if the user didn't explicity declare it to be.")
+
+;; Private Settings for Debugging
```
stylewarning, line 34: You probably want these DEFVAR, and comment what they do (unless you plan on removing them).

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+(defparameter *inline-methods-p* t)
+(defparameter *inline-globals-p* t)
+(defparameter *inline-lambdas-p* t)
+
+(defvar *functions-inlined*)
+
+(defun debug! (fmt &rest args)
+  "Convenience function to print debug when `settigns:*print-inlining-occurences*' is enabled."
+  (when settings:*print-inlining-occurrences*
+    (apply #'format t (uiop:strcat "~&" fmt) args)))
```
stylewarning, line 45: don't strcat; just (FRESH-LINE) before the APPLY

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+
+(defun max-depth-p (stack)
+  "Determine if the inliner is at its maximum depth, by default this is 16."
+  (declare (type list stack)
+    (values boolean))
+
+  (<= *inliner-max-depth* (length stack)))
+
+(defun lookup-code-global (node env)
+  "Try to lookup the code of an globally known function, returns null or abstraction."
```
stylewarning, line 94: an -> a

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
-        (list
-         (action (:after node-application) #'try-inline)
-         (action (:after node-direct-application) #'try-inline)
-         (make-traverse-let-action-skipping-cons-bindings))
-        nil)
-       inline-happened?))))
+(defun heuristic-inline-p (node)
+  "Determine if the node should be inlined based on heuristics."
+  (declare (type (or null ast:node-abstraction) node)
+    (values boolean))
```
stylewarning, line 71: where did this indent style come from? I think all clauses should be aligned

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+
+  (and node (funcall *inliner-heuristic* node)))
+
+;; Utilities
+
+(defun unrolledp (node stack)
```
stylewarning, line 77: these names will need some workshopping

it's not that node is unrolled, it's that node isn't allowed to be unrolled/inlined further (even mutually recursively) in the current context

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+  "Determine if the inliner has fully unrolled a recursive call."
+  (declare (type (or ast:node-application ast:node-direct-application) node)
+    (type list stack)
+    (values boolean))
+
+  (a:when-let ((name (ast:node-rator-name node)))
```
stylewarning, line 83: if-let and explicitly return nil for the false case

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+    (values boolean))
+
+  (<= *inliner-max-depth* (length stack)))
+
+(defun lookup-code-global (node env)
+  "Try to lookup the code of an globally known function, returns null or abstraction."
+  (declare (type ast:node node)
+    (values (or null ast:node-abstraction)))
+
+  (let ((code (tc:lookup-code env (ast:node-rator-name node) :no-error t)))
```
stylewarning, line 98: (let ((code ..)) (if (abs-p code) code nil)

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+    (and (ast:node-abstraction-p code) code)))
+
+(defun lookup-code-anonymous (node)
+  "Try to lookup the code of an anonymous application, returns null or abstraction."
+  (declare (type ast:node node)
+    (values (or null ast:node-abstraction)))
+
+  (let ((code (ast:node-application-rator node)))
```
stylewarning, line 106: same here.

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+    (and (ast:node-abstraction-p code) code)))
+
+(defun inlinable-function-p (name env)
+  "Check if a function is declared inlinable at its definition."
+  (declare (type symbol name)
+    (type tc:environment env)
+    (values boolean))
+
+  (a:when-let ((entry (tc:lookup-function env name :no-error t)))
```
stylewarning, line 115: if-let

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,361 @@
...
+    (tc:function-env-entry-inline-p entry)))
+
+(defun fully-applied-p (node code)
+  "Check if an an application node constitutes a fully applied abstraction."
+  (declare (type (or ast:node-application ast:node-direct-application) node)
+    (type (or null ast:node-abstraction) code)
+    (values boolean))
+
+  (and code
```
stylewarning, line 124: if (null code) nil (= ...)

i think its moderately dubious we allow nil as inputs in so many places. why are we so loose with this?

stylewarning (commented):
just some preliminary comments.

I didn't study the code deeply enough, but I think there should be some more rigor in when we do and don't use (or null x) types on inputs to functions.

The return types of these functions generally needs to be

    (values x &optional)

to pin just a single value return.

src/codegen/inliner.lisp
garlic0x1: idk but it is annoying, will fix

src/codegen/inliner.lisp
garlic0x1: I was optimizing for brevity in the `cond` in `inline-application`, I agree it isn't very good, but I also don't want that to become unreadable again

src/codegen/inliner.lisp
garlic0x1: I am not sure if they should be left in or not, no normal person would want to touch these, please advise

src/codegen/inliner.lisp
garlic0x1: in the `cond`, lookup-code functions also have duplicate calls, perhaps these should be let-bound... I just liked the look of flat `and`s but it is bad if it hurts performance or maintenance

src/codegen/inliner.lisp
stylewarning: I'm perfectly fine with useful developer settings that are not exported. This kind of thing could be useful for future debugging.

src/codegen/inliner.lisp
garlic0x1: `no-unroll-p` perhaps?

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,370 @@
...
-        nil)
-       inline-happened?))))
+(defun heuristic-inline-p (node)
+  "Determine if the node should be inlined based on heuristics."
+  (declare (type (or null ast:node-abstraction) node)
+           (values boolean &optional))
 
-(defun inline-applications (node env)
-  "Perform inlining on NODE in the environment ENV.
+  (and node (funcall *inliner-heuristic* node)))
```
stylewarning, line 79: what breaks if this doesn't accept NIL?

garlic0x1: the callsites might get `nil` from lookup-code, but could look like this:

```
(let ((code (lookup-code-anonymous node)))
  (and *inline-lambdas-p*
       code
       (heuristic-inline-p code)
       (fully-applied-p node code)))
```

garlic0x1:
perhaps the `cond`s in method-inline functions could be deduped, but they aren't really asking for it and it's probably less confusing as-is

src/codegen/inliner.lisp
stylewarning: I think this or something like this is preferable.

src/codegen/codegen-expression.lisp
```diff
@@ -93,6 +93,12 @@
        (node-variables expr :variable-namespace-only t)
        env)))
 
+  (:method ((expr node-locally) env)
+    (declare (type tc:environment env))
+    ;; TODO: Remove debug string
```
stylewarning, line 98: remove

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+       node))))
+
+(defun extract-dict (rands)
+  (declare (type ast:node-list rands)
+           (values (or null parser:identifier) ast:node-list &optional))
+
+  (cond
+    ((ast:node-variable-p (first rands))
+     (values (ast:node-variable-value (first rands))
+             (cdr rands)))
```
stylewarning, line 243: nit: cdr -> rest

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+
+    ((and (ast:node-application-p (first rands))
+          (ast:node-variable-p (ast:node-application-rator (first rands))))
+     (values (ast:node-variable-value (ast:node-application-rator (first rands)))
+             (append (ast:node-application-rands (first rands)) (cdr rands))))
```
stylewarning, line 248: not: cdr -> rest

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+           (values ast:node &optional))
+
+  (unless *inline-methods-p*
+    (return-from inline-method
+      node))
+
+  (let ((rator (ast:node-application-rator node))
+        (rands (ast:node-application-rands node)))
+    (multiple-value-bind (dict inner-rands) (extract-dict rands)
+      (if (and dict (ast:node-variable-p rator))
```
stylewarning, line 265: put simpler base case first

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+           (type tc:environment env)
+           (values ast:node &optional))
+
+  (unless *inline-methods-p*
+    (return-from inline-direct-method
+      node))
+
+  (let ((rands (ast:node-direct-application-rands node)))
+    (multiple-value-bind (dict inner-rands) (extract-dict rands)
+      (if dict
```
stylewarning, line 302: put simpler base case first, and do explicit null check

src/codegen/inliner.lisp
stylewarning: or do more imperative style:

```
(unless (and ...)
  (return-from ...))

...
```

same with below

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+   node
+   (list
+    (traverse:action (:before ast:node-locally node)
+      (dolist (sym (ast:node-locally-noinline-functions node))
+        (push sym noinline-functions))
+      node)
+    (traverse:action (:before ast:node-application node)
+      (push (ast:node-rator-name node) stack)
+      node)
+    (traverse:action (:before ast:node-direct-application node)
```
stylewarning, line 345: can we instead do a `:traverse` with a dynamic binding, so we don't need to do the pushes and pops manually?

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
+      (push (ast:node-rator-name node) stack)
+      node)
+
+    (traverse:action (:after ast:node-locally node)
+      (dolist (sym (ast:node-locally-noinline-functions node))
+        (pop noinline-functions))
+      node)
+    (traverse:action (:after ast:node-application node)
```
stylewarning, line 353: nit: fix spacing: newlines between each new action

src/codegen/inliner.lisp
```diff
@@ -1,313 +1,379 @@
...
-      (values
-       (traverse
-        node
-        (list
-         (action (:after node-application) #'inline-method)
-         (action (:after node-direct-application) #'inline-direct-method)))
-       inline-happened?))))
+           (values ast:node list &optional))
 
+  (let ((*functions-inlined* ()))
```
stylewarning, line 376: nit: quote `()` or use `(list)`

stylewarning (changes requested):
some initial comments

shirok (approved):
I tried hard to break the inliner but I concede.  So far, contrived ways to introduce recusion indirectly (e.g. a chain of inline expansions resulting self recurisve calls) are excluded from further inlining, so I conclude this is safe.

stylewarning (changes requested):

garlic0x1:
addressed stylewarnings, I think it is more clear now

stylewarning (approved):

PR #1448, opened 2025-04-27
Fix lisp-type for LispArray of type with multiple tyvars

shirok:
Compiler failed when LispArray's element type has more than one type variables.

For example, if you want to use something like this:
  (LispArray (T :a :b))

You got an error like:

  The value
    (T #T123)
  is not of type
    COALTON-IMPL/TYPECHECKER/TYPES:TYCON


stylewarning (approved):

stylewarning:
Thanks.

PR #1456, opened 2025-05-15
Add hash array mapped trie (HAMT)

shirok:
Immutable map using compact trie representation.
Interface is similar to hashtable (it is debatable; maybe we'd better to align to ord-map?).
Implements IntoIterator, FromIterator, Eq, Hash typeclasses.

Not highly optimized, but compared to ord-map, it is
  2.5x faster on building
  3.3x faster on lookup
  2.2x faster on entry removal
on maximum 1M entry mapping.  You can run benchmarks with benchmarks/mapping.lisp

## Notes on API

Currently it is a mixture of `hasthable` and `ord-map`:  `get` and `count` is from `hashtable`.  `ord-map` has pretty sophisticated API, but I feel they're a bit too much for casual use (e.g. when inserting an entry, sometimes you do want to know whether the key has already existed, but most of the time you don't care).  I'm inclined to devise a common Mapping API.   That PR would touch a lot more files, so I think it's better to merge this first.


shirok:
Hmm, stack overflow only on CCL for large HAMT test.  Need to investigate.

stylewarning:
@shirok  Do you have thoughts on a friendly name for this data structure? Should be short and easy, like Seq. Map? HashMap? Assoc? Dict?

shirok:
> @shirok Do you have thoughts on a friendly name for this data structure? Should be short and easy, like Seq. Map? HashMap? Assoc? Dict?

Since we have a package `ord-map`, we can call this package `hash-map`.  Then type name can be `HashMap`.  The asymmetry with `Map` is a bit awkward, though.   It is essential that key must be hashable for this, so having `hash` is relevant. 

`Dict` or `Assoc` can be an umbrella name over ordmap and hashmap.

stylewarning:
We should rename the other type OrdMap.

shirok:
> We should rename the other type OrdMap.

If we can do so, that's perfect.  I will change API to align to ord-map then.

library/hashmap.lisp
```diff
@@ -0,0 +1,523 @@
...
+      ((Cons e es)
+       (match e
+         ((HmEntry k1 _)
+          (if (== key k1)
+              (Cons (HmEntry key val) es)
+              (let ((es2 (replace key val es)))
+                (if (unchanged? es es2)
+                    es
+                    (Cons e es2)))))))))
+
```
macrologist, line 196: style nit. Consider deeper destructuring

``` lisp

((Cons (HmEntry k1 _) es)
  (if (== key k1)
      ....)
```

library/hashmap.lisp
```diff
@@ -0,0 +1,523 @@
...
+             None))
+        ((Bud (HmEntry k1 v1) (HmEntry k2 v2))
+         (cond ((== key k1) (Some v1))
+               ((== key k2) (Some v2))
+               (True None)))
+        ((Chain entries)
+         (match (iter:find! (fn (e) (== key (entry-key e)))
+                            (iter:into-iter entries))
+           ((None) None)
+           ((Some e) (Some (entry-value e)))))
```
macrologist, line 317: style nit. Consider `map`

``` lisp
(map entry-value (iter:find! (fn (e) ....))
```

or `do` 

``` lisp
(do 
  (e <- (iter:find! (fn (e) ...))
  (pure (entry-value e)))
```

Although, either one is possibly less clear than `match`

library/hashmap.lisp
```diff
@@ -0,0 +1,523 @@
...
+                     (tree-insert mask arr ind newelt)))))
+          )
+      (HashMap (walk 0 (.root hm)))))
+
+  (declare unchanged? (:a -> :a -> Boolean))
+  (define (unchanged? a b)
+    (lisp Boolean (a b) (cl:eq a b)))
+
+  ;; We avoid using list:remove-if, for we need to keep identity of input
+  ;; when lis is unmodified.  It is also assumed that KEY is unique in XS.
```
macrologist, line 373: s/lis/list

library/hashmap.lisp
macrologist: ahh yes. I missed the later use of the head

PR #1458, opened 2025-05-26
Add some monad transformers to the `monads/` standard library and add an example program demonstrating usage

Jason94:
This PR does two things:

~~1. Adds a MonadState typeclass and MonadTransformer typcelass, in the style of the Haskell MTL library. It wraps the existing ST functions in the new MonadState typeclass.~~
2. Adds an OptionalT, ResultT, and EnvT monad transformers. EnvT (EnvironmentT) is Haskell's ReaderT, which is a more category-theory correct name but IMO less descriptive of what it actually does in code.
3. Adds a somewhat involved bank account example to small-coalton-programs demonstrating using a ST, EnvT, and ResultT together.

The MonadState typeclass was causing some type inference problems in the examples, particularly around accessors. After some reflection, it's probably best to pull that out into a separate PR to follow this one, especially if it was risking breaking existing code. Everything else is perfectly workable with just using `lift`.


library/monad/environment.lisp
```diff
@@ -0,0 +1,133 @@
+(coalton-library/utils:defstdlib-package #:coalton-library/monad/environment
```
Jason94, line 1: Add unit tests for this and other new monads.

library/monad/state.lisp
```diff
@@ -43,7 +45,27 @@ Represented as a closure from initial state to updated state and value."
     (match sc
       ((ST fstate)
        fstate)))
-  
+
+  ;;
+  ;; MonadState Typeclass
+  ;;
+
+  (define-class (Monad :m => MonadState :state :m (:m -> :state))
```
Jason94, line 53: It seems like this has some issues with accessor functions, which is causing the current failure in the brainfold example program.

This code:
```lisp
  (declare move-right (Unit -> (state:ST BF-State Unit)))
  (define (move-right)
    "Moves the pointer one bf-cell to the right."
    (do
     (bfs <- state:get)
     (pure (cell:increment! (.pointer bfs)))
      (state:put bfs)))
```
is throwing this error:
```
error: Ambiguous accessor
  --> <macroexpansion>:8:30
   |
 8 |        (PURE (CELL:INCREMENT! (.POINTER BFS)))
   |                                ^^^^^^^^ accessor is ambiguous
```
I'll dig into the accessor functions that the compiler generates and see if there's some way the type inference can be improved with the typeclass.

stylewarning:
Very cool!

Jason94:
> Very cool!

Thank you! The PR should be ready for review now.

examples/small-coalton-programs/src/monads-bank.lisp
```diff
@@ -0,0 +1,345 @@
...
+         (pure (Ok account))
+         (pure (Err (InvalidAccountBalance (.name account) (.balance account)))))))
+
+  (declare set-account (Account -> BankM (BankResult Account)))
+  (define (set-account acc)
+    "Update/insert an account and return it for convenience."
+    (do
+     (lift (modify (fn (mp)
+               (m:insert-or-replace mp (.name acc) acc))))
+     (pure (Ok acc)))))
```
YarinHeffes, line 211: Is this right? I haven't seen `do` used like this, and it doesn't seem like the result of `lift` is bound to anything.

examples/small-coalton-programs/src/monads-bank.lisp
```diff
@@ -0,0 +1,345 @@
...
+  (declare get-account (AccountName -> BankState -> BankResult Account))
+  (define (get-account account-name accounts)
+    (opt->result (AccountNotFound account-name) (m:lookup accounts account-name)))
+
+  (declare get-accountM (AccountName -> BankM (BankResult Account)))
+  (define (get-accountM account-name)
+    (lift (map (get-account account-name) get)))
+
+  (declare account-valid? (Account -> BankM (BankResult Account)))
+  (define (account-valid? account)
```
YarinHeffes, line 198: Perhaps, `check-account-balance` or `validate-account`. Suffix`?` typically indicates Boolean return type.

examples/small-coalton-programs/src/monads-bank.lisp
```diff
@@ -0,0 +1,345 @@
...
+;;;; our static configuration), and the Result monad (to stop running the
+;;;; program whenever we encounter an error).
+;;;;
+;;;; To completely understand the workings of this example program, it will
+;;;; be helpful to already have a basic grasp of how the State, Environment,
+;;;; and Result monads work.
+;;;;
+;;;; Example usage:
+;;;;
+;;;; CL-USER> (in-package :small-coalton-programs/bank)
```
YarinHeffes, line 24: This example can be more thorough:


```lisp
(run-bankM
  (do-resultT
    (create-account "Alice" 100)
    (create-account "Bob" 100)
    ;; Alice and Bob make a bet... Bob loses
    (transfer "Bob" "Alice" 10)
    print-reportM
    ;; Some additional invalid things happen (use all the API functions)
    ...
    print-reportM)
  (Configuration 100 ; minimum balance
                 True ; overdraft protection is on
                 )
  m:empty ; initial BankState, a map from account names to accounts
  )

```

YarinHeffes (commented):
Overall, this looks good to me! - I looked at everything except the tests. My comments are just about style. 

There are a few instances of functions of the form
```lisp
(define (f (Constructor x)) x)
```
that could be `(inline)`d.

There are also some cases where a function is defined like this
```lisp
(define map (compose f g))
```
which may be better defined like this:
```lisp
(define (map func x) (f (g (func x))))
```

Those are both matters of style too, I didn't make in-line comments about them.

I am not an expert on Monads, so I want a second reviewer before merging, I think.

library/monad/optionalt.lisp
```diff
@@ -0,0 +1,87 @@
...
+          (opta <- ma)
+          (match opta
+            ((None) (pure None))
+            ((Some a)
+             (do
+               (optb <- mb)
+               (match optb
+                 ((None) (pure None))
+                 ((Some b)
+                  (pure (Some (fa->b->c a b))))))))))))
```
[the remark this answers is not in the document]
Jason94: I think the reason I did it this way was as a small performance optimization. If `opta = None`, then the current method avoids flatmapping into mb and allocating the result. I'm not sure if that's Coalton is able to optimize around that in your code (which is written extremely cleanly, that beatiful monad code).

I'll add a comment to the code, unless you don't think it's worth it.

library/monad/resultt.lisp
```diff
@@ -0,0 +1,101 @@
...
+        (resa <- ma)
+        (match resa
+          ((Err er) (pure (Err er)))
+          ((Ok a)
+           (do
+            (resb <- mb)
+            (match resb
+              ((Err er) (pure (Err er)))
+              ((Ok b)
+               (pure (Ok (fa->b->c a b))))))))))))
```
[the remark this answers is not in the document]
Jason94: Similar comment about performance.

library/monad/resultt.lisp
```diff
@@ -0,0 +1,101 @@
...
+  (define-instance (MonadTransformer (ResultT :err))
+    (define lift (compose ResultT (map Ok)))))
+
+;;
+;; Macros
+;;
+
+(cl:defmacro do-resultT (cl:&body body)
+  "Wrap and perform a series of ResultT's, returning the value of the last one
+if all of them succeed."
```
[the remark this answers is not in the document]
Jason94: Agreed. I also beefed it up a bit, so it should be able to handle any kind of `do`, not one that just uses bare statements.

library/monad/state.lisp
```diff
@@ -43,6 +44,13 @@ Represented as a closure from initial state to updated state and value."
     (match sc
       ((ST fstate)
        fstate)))
+  (declare modify ((:state -> :state) -> ST :state Unit))
+  (define (modify statef)
+    "Modify the state in a StatefulComputation, discarding the old state."
+    (do
+     (state <- get)
+     (put (statef state))))
```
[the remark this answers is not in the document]
Jason94: Regarding the `do` indentation, is that a new convention? If so, I'm happy to oblige, just let me know. I've been going off of, e.g., https://github.com/coalton-lang/coalton/blob/main/docs/intro-to-coalton.md#do-notation.

examples/small-coalton-programs/src/monads-bank.lisp
Jason94: (Although I think technically, at least in coalton, every monad function has to return at least a unit, which you could explicitly capture if you wanted. so you could do:

```
(run
  (do
    (x <- (put 10)) ;; this is just Unit, kind of pointless
    ...
```)

Jason94:
> Overall, this looks good to me! - I looked at everything except the tests. My comments are just about style.
> 
> There are a few instances of functions of the form
> 
> ```commonlisp
> (define (f (Constructor x)) x)
> ```
> 
> that could be `(inline)`d.
> 
> There are also some cases where a function is defined like this
> 
> ```commonlisp
> (define map (compose f g))
> ```
> 
> which may be better defined like this:
> 
> ```commonlisp
> (define (map func x) (f (g (func x))))
> ```
> 
> Those are both matters of style too, I didn't make in-line comments about them.
> 
> I am not an expert on Monads, so I want a second reviewer before merging, I think.

Thanks! I inlined a lot of the functions, mostly just going by size. I also agree about compose. It's a bad habit from Haskell, but even I think it makes code less readable to people who didn't write it.

PR #1466, opened 2025-06-30
inline defstruct ctors/readers and freeze type in release

### LINKED ISSUE #1461: Inline Lisp `defstruct` readers in release mode
ADTs are built out of `defstruct` in release mode. The getters can be inlined at the Lisp level by codegenning some `cl:inline` declaims.

h/t @cdegroot

stylewarning:
Adds inlining of generated `defstruct` constructor and readers in release mode. Also freezes the type of the structure.

Fixes #1461.

### DIFF
--- a/src/codegen/struct-or-class.lisp
+++ b/src/codegen/struct-or-class.lisp
@@ -20,34 +20,55 @@
   (name (util:required 'name) :type symbol :read-only t)
   (type (util:required 'type) :type t      :read-only t))
 
-(defun struct-or-class (&key
-                              (classname (error "Class Name required"))
-                              (constructor (error "Constructor required"))
-                              (superclass nil)
-                              (fields nil)
-                              mode)
-  (declare (type symbol classname)
-           (type symbol constructor)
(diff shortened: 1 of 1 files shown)


src/codegen/struct-or-class.lisp
```diff
@@ -20,34 +20,55 @@
...
-           (type list fields)
-           (type (member :class :struct) mode))
+    (type symbol constructor)
+    (type symbol superclass)
+    (type list fields)
+    (type (member :class :struct) mode))
 
-  (let ((field-names (mapcar #'struct-or-class-field-name fields)))
+  (let ((field-names (mapcar #'struct-or-class-field-name fields))
+        (accessor-names (loop :for field :in fields
```
stylewarning, line 41: Precompute these accessor names once instead of repeating it every time.

YarinHeffes (approved):
This is a great change, I have no comments/nits.

From SBCL docs these are stack allocated:
`user-defined structures when the structure constructor defined using defstruct has been declared inline`

PR #1467, opened 2025-06-30
Add type declarations for pattern bindings

### LINKED ISSUE #1074: Generate types for match bindings in codegen
`match` bindings don't codegen a Common Lisp type. This causes some Lisp compiler optimizations to be missed.

Add Lisp type declarations to `match` bindings' codegen.

Place to look (as of 26 Mar 2024) is [here](https://github.com/coalton-lang/coalton/blob/main/src/codegen/codegen-expression.lisp#L175). Look at similar functions for how they generate type declarations.

### LINKED ISSUE #1464: Inefficient monomorphized code involving ADTs
The use of parametric ADTs with pattern matching generates inefficient code, likely due to missing type declarations in the codegenned output.

Repro:

Consider the following definition:

```
(coalton-toplevel
  (define-type (F :a) (F :a))
  (define-type-alias Fs (F Single-Float))

  (declare twice (Num :a => F :a -> F :a))
  (define (twice (F x))
    (F (+ x x))))
```

Now consider its use for `Single-Float`:

```
(coalton-toplevel
  (monomorphize)
  (declare do-test (Fs -> Fs))
  (define (do-test)
    (twice (F 1.0f0))))
```

We see that `+` will get inlined (as desired) and it will be recognized as a single-float `+`:

```
;; Optimization run #1
;; Optimization run #1
;; Inlining method + -> COALTON-LIBRARY/MATH/NUM::|INSTANCE/CLS:NUM SINGLE-FLOAT-COALTON-LIBRARY/CLASSES:+|
;; Inlining COALTON-LIBRARY/MATH/NUM::|INSTANCE/CLS:NUM SINGLE-FLOAT-COALTON-LIBRARY/CLASSES:+|
;; Optimization run #2
;; Optimization run #1
;; Optimization run #1
;; Optimization run #1
;; DO-TEST :: ((F SINGLE-FLOAT) → (F SINGLE-FLOAT))
```

However, types are not emitted on the branch, leading to it being a `GENERIC-+` when compiled as Lisp code:

```
  (DEFUN |TWICE#4293| (#:G1)
    (DECLARE (IGNORABLE #:G1))
    (LET ((#:MATCH19 (THE F #:G1)))
      (DECLARE (IGNORABLE #:MATCH19))
      (LOCALLY
       (DECLARE (SB-EXT:MUFFLE-CONDITIONS SB-EXT:CODE-DELETION-NOTE))
       (COND
        ((AND (TYPEP #:MATCH19 'F/F) T)
         (LET ((X-4291 (F/F-_0 #:MATCH19)))  ; <-- *** (1)
           (DECLARE (IGNORABLE X-4291))
           (F
            (LET ((#:B-54316 X-4291))
              (DECLARE (IGNORABLE #:B-54316))
              (LET ((#:A-54215 X-4291))
                (DECLARE (IGNORABLE #:A-54215))
                (LOCALLY
                 (DECLARE (OPTIMIZE (SB-C::TYPE-CHECK 1)))
                 (THE (VALUES SINGLE-FLOAT &OPTIONAL)  ; <-- *** (2)
                      (LET ((COALTON-LIBRARY/MATH/NUM-DEFINING-MACROS::A
                             #:A-54215)
                            (COALTON-LIBRARY/MATH/NUM-DEFINING-MACROS::B
                             #:B-54316))
                        (VALUES
                         (PROGN
                          (+ COALTON-LIBRARY/MATH/NUM-DEFINING-MACROS::A
                             COALTON-LIBRARY/MATH/NUM-DEFINING-MACROS::B)))))))))))
        (T (ERROR "Pattern match not exhaustive error"))))))
```

The output is declared (see line `*** (1)`), but the ADT getters are not (see line `*** (1)`).

h/t @cdegroot

YarinHeffes:
Closes #1074.
Closes #1464.

### DIFF
--- a/src/codegen/codegen-pattern.lisp
+++ b/src/codegen/codegen-pattern.lisp
@@ -5,34 +5,37 @@
    #:coalton-impl/codegen/ast)
   (:local-nicknames
    (#:util #:coalton-impl/util)
+   (#:settings #:coalton-impl/settings)
    (#:tc #:coalton-impl/typechecker))
   (:export
    #:codegen-pattern))
 
 (in-package #:coalton-impl/codegen/codegen-pattern)
 
-(defgeneric codegen-pattern (pattern expr env)
-  (:documentation "Codegen the match branch PATTERN on lisp value EXPR, returning (VALUES PREDICATE BINDINGS).")
-  (:method ((pattern pattern-var) expr env)
-    (declare (type tc:environment env)
-             (ignore env))
+(defgeneric codegen-pattern (pattern expr expr-type env)
+  (:documentation "Codegen the match branch PATTERN on lisp value EXPR of (Coalton) type
+EXPR-TYPE, returning (VALUES PREDICATE BINDINGS BINDING-TYPES).")
+  (:method ((pattern pattern-var) expr expr-type env)
+    (declare (type tc:environment env))
     (values
      `t
-     (list (list (pattern-var-name pattern) expr))))
+     (list (list (pattern-var-name pattern) expr))
+     (list (tc:lisp-type expr-type env))))
 
-  (:method ((pattern pattern-wildcard) expr env)
-    (declare (type coalton-impl/typechecker::environment env))
-    (values `t nil))
+  (:method ((pattern pattern-wildcard) expr expr-type env)
+    (declare (ignore pattern expr expr-type env))
+    (values `t nil nil))
 
-  (:method ((pattern pattern-literal) expr env)
-    (declare (type coalton-impl/typechecker::environment env))
+  (:method ((pattern pattern-literal) expr expr-type env)
+    (declare (ignore expr-type env))
     (values
      (if (stringp (pattern-literal-value pattern))
          `(string= ,(pattern-literal-value pattern) ,expr)
          `(eql ,(pattern-literal-value pattern) ,expr))
+     nil
      nil))
 
-  (:method ((pattern pattern-constructor) expr env)
+  (:method ((pattern pattern-constructor) expr expr-type env)
     (declare (type coalton-impl/typechecker::environment env))
     (let* ((name (pattern-constructor-name pattern))
            (entry (tc:lookup-constructor env name))
@@ -44,6 +47,7 @@
         ((tc:type-entry-enum-repr type)
          (values
           `(eql ',(tc:constructor-entry-compressed-repr entry) ,expr)
+          nil
           nil))
 
         ;; For newtypes we can just codegen patterns for the inner type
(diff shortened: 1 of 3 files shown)


src/codegen/codegen-pattern.lisp
```diff
@@ -51,18 +55,20 @@
          (unless (= 1 (length (pattern-constructor-patterns pattern)))
            (util:coalton-bug "Unexpected number of fields in newtype pattern.~%    Expected: 1~%    Received: ~A~%"
                              (length (pattern-constructor-patterns pattern))))
-         (codegen-pattern (first (pattern-constructor-patterns pattern)) expr env))
+         (codegen-pattern (first (pattern-constructor-patterns pattern)) expr expr-type env))
```
YarinHeffes, line 58: I'm not familiar with newtypes, do I need to unbox `expr-type` here? How do I test this?

src/typechecker/environment.lisp
```diff
@@ -491,7 +491,7 @@
             :constructs 'coalton:List
             :classname nil
             :docstring "`Cons` represents a `List` containing a first element (`car`) and a nested `Cons` (`cdr`)."
-            :compressed-repr 'nil))
+            :compressed-repr 'cons))
```
YarinHeffes, line 494: Is this an actual fix, or was it meant to be `'nil` for some reason I'm unaware of?

src/typechecker/environment.lisp
```diff
@@ -1233,7 +1233,6 @@
   (declare (type symbol name)
            (type environment env)
            (values ty-list &optional))
-  (lookup-constructor env name)
```
YarinHeffes, line 1236: I don't believe this function has any side effects, so it can be removed. I would ask of a code reviewer to double-check this.

src/typechecker/environment.lisp
stylewarning: This isn't correct. It's only applicable for `(repr :enum)`.

src/typechecker/environment.lisp
stylewarning: you are correct.

stylewarning (changes requested):
This isn't a complete review; just wanted to leave a comment on the `compressed-repr` business.

src/codegen/codegen-pattern.lisp
stylewarning: Newtypes are for `repr :transparent` definitions of types. In that case, no new (runtime) type is created. As such, i think your code is correct.

src/codegen/codegen-pattern.lisp
```diff
@@ -5,34 +5,38 @@
...
+  (:documentation "Codegen the match branch PATTERN on lisp value EXPR of (Coalton) type
+EXPR-TYPE, returning (VALUES PREDICATE BINDINGS).")
+  (:method ((pattern pattern-var) expr expr-type env)
+    (declare (type tc:environment env))
     (values
      `t
-     (list (list (pattern-var-name pattern) expr))))
+     (list (list (pattern-var-name pattern)
+                 (if settings:*emit-type-annotations*
+                     `(the ,(tc:lisp-type expr-type env) ,expr)
```
stylewarning, line 24: I think the use of `the` is elegant, but it is preferable to use `declare`. This means we will probably need to make this generic function return a third value, which is a list of declarations. The declarations can be empty, or it can contain a literal `declare` form which is used by the caller.

We should do the same for the actual match expression `(match <expr> ...)` which also gets codegenned with a `the`.

src/codegen/codegen-pattern.lisp
```diff
@@ -80,6 +86,17 @@
         (t
          (let* ((ctor-name (pattern-constructor-name pattern))
                 (ctor (tc:lookup-constructor env ctor-name))
+                (generic-ctor-type (tc:qualified-ty-type
```
stylewarning, line 89: I have not reviewed this portion of the code. It seems suspect, but I haven't thought about it.

stylewarning (changes requested):
Second layer of review.

src/codegen/codegen-pattern.lisp
YarinHeffes: I originally had it like this but changed it because I saw the match expression using `the` 😢

src/codegen/codegen-pattern.lisp
YarinHeffes: I simplified it using `function-return-type`.

PR #1470, opened 2025-07-01
Use DEFINE-TYPE instead of DEFINE-STRUCT for HashMap

stylewarning:
This cleans up the documentation a bit and allows us to guarantee its type.

@shirok Do you have any thoughts about this change?


shirok (approved):
Neat.

PR #1474, opened 2025-07-01
Remove error branch from match codegen in release mode

### LINKED ISSUE #1463: Exhaustive matches still produce an `error` branch; don't in release mode
In release mode, if a match is exhaustive, then don't emit an `error` branch. This is useful for development (when data structures are being redefined and the like), but in release mode, there should be no such issues.

Easy repro:

```
(pprint-coalton-codegen
  (define-type D (D Integer))
  (define (f (D i)) i))
```

==> (snipped)

```
  (DEFUN F (#:G14)
    (DECLARE (IGNORABLE #:G14))
    (LET ((#:MATCH17 #:G14))
      (DECLARE (IGNORABLE #:MATCH17))
      (LOCALLY
       (DECLARE (SB-EXT:MUFFLE-CONDITIONS SB-EXT:CODE-DELETION-NOTE))
       (COND
        ((AND (TYPEP #:MATCH17 'D/D) T)
         (LET ((I-4291 (D/D-_0 #:MATCH17)))
           (DECLARE (IGNORABLE I-4291))
           I-4291))
        (T (ERROR "Pattern match not exhaustive error"))))))
```

h/t @cdegroot

YarinHeffes:
Closes #1463.

### DIFF
--- a/src/codegen/codegen-expression.lisp
+++ b/src/codegen/codegen-expression.lisp
@@ -226,11 +226,16 @@
                                   ,expr))))))
 
                ;; Only emit a fallback if there is not a catch-all clause.
-               ,@(unless (member-if (lambda (pat)
-                                      (or (pattern-wildcard-p pat)
-                                          (pattern-var-p pat)))
-                                    (node-match-branches expr)
-                                    :key #'match-branch-pattern)
+               ,@(unless (or (member-if (lambda (pat)
+                                          (or (pattern-wildcard-p pat)
+                                              (pattern-var-p pat)))
+                                        (node-match-branches expr)
+                                        :key #'match-branch-pattern)
+                             (and (settings:coalton-release-p)
+                                  (patterns-exhaustive-p
+                                   (mapcar #'match-branch-pattern (node-match-branches expr))
+                                   (node-type (node-match-expr expr))
+                                   env)))
                    `((t
                       (error "Pattern match not exhaustive error")))))))))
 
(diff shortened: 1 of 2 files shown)


stylewarning:
I guess the question is: Do we want to keep the error if the match is non-exhaustive?

YarinHeffes:
I think what we really need is to be able to check if a list of patterns covers every possible case for a particular type.

stylewarning:
@YarinHeffes I am referring to the codegenned `error` call. This `error` call seems to have been added conservatively. It is useful in the following two circumstances:

1. We are in development mode (to cope with redefinitions etc.).
2. We are in any mode, but we know for certain a match is not exhaustive.

The root problem is that we are emitting this `error` branch even when we know the match is exhaustive in `"release"` mode.

YarinHeffes:
> But, doesn't the error from `analyze-translation-unit` make this a non-issue, because we will never codegen a non-exhaustive match?

If the answer is no, then what we can do is add a boolean slot `exhaustivep` to `node-match` that is initialized to `nil` and set to `t` in `analyze-translation-unit`, and that boolean is checked in `codegen-expression`.

stylewarning:
> But, doesn't the error from `analyze-translation-unit` make this a non-issue, because we will never codegen a non-exhaustive match?

My concern is that it's not an error, it's a warning. We could make it an error, but there are valid programs that are not exhaustive. Only incidentally does it get turned into an error by some other process (ASDF, compile-file, etc.).

PR #1487, opened 2025-07-02
Update optional type with a more efficient runtime representation

### LINKED ISSUE #1480: Optimize representation of Option types
We could do it with something like [this](https://www.joachim-breitner.de/blog/787-A_mostly_allocation-free_optional_type). It introduces some complexity, but the benefit would be that most uses of Optional would be effectively free.

YarinHeffes:
Closes #1480.

Based entirely on @stylewarning's sketch here: https://github.com/coalton-lang/coalton/issues/1480#issuecomment-3028152370

As a bonus, the optional is still printable and readable in a convenient way

```lisp
COALTON-USER> (coalton (Some None))
#.(SOME NONE)

COALTON-USER> (coalton (Some 5))
5

                (define x
                  (lisp (Optional (Optional Integer)) () 
                    (cl:read-from-string "#.(SOME NONE)"))))
; No value

COALTON-USER> x
#.(SOME NONE)

COALTON-USER> (type-of 'x)
(OPTIONAL (OPTIONAL INTEGER))

                (define x
                  (lisp (Optional Integer) () 
                    (cl:read-from-string "5"))))
; No value

COALTON-USER> x
5

COALTON-USER> (type-of 'x)
(OPTIONAL INTEGER)
```

### DIFF
--- a/library/classes.lisp
+++ b/library/classes.lisp
@@ -73,11 +73,6 @@
     (first :a)
     (second :b))
 
-  (define-type (Optional :a)
-    "Represents something that may not have a value."
-    (Some :a)
-    None)
-
   (define-type (Result :bad :good)
     "Represents something that may have failed."
     ;; We write (Result :bad :good) instead of (Result :good :bad)
--- a/library/primitive-types.lisp
+++ b/library/primitive-types.lisp
@@ -40,6 +40,44 @@
     (lisp (List :a) ()
       cl:nil))
 
+  ;; Optional is an early type
+  ;;
+  ;; Defining the following is functionally
+  ;; equivalent to having defined an ADT as
+  ;; (define-type (Optional :a)
+  ;;   (Some :a)
+  ;;   None)
+  (inline)
+  (declare Some (:a -> Optional :a))
+  (define (Some x)
+    "A constructor for the type, `Optional`. This constructor can be used
+like any other algebraic data type constructor, including for pattern
+matching, as in the following example.
+
+```lisp
+(match x
+  ((Some value)
+    value)
+  (_ (error \"Oh, no!\")))
+```"
+    (lisp (Optional :a) (x)
+      (coalton-impl/runtime:cl-some x)))
+
+  (declare None (Optional :a))
+  (define None
+    "A constructor for the type, `Optional`. This constructor can be used
+like any other algebraic data type constructor, including for pattern
+matching, as in the following example.
+
+```lisp
+(match x
+  ((None)
+   \"Fantastic!\")
+  (_ (error \"Oh, no!\")))
+```"
+    (lisp (Optional :a) ()
+      coalton-impl/runtime:cl-none))
+
   (repr :native (cl:unsigned-byte 8))
   (define-type U8
     "Unsigned 8-bit integer capable of storing values in `[0, 255]`. Uses `(unsigned-byte 8)`.")
--- a/library/types.lisp
+++ b/library/types.lisp
@@ -105,6 +105,13 @@ The compiler will auto-generate instances of `RuntimeRepr` for all defined types
     (define (runtime-repr _)
       (lisp LispType () 'cl:list)))
 
+  (define-instance (RuntimeRepr (Optional :a))
+    (inline)
+    (define (runtime-repr _)
+      ;; If using `cl:t` proves to be inefficient we could try to
+      ;; improve this, perhaps using proxy-inner.
+      (lisp LispType () 'cl:t)))
+
   ;; The compiler will not auto-generate RuntimeRepr instances for
   ;; types defined in this file to avoid circular dependencies.
   
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -62,7 +62,8 @@
    #:String
    #:Fraction
    #:Arrow
-   #:List #:Cons #:Nil)
+   #:List #:Cons #:Nil
+   #:Optional #:Some #:None)
 
   ;; Primitive Syntax
   (:export
--- a/src/runtime/optional.lisp
+++ b/src/runtime/optional.lisp
@@ -1,0 +1,98 @@
+;;;; optional.lisp
+;;;;
+;;;; This package contains the underlying implementation of the
+;;;; Optional type for Coalton. The idea is to avoid unnecessary
+;;;; boxing by defining a structure that represents None nested in N
+;;;; Somes. Consider the general case of the type (Optional^M :t). In
+;;;; this case, we have exactly M+1 possible cases to match
+;;;; against. We have the cases (Some^N (None)) for N from 0 to M-1
+;;;; and we have the case (Some^M x) where x is of type :t. Therefore,
+;;;; by defining a struct CL-OPTIONAL with one slot, DEPTH, we can
+;;;; always represent (Optional^M :t) as either an element of type :t
+;;;; or as a CL-OPTIONAL with a depth between 0 and M-1.
+;;;;
+;;;; For example, (lisp (Optional (Optional :t)) () 5) is
+;;;; unambiguously equal to (Some (Some 5)).
+
+(defpackage #:coalton-impl/runtime/optional
+  (:use
+   #:cl)
+  (:export
+   #:cl-none
+   #:cl-none-p
+   #:cl-some
+   #:cl-some-p
+   #:unwrap-cl-some))
+
+(in-package #:coalton-impl/runtime/optional)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (declaim (inline make-cl-optional))
+  (defstruct (cl-optional (:constructor make-cl-optional (depth)))
+    "A CL-OPTIONAL represents of DEPTH N represents the Coalton
+value (Some^N None)."
+    (depth 0 :type (unsigned-byte 16) :read-only t))
+
+  ;; SBCL generates the load form by default, but CCL
+  ;; does not, so we generate it here regardless.
+  (defmethod make-load-form ((obj cl-optional) &optional env)
+    (make-load-form-saving-slots obj :environment env)))
+
+#+sbcl
+(cl:declaim (sb-ext:freeze-type cl-optional))
+
+(defmethod print-object ((opt cl-optional) stream)
+  (let ((depth (cl-optional-depth opt)))
+    (format stream "#.")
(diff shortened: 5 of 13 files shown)


library/classes.lisp
```diff
@@ -7,7 +7,6 @@
    #:Signalable
    #:error
    #:Tuple
-   #:Optional #:Some #:None
```
stylewarning, line 10: We might still want to import/re-export these symbols for backward compatibility.

library/primitive-types.lisp
```diff
@@ -40,6 +40,18 @@
     (lisp (List :a) ()
       cl:nil))
 
+  ;; Optional is an early type
```
stylewarning, line 43: Add to comment: "The following act as constructors. It is functionally equivalent to the definition

```
(define-type (Optional :a)
  (Some :a)
  None)
```

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
...
+                 (format nil "(~S ~A)" 'coalton:Some (helper (1- n))))))
+    (format stream "#.~A" (helper (cl-optional-depth opt)))))
+
+(defparameter coalton::cl-none (make-cl-optional 0))
+
+(defun cl-none-p (x)
+  (and (typep x 'cl-optional)
+       (zerop (cl-optional-depth x))))
+
+(defparameter cl-optional-cache
```
stylewarning, line 30: define a global constant

```
(defconstant +optional-cache-size+ 8)
```

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
...
+  (make-array 16 :element-type 'cl-optional
+                 :initial-element coalton::cl-none
+                 :fill-pointer 1))
+
+(defun coalton::cl-some (x)
+  (typecase x
+    (cl-optional
+     (let ((new-depth (1+ (cl-optional-depth x))))
```
stylewarning, line 38: this will simplify to basically

```
(if (< new-depth +optional-cache-size+) (aref +optional-cache+ new-depth) (make-cl-optional new-depth))
```

which should have pretty short disassembly and be OK to inline.

Don't do `length`, `vector-push`, or `ignore-errors` in this crucial function.

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
...
+          (aref cl-optional-cache new-depth))
+         (t
+          (let ((new-cl-optional (make-cl-optional new-depth)))
+            (ignore-errors
+             (vector-push new-cl-optional cl-optional-cache))
+            new-cl-optional)))))
+    (otherwise
+     x)))
+
+(defun cl-some-p (x)
```
stylewarning, line 50: inline

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
...
+  (not (cl-none-p x)))
+
+(defun unbox-cl-some (x)
+  (typecase x
+    (cl-optional
+     (aref cl-optional-cache (1- (cl-optional-depth x))))
```
stylewarning, line 56: similar to before, we'll have an `if (depth < +size+) ...`

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
+;;;; optional.lisp
+
+(defpackage #:coalton-impl/optional
+  (:use
+   #:cl)
+  (:export
+   #:cl-none-p
+   #:cl-some-p
+   #:unbox-cl-some))
```
stylewarning, line 9: unbox -> unwrap

src/optional.lisp
```diff
@@ -0,0 +1,59 @@
...
+(in-package #:coalton-impl/optional)
+
+(declaim (inline make-cl-optional))
+(defstruct (cl-optional (:constructor make-cl-optional (depth)))
+  (depth 0 :type (unsigned-byte 16) :read-only t))
+
+(defmethod print-object ((opt cl-optional) stream)
+  (labels ((helper (n)
+             (if (zerop n)
+                 (format nil "~S" 'coalton::None)
```
stylewarning, line 20: We need to find home packages, ideally not `coalton::` if it can be avoided. In this case, I think you can just print `#:none` directly without a package, since it's just for printing

this printing algorithm is rather slow, fwiw, which probably won't ever be an issue until somebody makes a truly bizarre program

src/package.lisp
```diff
@@ -62,7 +62,8 @@
    #:String
    #:Fraction
    #:Arrow
-   #:List #:Cons #:Nil)
+   #:List #:Cons #:Nil
```
stylewarning, line 65: Ok, this is definitely a reasonable package, given precedent.

src/typechecker/environment.lisp
```diff
@@ -500,6 +512,24 @@
...
+            :docstring "`Some` expresses a value that exists."
+            :compressed-repr 'nil))
+
+          ('coalton:None
+           (make-constructor-entry
+            :name 'coalton:None
+            :arity 0
+            :constructs 'coalton:Optional
+            :classname nil
+            :docstring "`None` expresses that a value does not exist."
```
stylewarning, line 532: "... expresses the absence of a meaningful value."

src/typechecker/environment.lisp
```diff
@@ -500,6 +512,24 @@
...
             :docstring "`Nil` represents an empty `List`."
+            :compressed-repr 'nil))
+
+          ('coalton:Some
+           (make-constructor-entry
+            :name 'coalton:Some
+            :arity 1
+            :constructs 'coalton:Optional
+            :classname nil
+            :docstring "`Some` expresses a value that exists."
```
stylewarning, line 523: "... expresses the presence of a meaningful value."

stylewarning (changes requested):
This is excellent. All of my comments are pretty minor in nature, except for a refactor of the caching logic.

library/primitive-types.lisp
```diff
@@ -40,6 +40,18 @@
...
+  (inline)
+  (declare Some (:a -> Optional :a))
+  (define (Some x)
+    (lisp (Optional :a) (x)
+      (coalton::cl-some x)))
```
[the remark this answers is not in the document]
stylewarning: (The symbols `Optional`, `Some`, and `None` can be in the `coalton` package; but these should be elsewhere, preferably the runtime package if it's not loaded too late.)

stylewarning:
Can we add a couple tests that pattern match?

```
(Some x)
(Some (Some x))
(Some (Some (None))
...
```

src/optional.lisp
YarinHeffes: The purpose of this print algorithm is so that it is both `write` and `read`able and so that it still looks like a Coalton ADT (see the PR comment). But yes, this should be `coalton:None` instead of `coalton::None`.

library/types.lisp
```diff
@@ -105,6 +105,11 @@ The compiler will auto-generate instances of `RuntimeRepr` for all defined types
     (define (runtime-repr _)
       (lisp LispType () 'cl:list)))
 
+  (define-instance (RuntimeRepr (Optional :a))
```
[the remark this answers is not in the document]
YarinHeffes: I could try something like this

```lisp
(define (runtime-repr proxy)
  (let ((inner-type (runtime-repr (proxy-inner proxy))))
    (lisp LispType (inner-type)
      `(cl:or ,inner-type coalton-impl/runtime/optional:cl-optional))))
```

what do you think?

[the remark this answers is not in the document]
stylewarning: We could try this; or maybe we leave it as a TODO for later since (1) we don't have a good performance case for it and (2) we also encode the type in the early types dictionary

src/optional.lisp
stylewarning: The algorithm makes sense, but you can still do it more efficiently.

YarinHeffes: Let me know if it's better now.

src/optional.lisp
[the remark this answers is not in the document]
YarinHeffes: So, I did this but it's giving me trouble with CCL - any advice?

src/runtime/optional.lisp
```diff
@@ -0,0 +1,97 @@
...
+(cl:declaim (sb-ext:freeze-type cl-optional))
+
+(defmethod print-object ((opt cl-optional) stream)
+  (let ((depth (cl-optional-depth opt)))
+    (format stream "#.")
+    (loop :repeat depth :do (format stream "(~S " 'coalton:Some))
+    (format stream "~S" 'coalton:None)
+    (loop :repeat depth :do (format stream ")"))))
+
+;; CCL doesn't recognize MAKE-LOAD-FORM for CL-OPTIONAL when
```
stylewarning, line 46: We can define a load form for it.

```
(defmethod make-load-form ((x cl-optional) &optional env)
  (make-load-form-saving-slots x :environment env))
```

SBCL seems to define this for you; CCL not.

src/runtime/optional.lisp
```diff
@@ -0,0 +1,97 @@
...
+;; evaluating CL-OPTIONAL-CACHE with ALEXANDRIA:DEFINE-CONSTANT, so we
+;; use DEFPARAMETER instead.
+(defmacro define-constant (name value &key (test ''eql))
```
stylewarning, line 49: we should be able to get rid of this

src/runtime/optional.lisp
```diff
@@ -0,0 +1,97 @@
...
+  (declaim (inline make-cl-optional))
+  (defstruct (cl-optional (:constructor make-cl-optional (depth)))
+    "A CL-OPTIONAL represents of DEPTH N represents the Coalton
+value (Some^N None)."
+    (depth 0 :type (unsigned-byte 16) :read-only t)))
+
+#+sbcl
+(cl:declaim (sb-ext:freeze-type cl-optional))
+
+(defmethod print-object ((opt cl-optional) stream)
```
stylewarning, line 39: I wouldn't attempt to fix this, but technically, this doesn't respect some aspects of pretty printing, like the print width and the like.

Don't worry about it though.

stylewarning (changes requested):
Just one quick fix and should be good to go.

src/optional.lisp
stylewarning: It's good. It won't respect certain aspects of the pretty printer, but I think it's fine.

stylewarning (approved):

Issue #1499, opened 2025-07-06
Something sticky causes `limit-unroll-test` fail


shirok:
On SBCL 2.5.6

1. Load Coalton
2. `(asdf:test-system :coalton)` .  This passes.
3. Touch some test file to trigger recompiling, e.g. `touch tests/lisparray-tests.lisp`
4. `(asdf:test-system :coalton)`.  This fails on `LIMIT-UNROLL-TEST`.

Clearing cache and restarting REPL fixes the test failure.

Can anybody reproduce this?

garlic0x1:
I believe recompiling messes with the inliner unrolling, since stored code is already optimized or something.  I can reproduce.

PR #1505, opened 2025-07-10
Fix redefinition of inlinable functions

shirok:
Addresses https://github.com/coalton-lang/coalton/issues/1499

If we're redefining functions that have defined inlinable previously, we shouldn't use the previous definitions to expand the current function body.  It is achieved by dropping previous function definitions.

Redefining inlinable instance methods appears to be immune from the redefinition issue.

### DIFF
--- a/tests/inliner-tests-1.lisp
+++ b/tests/inliner-tests-1.lisp
@@ -8,6 +8,12 @@
 
 (coalton-toplevel
   (inline)
+  (define (test-fact n)
+    (if (== 0 n)
+        2
+        (test-fact (1- n))))
+
+  (inline)
   (define (test-inlinable-rec-1 n)
     (if (== n 0)
         333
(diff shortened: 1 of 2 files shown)


stylewarning:
This approach LGTM.

tests/inliner-tests.lisp
```diff
@@ -90,6 +90,23 @@
 (define-test monomorphize-inline ()
   (is (== 5 (monomorph-for-inline 3))))
 
+;; Issue on redefining inlinable functions
+;; https://github.com/coalton-lang/coalton/issues/1499
+;; These functinos are used by inliner-tests-1.lisp, which is
```
garlic0x1, line 95: s/functinos/functions/

tests/inliner-tests-1.lisp
```diff
@@ -0,0 +1,35 @@
...
+        444
+        (test-inlinable-rec-1 (1- n))))
+
+  (define-instance (RecursiveInlineTestClass Integer)
+    (inline)
+    (define (test-fact-method n)
+      (if (== 0 n)
+          2
+          (test-fact-method (1- n)))))
+  )
```
garlic0x1, line 27: trailing close paren

src/codegen/program.lisp
```diff
@@ -148,7 +157,9 @@ Example:
            (append
             (definition-bindings (tc:translation-unit-definitions translation-unit) env offsets)
             (instance-bindings (tc:translation-unit-instances translation-unit) env offsets)))
-         (definition-names (mapcar #'car definitions)))
+         (definition-names (mapcar #'car definitions))
+         (env (clean-environment-for-redefinition env definitions))
+         )
```
garlic0x1, line 162: trailing close paren

garlic0x1 (commented):
I think this looks correct, just a few minor stylewarnings.
Let me play with it for a bit then I will approve.

src/codegen/program.lisp
```diff
@@ -137,6 +137,15 @@ Example:
               (setf (gethash (car binding) offsets) offset))
         :append instance-bindings))
 
+(defun clean-environment-for-redefinition (env definitions)
+  "If we're redefinining a function that has been declared inlinable, we should remove it from ENV to avoid the code from expanded using old definition."
+  (loop :for (name . _) :in definitions
+        :for fun := (tc:lookup-function env name :no-error t)
+        :when (and fun
```
garlic0x1, line 144: I don't think this will work on functions that are heuristically inlined, maybe we should use some of the predicates in inliner.lisp

stylewarning (approved):

Issue #1490, opened 2025-07-02
Add pattern captures


stylewarning:
Sometimes we want to capture a value in a pattern match. I propose we add `@` syntax for that. Open to other syntax ideas.

```
(match x
  ((Cons (@ x (Some _)) (None))
   ; do something with x, which will be (Some blah)
  )
  (_
   ;otherwise
  ))

PR #1507, opened 2025-07-10
Add: pattern capture language feature

macrologist:
Addresses #1490 

This PR adds a pattern capture syntax to coalton pattern matching.

```
(match (make-list 1 2 3 4 5)
        ;; match cons with 2 in second position
        ;; but bind whole list and tail to vars
        ((@ lst (Cons a (@ tl (Cons 2 _))))
         (== lst (Cons a tl)))
        (_ False))
```

- [x] pattern capture in `match`
- [x] pattern capture in inline `let`
- [x] pattern capture in function `define`s
- [x] tests
- [x] intro to coalton updates


src/analysis/analysis.lisp
```diff
@@ -32,56 +35,66 @@
   "Perform analysis passes on TRANSLATION-UNIT, potentially producing errors or warnings."
   (declare (type tc:translation-unit translation-unit)
            (type tc:environment env))
+  (flet ((analysis-match-traversal (node)
+           (let ((patterns
+                   ;; for exhaustiveness checking, we can collapse
+                   ;; binding patterns, (@ VAR PAT) is part of an
+                   ;; exhaustive match expression iff PAT is. 
+                 (loop :for branch :in (tc:node-match-branches node)
```
stylewarning, line 43: nit: Indentation looks funky here.

src/analysis/pattern-exhaustiveness.lisp
```diff
@@ -19,6 +20,61 @@
 ;;; Maranget 2007, http://moscova.inria.fr/~maranget/papers/warn/index.html
 ;;;
 
+;;; Fundamental concepts for the reader:
```
stylewarning, line 23: Comment: This is great.

src/codegen/translate-expression.lisp
```diff
@@ -934,6 +934,15 @@ Returns a `node'.")
        :type (tc:qualified-ty-type qual-ty)
        :name (tc:pattern-var-name pat))))
 
+  (:method ((pat tc:pattern-binding))
+    (let ((qual-ty (tc:pattern-type pat)))
+      (assert (null (tc:qualified-ty-predicates qual-ty)))
```
stylewarning, line 939: is this a coalton-bug compiler assertion, or an assertion against the user?

tests/pattern-matching-tests.lisp
```diff
@@ -73,3 +73,32 @@
   (is (match #\c
         (#\c True)
         (_ False))))
```
stylewarning, line 75: We need to add a couple parser tests in the list, and show that good error messages happen

stylewarning (changes requested):
One thing I didn't see is that we don't disallow patterns to contain the same variables (i.e., we don't allow [_non-linear patterns_](https://stackoverflow.com/questions/35891663/what-are-nonlinear-patterns)). I think we should disallow this somewhere in the chain of analysis.

Otherwise, this is a fantastic addition and looks very promising.

src/codegen/translate-expression.lisp
macrologist: not sure - very much copy-and-paste to look like all the other methods. I'll try to understand why

src/analysis/analysis.lisp
```diff
@@ -32,56 +35,66 @@
...
-          (tc:make-traverse-block
-           :match (lambda (node)
-                    (let ((patterns (mapcar #'tc:node-match-branch-pattern (tc:node-match-branches node))))
-                      (loop :for pattern :in patterns
-                            :do (check-for-var-matching-constructor pattern env))
+             (let ((exhaustive-or-missing
+                      (find-non-matching-value (mapcar #'list patterns) 1 env)))
+               (unless (eq t exhaustive-or-missing)
+                 (apply #'source:warn "non-exhaustive match"
+                        (cons (source:note node "non-exhaustive match")
```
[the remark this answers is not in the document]
macrologist: I didn't even write this code - just changed the file enough for it to count as mine, evidently. I'll audit for style

src/codegen/pattern.lisp
```diff
@@ -50,6 +55,12 @@
             (:copier nil))
   (name (util:required 'name) :type symbol :read-only t))
 
+(defstruct (pattern-binding
```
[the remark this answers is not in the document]
stylewarning: @macrologist I guess I mean closer to the parser, not in the `codegen.lisp` (though it's welcome there too...)

src/codegen/translate-expression.lisp
macrologist: Most of the methods in `translate-expression` expect the `qualified-ty` instance in their type field to actually be unqualified. Not all of them, but most. The only genuine exception is in the translation of `node-var`, where `apply-dicts` is called, where the node unqualified concrete type is extracted and applied to the returned node. 

If we're going to yank out a concrete type from a `qualified-ty`, we want to make sure that there aren't any dangling predicates that we've are yet to contend with.

garlic0x1:
This seems great, just played with it and had no issues

stylewarning (approved):

PR #1526, opened 2025-07-16
Rename ord-map:Map to ordmap:OrdMap etc. for the consistency

shirok:
Now that we have HashMap, this PR renames Map based on red-black-tree to OrdMap, an ordered map.  The package names ord-map and ord-tree are also renamed to ordmap and ordtree, to be consistent of hashmap.   (An alternative choice is to keep them intact and use hash-map; I'm open to either way).

The plan was to proceed making HashMap and OrdMap APIs symmetric.  However, it is better be achieved with typeclass (e.g. `(define-class (Mapping :key :value) ...)`).  Since the collection API discussion is ongoing, I split just the renaming part to this PR, for it still touches lots of places.


stylewarning:
@shirok CI is still borked.

shirok:
> @shirok CI is still borked.

It works now.

stylewarning (approved):

Issue #1395, opened 2025-02-24
Misleading error for over-applied arithmetic operations.


YarinHeffes:
The following code applies three arguments to `max`, which is defined for two arguments.
```lisp
(coalton (max 1.0 2.0 3.0)))
```
The error message is misleading.
```
Failed to unify types SINGLE-FLOAT and (#T39871 → #T39872)
   [Condition of type COALTON-IMPL/TYPECHECKER/TYPE-ERRORS:UNIFICATION-ERROR]
```
The same error is emitted for similar functions, including `+`.

---

This does not apply in general. For example, the following code applies three arguments to `append`.
```lisp
(coalton (append nil nil nil))
```
The error message is clear.
```
error: Argument error
  --> <macroexpansion>:1:9
   |
 1 |  (COALTON (APPEND NIL NIL NIL))
   |           ^^^^^^^^^^^^^^^^^^^^ Function call has 3 arguments but inferred type '((LIST :A) → (LIST :A) → (LIST :A))' only takes 2
   [Condition of type COALTON-IMPL/TYPECHECKER/BASE:TC-ERROR]
```

---

Recommendations:
- [ ] fix error messages for over-applied arithmetic operations.
- [ ] extend unification errors to refer to source code, including line number(s).

PR #1533, opened 2025-07-23
ensure proper error message for overapplied functions

YarinHeffes:
This PR begins to address #1395.

The error message for `(max 1.0 2.0 3.0)` is fixed.

However, the error message for cases like `(max 1 2 3)` for which the types of the arguments are not fully resolved at the point at which the arity is checked are not fixed.  We check arity by reading an expression such as `(max 1.0 2.0 3.0)` as `((((max) 1.0) 2.0) 3.0)` and checking after each application if we have a function type. The issue remains that we cannot determine if `(max 1 2)` is a function type at this point, since it just has type `Num :a => :a` and `:a` which may be a function type.

This will be more complicated. It would be resolved by simply removing this block
https://github.com/coalton-lang/coalton/blob/main/src/typechecker/define.lisp#L374-L393

But I am not sure what adverse consequences my arise thereafter. It will require further investigation.

### DIFF
--- a/src/typechecker/define.lisp
+++ b/src/typechecker/define.lisp
@@ -366,7 +366,7 @@ Returns (VALUES INFERRED-TYPE PREDICATES NODE SUBSTITUTIONS)")
                                    (setf preds (append preds preds_))
                                    (setf accessors (append accessors accessors_))
                                    (setf subs subs_)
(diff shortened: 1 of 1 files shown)


stylewarning (approved):

stylewarning:
@YarinHeffes can you update the issue with the newest info?

Issue #1534, opened 2025-07-24
Storing unexpanded instance context


garlic0x1:
# Problem

The new context expander for class instances loses information about
the original constraints specified by the programmer.  This is
apparent when the auto-generated documentation is affected:
https://github.com/coalton-lang/coalton-website/pull/277/files

The change is "correct" but it is not reflecting exactly what the
programmer wrote, and this makes it suboptimal.

# Solution

To avoid mutating the environment when expanding constraints, as well
as keep the original context, we have two options:

1. After creating instances, update the environment with expanded
   context in another slot, preserving the original user-defined
   context.

   Pros: Anyone browsing the type environment can see both original
   and expanded context

   Cons: Swapping instances in the environment is complicated and
   expensive, since it involves finding things in the environment (by
   codegen symbol?) and indexing them, then swapping the new instance
   into multiple locations.

2. Create a new accessor that performs expansion when other parts of
   the system read the context.

   Pros: Simple solution, should only require replacing a few accessor
   calls with one that expands.  No mutations or swaps of the environment 
   are needed.

   Cons: Expansion might be done multiple times on an instance, but
   can be more easily solved with a mutable caching slot than by
   treating the whole environment as immutable.

# Conclusion

The reason why this is all necessary is because expanding mutually
recursive constraints requires all dependencies to be in the
environment already.

I think the second solution is more appealing, partly because it would be 
significantly easier to implement, but also I think it would be a lot easier 
to read later on.

I welcome any input to help me decide what to do.

stylewarning:
I agree that a complete re-update of the environment is probably not the solution we are looking for.

Where are all the places the expanded context is needed now?

PR #1535, opened 2025-07-24
Make instance constraints immutable and expand them on read

garlic0x1:
Addresses https://github.com/coalton-lang/coalton/issues/1534 using the second proposed solution (without any caching implemented)


stylewarning (approved):
LGTM

PR #1537, opened 2025-07-24
Derive Hash

garlic0x1:
I chose to not allow deriving `Hash` if there are constructors without fields, since I noticed that `Optional` does not implement `Hash`, but I'm not sure if this is a good idea


stylewarning:
Constructors without fields should have hashes. Optional was an oversight.

garlic0x1:
Implemented and tested, each constructer combines a hash of ctor index, this could be optimized away for types where this isn't required though

library/derivers.lisp
```diff
@@ -120,3 +120,77 @@
...
+                     :name 'x
+                     :orig-name 'x))
+      :body (parser:make-node-body
+             :nodes nil
+             :last-node (parser:make-node-match
+                         :location location
+                         :expr (parser:make-node-variable
+                                :location location
+                                :name 'x)
+                         :branches (mapcar
```
stylewarning, line 143: this style is inconsistent and difficult to read: MAPCAR over a LOOP-COLLECT is kind of the worst of both worlds in my opinion. Can we just collapse this all into a simple loop?

    (loop :for index :from 0
          :for ctor :in ...
          ...
          :collect ...)

stylewarning (changes requested):
This mostly looks good. Requesting to
simplify the mapcar-lambda-reduce-lambda-loop amalgamation a bit.

A comment on whether the hash function is stable under redefinition would be good. Hashing on index for example is unstable under adding fields. Hashing on ctor name/package is stable on adding fields, but not on renaming fields. etc. (In general I would try to make definitions robust to typical incremental development redefinitions as much as possible.)

A comment on the shape of the hash function generated would be appreciated.

library/derivers.lisp
```diff
@@ -120,3 +120,93 @@
                                                              :name 'coalton:False))))))))
       :location location
       :inline nil))))
+
+(defmethod tc:derive-methods ((class (eql 'classes:hash)) def env)
```
stylewarning, line 124: Typo in docstring: "define" -> "redefine"

I don't think this is an ideal hash function. Somehow it should take into account the type itself (so as to hash differently for same-shaped types). I would advocate also dispensing with the index and statically hashing the symbol name+package. Concretely

```
(defun hash-symbol (s)
  mix hashes of symbol name and symbol package)
```

Then for a type like

```
(define-type A
  ...
  (Ak f1 ... fn)
  ...)
```

the hash of an Ak object would be a mix of:

- (hash-symbol A), computed statically at compile time
- (hash-symbol Ak), computed statically at compile time
- hash of f1 to fn, computed dynamically.

This will be no less efficient than the index approach, but be far more robust to (1) hash collisions from same-shaped types, and (2) be invariant to adding, removing, or permuting fields.

stylewarning (changes requested):

library/derivers.lisp
garlic0x1: I agree hashing symbols is better, but why would we care about collisions between different types?

library/derivers.lisp
```diff
@@ -120,3 +120,101 @@
                                                              :name 'coalton:False))))))))
       :location location
       :inline nil))))
+
+(defun hash-symbol (sym)
```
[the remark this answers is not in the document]
stylewarning: probably yes

library/derivers.lisp
```diff
@@ -120,3 +120,101 @@
...
+      :body (parser:make-node-body
+             :nodes nil
+             :last-node (parser:make-node-match
+                         :location location
+                         :expr (parser:make-node-variable
+                                :location location
+                                :name 'x)
+                         :branches (loop :for index :from 0
+                                         :for ctor :in (parser:type-definition-ctors def)
+                                         :for cfields := (mapcar (lambda (_)
```
stylewarning, line 173: i would lift

    :for patterns := ...

here too

stylewarning (changes requested):
just a small style note pending further changes

library/derivers.lisp
```diff
@@ -120,3 +120,101 @@
...
+                                                                                  :rator (parser:make-node-variable
+                                                                                          :location location
+                                                                                          :name 'classes:hash)
+                                                                                  :rands (list
+                                                                                          (parser:make-node-variable
+                                                                                           :location location
+                                                                                           :name cfield)))
+                                                                                 acc)))
+                                                                      cfields
+                                                                      :initial-value (parser:make-node-lisp
```
stylewarning, line 210: hash this statically into a node-integer-literal; no need to make it an ast and runtime computation

stylewarning (changes requested):

library/derivers.lisp
garlic0x1: It isn't a runtime computation, I am using a `node-lisp` in order to make the type `Hash`, which I couldn't figure out how to do with `node-integer-literal`.

Codegen output:

```lisp
(COALTON-LIBRARY/HASH:COMBINE-HASHES
            (COMMON-LISP:LET ((#:ITEM-6879084 #:|ctor-field9071|))
              (COMMON-LISP:DECLARE (COMMON-LISP:IGNORABLE #:ITEM-6879084))
              (COMMON-LISP:LOCALLY
               (COMMON-LISP:DECLARE
                (COMMON-LISP:OPTIMIZE (SB-C::TYPE-CHECK 1)))
               (COMMON-LISP:LET ((COALTON-LIBRARY/MATH/HASH-DEFINING-MACROS::ITEM
                                  #:ITEM-6879084))
                 (COMMON-LISP:VALUES
                  (COMMON-LISP:SXHASH
                   COALTON-LIBRARY/MATH/HASH-DEFINING-MACROS::ITEM)))))
            (COMMON-LISP:LOCALLY
             (COMMON-LISP:DECLARE (COMMON-LISP:OPTIMIZE (SB-C::TYPE-CHECK 1)))
             (COMMON-LISP:PROGN (COMMON-LISP:VALUES 617796614386979468))))
```

library/derivers.lisp
```diff
@@ -120,3 +120,101 @@
...
+                                                                                      :location location
+                                                                                      :type (parser:make-tycon :location location :name 'classes:hash)
+                                                                                      :vars '()
+                                                                                      :var-names '()
+                                                                                      :body (list
+                                                                                             (hash-symbol
```
stylewarning, line 216: Can we mix the hash of the type's name as well?

PR #1540, opened 2025-07-25
Use (derive Eq) in standard library

garlic0x1:


library/ordtree.lisp
```diff
@@ -41,6 +41,7 @@
 
   ;; unexported; a marker held by trees to enable self-balancing.
   (repr :enum)
+  (derive Eq)
```
stylewarning, line 44: Point for future: (repr :enum) can be compared with `cl:eq`. Would be nice in the future for the deriver to know about this.

stylewarning (approved):
See note about future extension of the deriver. Maybe you can write an issue for it? It's low priority.

PR #1546, opened 2025-07-29
Fix declare faux-macro to optimize for CL editing

garlic0x1:
Reverts this commit: https://github.com/coalton-lang/coalton/commit/39ca00c38e48ab02237fbc9eb768610b0eccb019

Making `declare` use `&body` improves indentation in Coalton type declarations, but makes it worse for CL declarations.  In my opinion it is more annoying than it is helpful, and I would like to undo this change.


garlic0x1:
May require recompiling emacs config to take effect

stylewarning (approved):

Issue #1549, opened 2025-08-03
Internal coalton bug: Unknown binding MAKE-LIST


peterlopen:
as suggested by error message:

```Internal coalton bug: Unknown binding MAKE-LIST

If you are seeing this, please file an issue on Github.
   [Condition of type COALTON-IMPL/UTIL:COALTON-BUG]

Restarts:
 0: [RETRY] Retry SLIME REPL evaluation request.
 1: [*ABORT] Return to SLIME's top level.
 2: [ABORT] abort thread (#<THREAD tid=16895 "new-repl-thread" RUNNING {1002518223}>)

Backtrace:
  0: (COALTON-IMPL/UTIL:COALTON-BUG "Unknown binding ~S" MAKE-LIST)
  1: (COALTON-IMPL/TYPECHECKER/ENVIRONMENT:LOOKUP-VALUE-TYPE #<COALTON-IMPL/TYPECHECKER/ENVIRONMENT:ENVIRONMENT {1001CB79C3}> MAKE-LIST :NO-ERROR COMMON-LISP:NIL)
  2: (SB-INT:SIMPLE-EVAL-IN-LEXENV (TYPE-OF (COMMON-LISP:QUOTE MAKE-LIST)) #<NULL-LEXENV>)
  3: (COMMON-LISP:EVAL (TYPE-OF (COMMON-LISP:QUOTE MAKE-LIST)))
  4: (SWANK::EVAL-REGION "(type-of 'make-list) ..)
```

when (type-of 'make-list) evaluated.

PR #1550, opened 2025-08-03
Check that symbol doesn't name a macro in type-of

macrologist:
Addresses #1549 

Or something like this.

@garlic0x1 @stylewarning @YarinHeffes


macrologist:
Possibly it should return void and and maybe print something

YarinHeffes (changes requested):

macrologist:
feel free to steal my thunder

Issue #1488, opened 2025-07-02
Consider renaming `Ord` to `Ordering` and class `Complex :t` to `ComplexComponent :t`


stylewarning:
Consider these renamings to make things less confusing.

- [ ] Rename type `Ord`
- [x] Rename type class `Complex`

PR #1552, opened 2025-08-05
rename Complex type class to ComplexComponent

stylewarning:
Partially addresses #1488.

This PR also removes some instances that don't seem useful, like `ComplexComponent (Complex :t)` etc.


YarinHeffes:
I think tests that make use of the `Complex` type class need to be updated as well

Issue #1572, opened 2025-08-07
More tail calls involving class instances not being eliminated [fixed in upstream SBCL]


stylewarning:
`(foo (the Integer 0))` will blow the stack in the following example:

```
(coalton-toplevel
  (define-class (C :a)
    (p (:a -> :a)))

  (define-instance (C Integer)
    (define (p x)
      (if (< x 1000000000)
          (foo x)
          x)))

  (define (foo x)
    (p (1+ x))))
```

It seems this is because of similar reasons to #1341 and #1397.

(Reported by @garlic0x1.)

stylewarning:
The tail calls are definitely all there even without monomorphization. If we take the generated FTYPE declaims out of codegen, everything works. So the FTYPEs are the culprit.

stylewarning:
@YarinHeffes The reason monomorphize "works" is because more precise types are being emitted: there is no "more specific" function calling a "more general" (but perhaps not actually at runtime) necessitating a return value type check. That is,

    F : Specific -> Specific (returns G)
    G : Specific -> General (returns F)

The call to G in F will need a type check to ensure G's "General" is actually F's "Specific". Monomorphization will make a G*

    G* : Specific -> Specific

which allows SBCL to statically eliminate the type check.

This issue isn't actually relevant to instance methods. It will also happen with toplevel functions matching the above pattern.

YarinHeffes:
I'm on the same page as you regarding the example you shared that doesn't include any methods, just functions.

But in the case of instance methods, the general function is defined, for example, like this.
```lisp
(DEFUN BAR (COALTON-IMPL/CODEGEN/CODEGEN-CLASS::DICT |_0|)
     (DECLARE (OPTIMIZE (SPEED 3) (SAFETY 0)))
     (CALL-COALTON-FUNCTION
      (CLASS/BAR-BAR COALTON-IMPL/CODEGEN/CODEGEN-CLASS::DICT) |_0|))
```
The absence of `foo` in the definition of `bar` is what I was pointing out.

stylewarning:
Yes, it is "trampolining". It will still be a tail call, just more circuitous. 🫠

Also, SBCL folks committed a fix for the tail calls [here](https://github.com/sbcl/sbcl/commit/e242d3d5af5b5d5d9179e56124af9af58c005a35). I have not tested this fix works yet.

We should add a unit test at some point for this as well.

YarinHeffes:
[...]

PR #1578, opened 2025-08-07
Add a test for a non-trivial tail-call elimination

YarinHeffes:
See issue #1572.

**[DO NOT MERGE]** 

- [x] Awaiting official release of SBCL >= 2.5.7.38. The CI runner will need to be updated accordingly.


tests/runtime-tests.lisp
```diff
@@ -206,3 +206,39 @@
...
+
+(coalton-toplevel
+
+  (define-class (Bar1 :t)
+    (bar1 (:t -> :t)))
+
+  (define-instance (Bar1 UFix)
+    (define (bar1 x)
+      (if (< x 1000000000)
+          (foo x)
```
stylewarning, line 225: typo: should be foo1

tests/runtime-tests.lisp
```diff
@@ -206,3 +206,39 @@
...
+      (if (< x 1000000000)
+          (foo1 x)
+          x)))
+
+  (define (foo1 x)
+    (bar1 (1+ x))))
+
+(cl:in-package #:coalton-native-tests)
+
+#+SBCL
```
stylewarning, line 239: I think we should enable this for all platforms. Not sure if CCL will fail or not. We should see.

stylewarning (changes requested):

tests/runtime-tests.lisp
YarinHeffes: Looks like CCL had no trouble.

stylewarning (approved):
Cool that CCL passes.

Add a comment above the test that it's expected to fail on SBCL <=2.5.7.

This PR can be merged when CI has updated SBCL (which can happen when SBCL is released). Then the tests should pass.

PR #1585, opened 2025-08-09
Do not declare dynamic extent when a match expression is captured

YarinHeffes:
Addresses cases like the following, for which the match expression should not be declared to have dynamic extent.
```lisp
(match (Tuple x y)
  ((= v (Tuple x y))
   (print x)
   (print y)
   v))
```


stylewarning:
good catch

Issue #1588, opened 2025-08-09
Multiple functional dependency bugs


YarinHeffes:
This issue collects multiple bugs related to functional dependencies.

- [ ] Issue no. 1.
The following example should compile, but it does not.
```lisp
;; Functional dependencies are not inherited as they should be.
(define-class (C :a :b (:a -> :b)))
(define-class (C :a :b => D :a :b)
  (m :a))
```

- [ ] Issue no. 2. (Taken from #1050.)
The following example should compile, but it does not.
```lisp
(define-class (C :a :b (:a -> :b))
  (cm (:a -> UFix)))
(define-class (C :a :b => D :a :b (:a -> :b))
  (dm (:a -> UFix -> :b)))

;; "Missing inferred predicate (C :a :c)"
(declare f (D :a :b => :a -> :b))
(define (f x)
  (dm x (cm x)))
```

- [ ] Issue no. 3.
The following example compiles, but it should not.
```lisp
(define-class (C :a :b (:a -> :b)))
(define-instance (C :a :b))
```

YarinHeffes:
Fixed in #1607 and #1608.

PR #1608, opened 2025-08-12
Fix functional dependencies

YarinHeffes:
This PR addresses the remaining three cases outlined in #1588.

There may be additional cases for which the support of functional dependencies fails. I can address them as they arise, but in the meantime, my aim with this PR is to provide support for the common/essential uses of functional dependencies without introduces new bugs.


stylewarning (approved):
I think we need more tests that have longer subclass chains, and we need a bunch of documentation.

Issue #1502, opened 2025-07-07
Make-over for compiler internals


macrologist:
Compilers, more than most other kinds of software, should be well-documented and approachable. 

Concepts that provide context to the operation of a module should be documented, whenever possible with figures and examples, at the top of the module. 

All exported functions that are not auto-generated (e.g. by `defstruct`) should have meaningful docstrings. The same goes for exported types -- each one should have a meaningful docstring. By "meaningful", I mean descriptive beyond a repetition of its name. 

I.e. there are many docstrings like this:

    (defun thing-getter (x)
        "gets x's thing" 
         ...)

You should not be able to compress the docstring to the name of the thing it documents.

For example, the above could be improved:

    (defun thing-getter (x)
        "Get's x's thing. X is a Y that is used to do Z. 
         The thing of x is a doohickey. 
         See docstring for doohikey and Y."
        ...)

- [ ] meaningful docstrings for every exported function and type
- [ ] meaningful context provided as preamble to /most/ modules, suffiecient to explain the module's purpose
- [ ] declares for every exported function

PR #1627, opened 2025-08-29
Some notes on AST and Environment

jbouwman:
This is a contribution towards https://github.com/coalton-lang/coalton/issues/1502 with some notes on typechecking environment and roles of the dual ASTs (parser, code generator) -- it would benefit from review for terminological correctness.

Also, there is an incompatible change to the return value of tarjar-scc: almost invariably, the returned SCC list is immediately reversed, since it's so strongly bound up with compilation dependency ordering. I found this confusing on first seeing it: my feeling is that it's more intuitive to return values in least-to-most dependent order.

Also, changed 'dag' to 'graph' in the SCC routing, since that was also a bit of a red herring.

### DIFF
--- a/src/algorithm/tarjan-scc.lisp
+++ b/src/algorithm/tarjan-scc.lisp
@@ -1,37 +1,92 @@
 (defpackage #:coalton-impl/algorithm/tarjan-scc
   (:use #:cl)
   (:export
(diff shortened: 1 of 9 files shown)


macrologist (approved):
Magnifique!

stylewarning (approved):
excellent

PR #1628, opened 2025-08-29
Add `string:upcase` and `string:downcase` functions to the standard library

garlic0x1:
These simply call `cl:string-upcase` and `cl:string-downcase` respectively.  I am thinking it might also be worth adding a function that wraps `cl:string-equal` for case-insensitive equality testing, but I'm not sure what it should be named.

### DIFF
--- a/library/string.lisp
+++ b/library/string.lisp
@@ -25,7 +25,9 @@
    #:ref-unchecked
    #:substring-index
    #:substring?
-   #:chars))
+   #:chars
+   #:upcase
+   #:downcase))
 
 
 (in-package #:coalton-library/string)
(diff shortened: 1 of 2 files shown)


shirok:
Scheme names it `string-ci=`.

How comprehensive do we want to be?  Case conversion is very interesting topic, when Unicode is involved.  We can either do "the right thing", or just be pragmatic and piggyback on the underlying CL implementation.

garlic0x1:
I would assume these functions to behave the same way as the CL equivalents.  I'm not sure what edge cases exist, but it might be worth doing more here.

shirok:
There are several dozens of characters of which case mapping changes number of codepoints.  Looks like SBCL hasn't been handling them.

stylewarning:
for this PR i'm fine to mimic whatever lisp does

other points are valid

PR #1641, opened 2025-09-09
add Show class and friends

stylewarning:
This is to scratch an itch of being able to print stuff, superseding `trace` and friends.


library/show.lisp
```diff
@@ -0,0 +1,214 @@
...
+      (cl:let ((*show-stream* s))
+        (call-coalton-function f Unit)))))
+
+(coalton-toplevel
+  (define-class (Show :a)
+    "Objects which have a convenient, textual, linear printed representation for
+display in a terminal. This is principally for program output and debugging."
+    (show-to
+     "Display an object to a Lisp stream."
+     (LispOutputStream -> :a -> Unit)))
```
Jason94: Maybe I'm missing something obvious, but couldn't we just do it the way Haskell does it and do the simpler:
```lisp
(define-class (Show :a)
    (show (:a -> String)))
```
Then we could have functions keyed off that which output to streams, etc.

I'm not saying we should always just do things the same way Haskell does them. But in this case what is the benefit of the complexity? And it's not just Haskell that does this. Every other language besides CL I can think of has its override-able "convert this object into a string" function morally equivalent to (Mytype -> String).

shirok: One possible concern is that Coalton is eager, so when you `show` huge structure you create all the temporary strings before they are output?

shirok: Gauche has rope-like library that allows O(1) string concatenation.  It is rudimental and trivial, so any library author can hack up its own, but I found it pretty handy to have it in the bundled library, as other libraries can use it as the common interface.  If we can invent the standard library, it may be an option.

library/types.lisp
```diff
@@ -40,17 +45,29 @@
...
+  (declare proxy-function-from (Proxy (:a -> :b) -> Proxy :a))
+  (define (proxy-function-from _)
+    "Get a `Proxy` representing the \"domain\" of a function type."
+    Proxy)
+
+  (inline)
+  (declare proxy-function-to (Proxy (:a -> :b) -> Proxy :b))
+  (define (proxy-function-to _)
+    "Get a `Proxy` representing the \"codomain\" of a function type."
+    Proxy)
```
YarinHeffes: What's your opinion on using the following instead?

```suggestion
  (inline)
  (declare proxy-left (Proxy (:T :a :b) -> Proxy :a))
  (define (proxy-left _)
    "Get a `Proxy` representing the \"domain\" of a relational type."
    Proxy)

  (inline)
  (declare proxy-right (Proxy (:T :a :b) -> Proxy :b))
  (define (proxy-right _)
    "Get a `Proxy` representing the \"codomain\" of a relational type."
    Proxy)
```

PR #1644, opened 2025-09-12
New Vector Operations

garlic0x1:
`vec:subseq` is the same as `string:substr`, and `vec:kill!` kills a range in a vector.

### DIFF
--- a/library/vector.lisp
+++ b/library/vector.lisp
@@ -147,7 +147,7 @@
   (inline)
   (declare kill! (Vector :a -> UFix -> UFix -> Vector :a))
   (define (kill! v start end)
-    "Destructively kills subsequence in a vector bounded by given indices."
+    "Destructively kills a subsequence in a vector bounded by given indices."
     (let ((real-start (max 0 (min start end)))
           (real-end (min (length v) (max start end)))
           (new-size (- (length v) (- real-end real-start))))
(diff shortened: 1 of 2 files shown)


stylewarning (approved):

library/vector.lisp
```diff
@@ -18,11 +18,13 @@
    #:with-initial-element
    #:singleton
    #:length
+   #:subseq
    #:capacity
    #:empty?
    #:singleton?
    #:copy
    #:set-capacity!
+   #:kill!
```
stylewarning, line 27: i prefer something like `resect!`. What do you think?

library/vector.lisp
```diff
@@ -94,6 +96,16 @@
     (lisp UFix (v)
       (cl:length v)))
 
+  (inline)
+  (declare subseq (Vector :a -> UFix -> UFix -> Vector :a))
+  (define (subseq v start end)
+    "Compute a subseq of a vector bounded by given indices."
```
Jason94, line 102: Could you clarify whether the `end` index is inclusive/exclusive in the docstring?

stylewarning: also just use cl:subseq in implementation

garlic0x1: I can't figure out how to make `cl:subseq` create an adjustable vector.

PR #1650, opened 2025-09-13
Nits for `string:substring`

garlic0x1:
from #1644

### DIFF
--- a/library/string.lisp
+++ b/library/string.lisp
@@ -69,8 +69,10 @@
 
   (declare substring (String -> UFix -> UFix -> String))
   (define (substring str start end)
-    "Compute a substring of a string bounded by given indices."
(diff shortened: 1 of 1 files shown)


library/string.lisp
```diff
@@ -69,8 +69,10 @@
 
   (declare substring (String -> UFix -> UFix -> String))
   (define (substring str start end)
-    "Compute a substring of a string bounded by given indices."
-    (let ((real-start (max 0 (min start end)))
+    "Compute a substring of a string bounded by given indices.
+
+`start` index is inclusive and `end` index is exclusive."
```
shirok, line 74: If you update the docstring, how about mentioning that
- if start > end, it takes from end below start
- if end is greater than the length of str, the result is taken up to the end of str.
It saves users to read the code for the boundary cases.

Issue #1580, opened 2025-08-09
emit `case` when matching on `(repr :enum)`


stylewarning:
This should generate a `case` not a `cond`
```
                (repr :enum)
                (define-type Color R G B)
                (define (blah x)
                  (match x
                    ((R) R)
                    ((G) G)
                    ((B) B)))
```

Care must be taken if the patterns have captures, wildcards, etc.

See also: #1323

garlic0x1:
Should I rewrite this PR? https://github.com/coalton-lang/coalton/pull/1399

stylewarning:
> Should I rewrite this PR? https://github.com/coalton-lang/coalton/pull/1399

I think yes, doing one optimization at a time. One match optimization will land soon as soon as it's reviewed.

PR #1653, opened 2025-09-17
Emit `case` when matching on `(repr :enum)`

garlic0x1:
[!1580](https://github.com/coalton-lang/coalton/issues/1580)

- New package and file for functions to help with match codegen.
- Use a `cl:case` jumptable on enums.
- Improved handling of `Boolean` matches.


src/codegen/codegen-match.lisp
```diff
@@ -0,0 +1,320 @@
...
+      ((and (equalp true-pattern (ast:match-branch-pattern (second branches)))
+            (equalp false-pattern (ast:match-branch-pattern (first branches))))
+       (values t
+               (ast:match-branch-body (second branches))
+               (ast:match-branch-body (first branches))))
+
+      ;; Case #4:
+      ;;
+      ;; Do not emit `cl:if' unless exhaustively matching boolean with
+      ;; constructor pattern branches.
```
stylewarning, line 154: add a TODO

```
TODO: Deal with one branch + fallback.
```

We can still emit `CL:IF` when we match against a single boolean with a fallback.

src/codegen/codegen-match.lisp
```diff
@@ -0,0 +1,320 @@
...
+           (type symbol match-var)
+           (type tc:ty match-expr-type)
+           (type tc:environment env)
+           (values t &optional))
+
+  (multiple-value-bind (pred bindings types)
+      (codegen-pattern:codegen-pattern pattern match-var match-expr-type env)
+
+    `(,pred
+      ,(if (null bindings)
```
stylewarning, line 190: nit: `endp`

src/codegen/codegen-match.lisp
```diff
@@ -0,0 +1,320 @@
...
+
+  (multiple-value-bind (pred bindings types)
+      (codegen-pattern:codegen-pattern pattern match-var match-expr-type env)
+    (declare (ignore pred))
+
+    (cond
+      ;; Case #1:
+      ;;
+      ;; Emit a list of a single literal pattern to match on.
+      ((pattern:pattern-literal-p pattern)
```
stylewarning, line 213: I haven't looked so carefully. If we have a string literal, will this proc?

```
(case x
  (("string") ...))
```

won't work, because strings are not `eql`.

I think the answer is "no" because `case` should only work with enumeration types in this PR. But then I wouldn't expect a `pattern-literal` to exist. So maybe that should become an `assert`.

src/codegen/codegen-match.lisp
```diff
@@ -0,0 +1,320 @@
...
+  "Generate code for a match branch."
+  (declare (type t code)
+           (type pattern:pattern pattern)
+           (type symbol match-var)
+           (type tc:ty match-expr-type)
+           (type tc:environment env)
+           (type boolean jumptablep)
+           (values t &optional))
+
+  (funcall (if jumptablep
```
stylewarning, line 245: nit: I generally think it's not good style to bury the `if`, because `(funcall (if x a b) y)` cannot be optimized as easily as `(if x (a y) (b y))`. Here it's not a big deal.

src/codegen/codegen-match.lisp
```diff
@@ -0,0 +1,320 @@
...
+               #'codegen-case-match-branch
+               #'codegen-cond-match-branch)
+           code
+           pattern
+           match-var
+           match-expr-type
+           env))
+
+(defun codegen-cond-match (branches fallbackp)
```
stylewarning, line 254: In these two functions, can you document `fallbackp`? It's not clear if `fallbackp` means "generate a fallback" or "assume branches includes a fallback".

stylewarning (changes requested):
small requests, mostly documentation. others are nits.

Issue #1663, opened 2025-09-30
add debug-type intrinsic


stylewarning:
add debug-type function that prints the type of the contained expression at compile time

PR #1664, opened 2025-09-30
Crude debug-type intrinsic

garlic0x1:
https://github.com/coalton-lang/coalton/issues/1663

Since intrinsics get transformed at codegen-time, I am not sure how to resolve tyvars to a nice informative message.  I might have run intrinsics at different stages of compilation. and specialize the methods on AST type, where the current inliner intrinsics specialize on codegen nodes, and this specializes on a typechecker node.

```lisp
                (declare foo (Into :a String => :a -> UFix -> String))
                (define (foo str x)
                  (debug-type 
                   (string:substring (into (debug-type str)) 0 x))))
;; #T34326
;; STRING
; No values
```

### DIFF
--- a/src/codegen/intrinsic-applications.lisp
+++ b/src/codegen/intrinsic-applications.lisp
@@ -2,6 +2,7 @@
   (:use #:cl)
   (:local-nicknames
    (#:ast #:coalton-impl/codegen/ast)
+   (#:tc #:coalton-impl/typechecker)
    (#:traverse #:coalton-impl/codegen/traverse))
   (:export
    #:transform-intrinsic-application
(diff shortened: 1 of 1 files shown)


stylewarning:
there's a function somewhere to pretty print tyvars or types as a whole

src/codegen/intrinsic-applications.lisp
```diff
@@ -58,6 +59,19 @@ Returns a new `ast:node'."))
       (t
        child))))
 
+(defmethod transform-intrinsic-application ((rator-name (eql 'coalton:debug-type)) node)
```
stylewarning, line 62: This is probably too late for the current implementation of intrinsic applications, but we really want the `qualified-type` from the type checker AST in `coalton-impl/typechecker/expression`.

stylewarning (changes requested):

PR #1672, opened 2025-10-16
ADT named pattern-matching

fukamachi:
This allows pattern-matching with field names.

ref https://github.com/coalton-lang/coalton/discussions/1671

## Example

```common-lisp
(define-struct Point (x F32) (y F32))

;; Can add types with (.name Type)
(define-type Shape
  (Circle (.radius F32) (.center Point))
  (Square (.side F32) (.center Point))
  (Triangle Point Point Point))    ;; Can mix positional definitions

(define shape (Circle 5.0 (Point 10.0 2.5)))

;; Positional matching (backward compatible)
(match shape
  ((Circle _ center) center))
;=> #.(POINT 10.0 2.5)

;; Named pattern-matching
(match shape
  ((Circle .radius) radius))
;=> 5.0

### DIFF
--- a/src/typechecker/define-type.lisp
+++ b/src/typechecker/define-type.lisp
@@ -502,7 +502,7 @@
                     := (typecase ctor
                          (parser:constructor (parser:constructor-field-names ctor))
                          (parser:toplevel-define-struct nil)
-                         (t nil))
+                         (otherwise nil))
                   :collect (tc:make-constructor-entry
                             :name ctor-name
                             :arity (length (parser:type-definition-ctor-field-types ctor))

### CODE REVIEW

File: src/codegen/pattern.lisp [hunk 1]
```diff
@@ -73,8 +74,9 @@
 (defstruct (pattern-constructor
             (:include pattern)
             (:copier nil))
-  (name     (util:required 'name)     :type symbol       :read-only t)
-  (patterns (util:required 'patterns) :type pattern-list :read-only t))
+  (name        (util:required 'name)     :type symbol                       :read-only t)
+  (patterns    (util:required 'patterns) :type pattern-list                 :read-only t)
+  (field-names nil                       :type (or null (cons symbol list)) :read-only t))
```
stylewarning: i would just make a new `symbol-list` type if it doesn't exist already

src/typechecker/define.lisp
```diff
@@ -1738,11 +1738,29 @@ Returns (VALUES INFERRED-TYPE NODE SUBSTITUTIONS)")
...
+      (let* ((arity (tc:constructor-entry-arity ctor))
+             (pat-field-names (parser:pattern-constructor-field-names pat))
+             (ctor-field-names (tc:constructor-entry-field-names ctor))
+             (num-args (length (parser:pattern-constructor-patterns pat))))
+
+        ;; Handle named field patterns
+        (when pat-field-names
+          ;; Validate: constructor must have named fields
+          (unless ctor-field-names
+            (tc-error "Named field pattern on positional constructor"
```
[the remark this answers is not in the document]
stylewarning: These kinds of errors are also where we would naturally want lexical (line/col) info.

stylewarning (changes requested):
some initial comments

tests/adt-accessor-tests.lisp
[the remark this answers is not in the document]
YarinHeffes: Take a look at this folder specifically for testing error messages.

https://github.com/coalton-lang/coalton/tree/main/tests/test-files

src/parser/toplevel.lisp
```diff
@@ -1873,6 +1875,73 @@ consume all attributes")))
    :name (cst:raw form)
    :location (form-location source form)))
 
+(defun parse-constructor-fields (unparsed-fields source enclosing-form)
+  "Parse constructor fields, detecting dot-prefix named fields.
+Returns (values parsed-types field-names) where field-names is NIL for positional
+or a list of strings for named fields."
```
YarinHeffes: I'm curious why we disallow users from naming some but not all slots. I understand that it might be a rare case, but perhaps we should support it anyway?

src/parser/toplevel.lisp
```diff
@@ -1873,6 +1875,73 @@ consume all attributes")))
...
+  (declare (type list unparsed-fields)
+           (values list (or null util:string-list) &optional))
+
+  (when (null unparsed-fields)
```
YarinHeffes: nit
```suggestion
  (when (endp unparsed-fields)
```

macrologist (commented):
I noticed that I cannot do stuff like this:

```lisp 

  (define-type (Point :num)
    (P2D (.x :num) (.y :num))
    (P3D (.x :num) (.y :num) (.z :num)))

(define special-x (.x (P2D 10 20))
```

It errors with:

```
The value
  #.(P2D 10 20)
is not of type
  NAMED-FIELDS-FUN::POINT/P3D
```

There are two mutually exclusive ways to "fix" this:

1. If the behavior *should* be illegal, then it ought to be caught at compile time, when compiling the define-type.
2. But I think that the behavior should actually be allowed. In this case we need to allow the same name to appear across constructors. 

For example, the same kind of thing with `define-struct`s works:

``` lisp
  (define-struct (Point2d :num)
    (x :num)
    (y :num))

  (define-struct (Point3d :num)
    (x :num)
    (y :num)
    (z :num))


  (define special-x (.x (Point2d 10 20)))
```

Somehow, the accessors need to be keyed by both type and the constructor

src/typechecker/environment.lisp
```diff
@@ -462,6 +464,10 @@
   (classname       (util:required 'classname)       :type symbol                         :read-only t)
   (docstring       (util:required 'docstring)       :type (or string null)               :read-only t)
 
+  ;; Field names for constructors with named fields (using dot-prefix syntax)
+  ;; NIL for positional constructors, (cons string list) for named field constructors
+  (field-names     nil                              :type (or null (cons string list))   :read-only t)
```
garlic0x1: I think we can use `coalton-impl/util:string-list`

garlic0x1 (commented):
Sorry, just spamming a type suggestion, seems like a good idea but I have yet to experiment with it.

stylewarning:
`(.x f)` could be defined as

```
(match f
  ((2d .x) x)
  ((3d .x) x))
```

or whatever equivalent.

fukamachi:
How is it about `.z`?

macrologist:
I know people do not like this kind of thing, but what if ADT named fields generated accessor functions like `P2d.x` and the like?

Alternatively, in lieu of finding a good solution, I also think it is acceptable to require all slot names across type constructors to be unique.  Something like:

``` lisp

(define-type Point
  (Pt2 (.x2 F32) (.y2 F32))
  (Pt3 (.x3 F32) (.y3 F32) (.z3 F32)))

```

This is a little inelegant, but it may be the best we can do. Also, if somebody finds a good way to lift the restriction, the above code will still work.

src/parser/pattern.lisp
```diff
@@ -101,8 +102,77 @@ runtime value and binds it to a variable."
 (defstruct (pattern-constructor
             (:include pattern)
             (:copier nil))
-  (name     (util:required 'name)     :type identifier           :read-only t)
-  (patterns (util:required 'patterns) :type pattern-list         :read-only t))
+  (name        (util:required 'name)     :type identifier                 :read-only t)
+  (patterns    (util:required 'patterns) :type pattern-list               :read-only t)
+  (field-names nil                       :type (or null util:string-list) :read-only t))
```
garlic0x1: don't need null on these
```lisp
CL-USER> (typep nil 'coalton-impl/util:string-list)
T
```

macrologist:
> Or, it may be better to change the return value to Optional for partial accessors.

I would favor this over runtime errors

PR #1681, opened 2025-10-31
Raise an error for incompatible function redefinitions

fukamachi:
When a function is redefined with an incompatible type during interactive development, existing functions that call it break.

To prevent this, the system now raises an error when there are dependent functions, clearly showing all affected callers.

Screenshot 2025-10-31 at 11 29 34


stylewarning:
How about a restart and a global setting to allow, while still printing some kind of warning?

PR #1701, opened 2025-12-01
Implement Collections API for just List and Vector

Jason94:
Inspired by #1700, I'm taking a _much_ smaller bite of the elephant from #1337. This PR _just_ creates the new versions of `list.lisp` and `vector.lisp`. It doesn't start creating the typeclasses or rewrite anything else. I'm going to leave that PR up for reference, but probably the best plan is just to keep breaking it up into smaller pieces. Those can come later.

Because this doesn't have the typeclasses yet, there's some duplication that we can remove later, but I think this is the cleanest starting point.

### DIFF
--- a/library/list.lisp
+++ b/library/list.lisp
@@ -1,3 +1,9 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; DEPRECATED ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; This package is deprecated, replaced by:
+;;; coalton-library/collections/immutable/list
+;;; Please edit that file instead!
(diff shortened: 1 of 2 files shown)


Jason94:
@YarinHeffes & @stylewarning

library/collections/immutable/list.lisp
```diff
@@ -0,0 +1,858 @@
...
+  
+  (declare subseq (UFix -> UFix -> List :a -> List :a))
+  (define (subseq start end lst)
+    "Get the subsequence of LST from [start, end)."
+    (if (> start end)
+      Nil
+      (take (- end start) (drop start lst))))
+
+  (declare split-at ((UFix -> List :a -> Tuple (List :a) (List :a))))
+  (define (split-at i lst)
```
Jason94: TODO: align with https://github.com/coalton-lang/coalton/pull/1707#discussion_r2595810251

PR #1727, opened 2025-12-19
Add HTML backend for docgen

Jason94:
The CSS isn't a complete match to the standard library, but it's pretty close.

Example output here:
https://jason94.github.io/coalton-io/


Jason94:
For some reason the standard lib docs build is failing on the CI, but it's working for me locally. I'll dig into that tomorrow.

stylewarning:
Cool!

stylewarning (approved):

PR #1728, opened 2025-12-19
stdlib-p would error on uninterned symbols

stylewarning:


stylewarning:
potential fix for #1727

PR #1748, opened 2026-02-06
Fix Seq concat rebalancing crash near 1024 boundary

stylewarning:
Seq was broken around 1024 elements. This is a fix that I think works, with a regression test, but it might be worth double checking.

### DIFF
--- a/library/seq.lisp
+++ b/library/seq.lisp
@@ -367,8 +367,8 @@ to `len`."
                 ret))
             (iter:into-iter subtrees)))))
 
-  (declare shift-n-onto! (vector:Vector :a -> vector:Vector :a -> UFix -> Unit))
-  (define (shift-n-onto! target source n0)
+  (declare %shift-n-onto! (vector:Vector :a -> vector:Vector :a -> UFix -> Unit))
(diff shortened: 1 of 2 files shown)


stylewarning:
@shirok review appreciated!

library/seq.lisp
```diff
@@ -515,7 +515,9 @@ It attempts to rebalance with a minimum of array copying."
       ;;make a node from rebalanced branches
       (make-node-from-branches
        (iter:collect!
-        (iter:filter! (compose (< 0) size)
+        ;; `size` can observe transiently invalid branches during rebalancing.
```
shirok: I believe this fix works.   It would be helpful to note in `%shift-n-branches-onto` that it may make `seq2` invalid (not only "dirty", but the state that `size` cannot be applied).

Btw, `shift-n-onto!` is only used by `%shift-n-branches-onto`, so isn't it better to name it something like `%shift-n-onto!` to indicate it's internal?  Or, make it internal function of `%shift-n-branches-onto`.  It is another layer that obscures the fact that the operation may create an invalid node.

shirok (approved):
Fix is ok.  Just a general comment.

PR #1749, opened 2026-02-06
codegen: reuse analysis exhaustiveness checker for fallback matching

### LINKED ISSUE #1730: `TVar` error shows up in release mode but not development mode
(Reported by @Jason94.)

I'm getting a strange error when I try to compile a file in release mode. In dev mode, it compiles and runs fine. Here is the error:

```
; compiling file "/home/jason/programming/workspace/coalton-simple-io/examples/channels-threading.lisp" (written 19 DEC 2025 11:45:43 AM):
; 
; caught COMMON-LISP:ERROR:
;   COMMON-LISP:READ error during COMMON-LISP:COMPILE-FILE:
;   
;     The value
;       #T71411
;     is not of type
;       COALTON-IMPL/TYPECHECKER/TYPES:TYCON
;   
;     (in form starting at line: 34, column: 0, position: 1005)
; 
; compilation unit aborted
;   caught 1 fatal ERROR condition
;   caught 1 ERROR condition
```
[This is the file](https://github.com/Jason94/coalton-io/blob/concurrent/examples/channels-threading.lisp#L102) in question. L102 seems to be the offending line. This compiles:
```lisp
  (declare sum-file (IO Integer))
  (define sum-file
    (do
     write-data-file
     (_input-chan <- mv:new-empty-chan)
     (_ints-chan <- mv:new-empty-chan)
     (write-line "Forking threads...")
     ;; (fork-thread (reader-thread input-chan))
     ;; ...
     (pure 0)))
```
But with 102 uncommented, it fails:
```lisp
  (declare sum-file (IO Integer))
  (define sum-file
    (do
     write-data-file
     (_input-chan <- mv:new-empty-chan)
     (_ints-chan <- mv:new-empty-chan)
     (write-line "Forking threads...")
     (fork-thread (reader-thread _input-chan))
     ;; ...
     (pure 0)))
```
(And again, that compiles just fine in development mode. The whole thing does.

I'm *pretty sure* this is a bug, because if it compiles in dev mode it should be compiling in release mode. Plus it shouldn't leak TVars.

The steps I took switching between modes:

1. Switch mode in my init
2. Run `rm **/*.fasl` in my repo root
3. Delete `~/.cache/common_lisp` to force
4. Restart my REPL

stylewarning:
The codegen fallback exhaustiveness pass had a separate implementation with different assumptions than analysis-time checking. In particular, it assumed constructor-only matches imply a `TYCON`/`TAPP` scrutinee head. In release codegen this pass can still see `TYVAR` scrutinees, which caused the #1730 failure (`... not of type ... TYCON`) and made behavior diverge from the earlier analysis pass.

Replace the codegen-specific exhaustiveness logic with the analysis checker:
- convert codegen pattern structs into typechecker pattern structs
- collapse binding patterns before checking
- delegate to `analysis-pattern-exhaustiveness:exhaustive-patterns-p`

This removes duplicate logic and keeps fallback emission decisions consistent with analysis semantics.

Add codegen exhaustiveness regression tests for:
- tyvar-head constructor patterns do not crash
- tyvar-head boolean constructor signatures are recognized
- binding patterns are collapsed consistently

Also register `tests/codegen-pattern-tests.lisp` in `coalton.asd`.

Fixes #1730.

### DIFF
--- a/coalton.asd
+++ b/coalton.asd
@@ -294,6 +294,7 @@
                 :serial t
                 :components ((:file "cursor-tests")))
                (:file "entry-tests")
+               (:file "codegen-pattern-tests")
                (:file "toplevel-tests")
                (:file "type-inference-tests")
                (:file "fundep-tests")
--- a/src/codegen/pattern.lisp
+++ b/src/codegen/pattern.lisp
@@ -2,6 +2,7 @@
   (:use
    #:cl)
   (:local-nicknames
+   (#:pe #:coalton-impl/analysis/pattern-exhaustiveness)
    (#:util #:coalton-impl/util)
    (#:tc #:coalton-impl/typechecker))
   (:export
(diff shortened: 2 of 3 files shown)


stylewarning:
Going to merge just to truck ahead. Please review if you don't mind nonetheless.

PR #1760, opened 2026-02-09
Compatibility layer in small pieces

c-kloukinas:
Created a minimal compatibility layer with 4 "sections" (marked with ;;;):

- hashing (should be generic),
- unsetting float traps (or sbcl allegro ccl abcl ecl),
- get bytes consed (or sbcl abcl), and
- package locks (or sbcl ecl)

Each of these has been added in a separate commit, so hopefully it's easier to see what changed each time.

The package locks have been applied only to library/system.lisp so as not to touch all files that (try to) use this functionality.

I've kept out of the compatibility layer macros that have an implementation only on sbcl (as far as I can tell), opting to include those that have an implementation in at least another lisp (even if not currently supported, like ecl & abcl), since there's a higher probability that another lisp may also offer something that can be used there.

I've tested it locally, with my own fork of fset. Sbcl passed all tests.


c-kloukinas:
Testing failed because it tried to use the official repo images, that are not accessible to me (are they supposed to be readable by non-members of coalton-lang?).

When run with my images, [all tests pass](https://github.com/c-kloukinas/coalton-compat/actions/runs/23485798426).

PR #1775, opened 2026-02-17
Compat build n test

c-kloukinas:
This PR tries to:
1) simulate the github test actions in compat/build-test-configuration.lisp, and keeping the fasls of each individual test configuration in a separate directory, so that they don't clash.
2) provide a quick abcl-specific patch for fiasco's define-test-package
3) update the Makefile so one can use more lisps, control where they'll store their quicklisp packages, install these, test all configurations for a specific lisp, clean up everything.


c-kloukinas:
Here I've learned to stop worrying and love the agent of chaos (aka, QL) - this helped discover this [issue with benchmarking](https://github.com/coalton-lang/coalton/issues/1776).

PR #1777, opened 2026-02-21
Add #T tuple shorthand (for both types and values)

stylewarning:
I didn't overhaul the standard library, and I also didn't add any notes to the README. That may cause some inadvertent confusion, but I want this feature to be exercised before going whole-hog on refactoring.

Syntax is only supported in the Coalton readtable.


stylewarning:
I won't rush to merge this, and I'm open to other syntax ideas or directions.

PR #1779, opened 2026-03-05
Code in package coalton-impl/runtime/function-entry but defines functions in (locked) package coalton.

c-kloukinas:
sbcl gave me an error about trying to modify locked package coalton (configuration: release-3-1-1-sbcl-2.6.1-linux-x64).

Issue seems to go away with definitions in an unlocked coalton package.


c-kloukinas:
Issue could have been the result of an unclear ~/.cache.

PR #1868, opened 2026-04-05
separate result packs from value variables

### LINKED ISSUE #1866: Bug: (pure) Captures Void Values
`(pure (values))` or its equivalents shouldn't be allowed.

Currently, this is compiling when it shouldn't:

```lisp
  (declare void-identity (Void -> Identity Void))
  (define (void-identity)
    (pure (values)))

  (declare foo (Void -> Void))
  (define (foo)
    (run-identity
     (do
      (pure (values))
      (void-identity)
      (a <- (void-identity))
      (let _ = (traceobject "a:" a))
      (pure 1)))
    (values))
``` 
At runtime, it assigns `nil` to `a`.

stylewarning:
Track whether a type variable may unify with result packs and preserve that across parsing, quantification, and instantiation.

Reject Void and multi-value packs in ordinary value positions, and move the regression coverage into file-based compiler tests.

fix #1866

### DIFF
--- a/examples/small-coalton-programs/src/brainfold.ct
+++ b/examples/small-coalton-programs/src/brainfold.ct
@@ -172,9 +172,9 @@
     "Takes and stores a character as an ascii code at the pointer."
     (do
      (bfs <- state:get)
-     (pure (vec:set! (cell:read (.pointer bfs))
-                     (into (char:char-code (prompt-char)))
-                     (.memory bfs)))
+     (let _ = (vec:set! (cell:read (.pointer bfs))
+                        (into (char:char-code (prompt-char)))
+                        (.memory bfs)))
       (state:put bfs))))
 
 ;;;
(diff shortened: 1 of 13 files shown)


Jason94:
Yes, that seems to have done it.

Compiling this:
```lisp
  (declare void-identity (Void -> Identity Void))
  (define (void-identity)
    (pure (values)))
```
gives this error:
```
 2 |    (DECLARE VOID-IDENTITY (VOID -> IDENTITY VOID))
   |                                             ^^^^ Void and multi-value types cannot be used as type arguments
```

Compiling this:
```lisp
  ;; (declare void-identity (Void -> Identity Void))
  (define (void-identity)
    (pure (values)))
```
gives this error:
```
 3 |      (PURE (VALUES))))
   |            ^^^^^^^^ Expected type ':A' but got 'Void'
```

Compiling this:
```lisp
  (define (foo)
    (run-identity
     (do
      (pure (values))
      (pure 1)))
    (values))
```
gives this error:
```
 5 |         (PURE (VALUES))
   |               ^^^^^^^^ Expected type ':A' but got 'Void'
```

PR #1919, opened 2026-04-28
Add or-def utility function

Jason94:
Adds a small utility function. Mirrors `ok-or-def` from Result, but with a shorter name.

### DIFF
--- a/library/optional.ct
+++ b/library/optional.ct
@@ -6,7 +6,8 @@
   (:export
    #:from-some
    #:some?
-   #:none?))
+   #:none?
+   #:or-def))
 
 (in-package #:coalton/optional)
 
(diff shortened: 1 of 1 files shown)


stylewarning:
What about WITH-DEFAULT? https://coalton-lang.github.io/reference/#coalton-classes-with-default-value

Jason94:
> What about WITH-DEFAULT? https://coalton-lang.github.io/reference/#coalton-classes-with-default-value

Oh, yeah that would work. One of the downsides of not having the Optional type in the https://coalton-lang.github.io/reference/#coalton-optional-package docs is you can't see all the instances!

I guess the only benefit of my version is it would be a teeensy bit more efficient because it avoids a few function calls. Feel free to close if it's not worth it. (Very well may not be worth it.)

stylewarning:
If it's renamed to FROM-SOME-{OR,ELSE} we can add it and inline it.

PR #1923, opened 2026-04-28
mine: support TIOCGWINSZ on OpenBSD

rpx99:
This enables the existing Unix terminal size ioctl path on OpenBSD.

`mine/src/bindings/platform-unix.lisp` already uses `TIOCGWINSZ` on Darwin and Linux to query terminal rows and columns via `ioctl`. OpenBSD uses the same BSD-style `TIOCGWINSZ` request value, so this change extends that path to OpenBSD as well.

I intentionally kept this patch narrow to the platform I tested, rather than broadening it to other BSDs in the same change.


stylewarning (approved):

stylewarning:
Thank you!

PR #1929, opened 2026-04-29
Add a link to awesome-coalton to the readme

Jason94:
Does what it says on the tin. I'm open to the readme not being the right place for this - maybe the website?


stylewarning (approved):

stylewarning:
Readme is fine. We could also put it in the manual page of the website.

PR #1944, opened 2026-05-01
Inline several functions in the functions package

Jason94:


library/functions.ct
```diff
@@ -68,6 +68,7 @@ Print a line to `cl:*standard-output*` in the form \"{STR}: {ITEM}\"."
   ;; Function combinators
   ;;
 
+  (inline)
```
stylewarning: I wouldn't inline this.

stylewarning (changes requested):

PR #1951, opened 2026-05-01
Expose `mine-core` in the Nix flake

Xophmeister:
WIP...

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -85,6 +85,44 @@
               value = recipe.${impl}.mainLib;
             }
           ];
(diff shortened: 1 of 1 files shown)


Xophmeister:
@stylewarning @tani: This is a first approximation of exposing `mine-core` via the Nix Flake. I'm not great at Nix, so I've worked on it with Claude Code until I got it working; there may be more optimal solutions. What do you think, so far?

`mine-app` looks much more complex to package, with the Tauri app, etc. so I'll probably just leave it with core for now, in this PR. If you think this looks good, I'll write up some documentation.

PR #1981, opened 2026-05-09
Add statically typed format function

Jason94:
Examples:

```lisp
(coalton (format Out "~a" (make-list 1 2 3)))
#<List [1 2 3]>

(coalton (format Out "~(~{~c~^ ~}~)!~%" (make-list #\H #\E #\L #\L #\O)))
h e l l o!

(coalton (format Out "~{~{~a~^ ~}~^|~}~%" (make-list
                                           (make-list 1 2 3)
                                           (make-list 10 20 30)
                                           (make-list -1 -2 -3))))
1 2 3|10 20 30|-1 -2 -3

;    1 |  (coalton (format Out "~(~{~c~^ ~}~)!~%" (make-list 1 2 3)))
;      |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expression has ambiguous type Num Char ⇒ String
;      |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Add a type assertion with THE to resolve ambiguity

;    1 |  (coalton (format Out "~a ~(~{~c~^ ~}~)!~%" (make-list 1 2 3)))
;      |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Parse Error: Not enough arguments provided for control string
```

---
(taken from the header in the file)
This file provides a statically-typed wrapper around the Common Lisp format function.
At compile time, the validity of the control string, the number of arguments, and
the types of the arguments are checked. At runtime, Common Lisp's format receives
the same control string and, for the most part, the same arguments. The exception is
~a, which uses `show-with-string` first.

The code implements a monadic parser to parse the control string and compile Coalton
forms that typecheck/transform the arguments. The file is broken up as follows:

0. Build a small monadic parser library. Aslo build a TResult type that's used to
   separate recoverable and unrecoverable errors while parsing.

1. Lex the plain control string into string segments and raw directives, which
   just contain character data internally. Validates that the individual components
   of the control string follow the grammar, but not that they are real directives,
   etc.

2. Parse the raw data into a control string AST, which represents each format
   directive as a separate ADT constructor. Validates that individual directives
   are well-formed, for example, that the directive character exists.

3. Compile the control string AST into a series of argument code transformation
   steps. This is the step that determines that ~C needs to correspond to a Char
   argument. Validates inter-directive properties, such as that a ~{ needs to be
   followed by a ~}.

4. Iterate over the provided arguments' code in the format macro and transform it
   based on the previously compiled transformation steps.
---

Known limitations:
1. There are a few directives I didn't implement because it's hard to write the algorithm to determine the number of args and get the types to line up right. I think that ~[ ~] could be implemented with the limitation that each branch needs to process the args in the same way, and ~* might be implementable (or it could be equivalent to the halting problem idk).
2. Iteration is more restricted in this version. But nested iteration is doable, as shown above.
3. I'm using `Show` for `~a`. Having worked on this and thought about this more, I agree that we should follow the Common Lisp split of printable and readable formats. `#<List [1 2 3]>` might be fine, but `"string"` is a bit annoying with format. In any case, that can come later, and we can always make a custom typeclass for ~a if we want.


Jason94:
It seems to be blowing the heap when trying to compile. It works locally, so I'll clear my fasl cache and investigate.

Update: It's still running locally. There was a platform-specific character rendering issue in the tests that I fixed, and now most of the tests are passing. Some of the runs are still blowing the heap. It says "0 bytes available" here:

```
[package coalton/format]..........................
..................................................
Heap exhausted during garbage collection: 0 bytes available, 32784 requested.
     | Immobile Objects |
 Gen layout symbol   code  Boxed   Cons    Raw   Code  SmMix  Mixed  LgRaw LgCode  LgMix Waste%       Alloc        Trig   Dirty GCs Mem-age
  1      0      0      0    114     29  33605      0      3     19      0      0      0   48.3   571777760   343405428   33770   1  1.3432
  2     72     55    981   4568   1038  24212      2    109    113     20      0    174   38.6   608450512    21474836   24630   0  0.8457
  3    956   2657  12319    444    247    254     10     67     91      0      0     21    3.5    35873216     2000000     324   0  0.0000
  4      0      0      0      0      0      0      0      0      0      0      0      0    0.0           0     2000000       0   0  0.0000
  5      0      0      0      0      0      0      0      0      0      0      0      0    0.0           0     2000000       0   0  0.0000
  6    769  23343  19621    223    114      7      3     11     12      0      0      9    3.5    11990576     2000000      19   0  0.0000
Tot   1797  26055  32921   5349   1428  58078     15    190    235     20      0    204   42.8  1228090448 [57.2% of 2147483648 max]
```
So maybe that means that something on GitHub messed up or ran out of space? I could be interpreting it wrong. I'll try again tomorrow.

I might also need to back off the inlining.

Jason94:
> It seems to be blowing the heap when trying to compile. It works locally, so I'll clear my fasl cache and investigate.
> 
> Update: It's still running locally. There was a platform-specific character rendering issue in the tests that I fixed, and now most of the tests are passing. Some of the runs are still blowing the heap. It says "0 bytes available" here:
> 
> ```
> [package coalton/format]..........................
> ..................................................
> Heap exhausted during garbage collection: 0 bytes available, 32784 requested.
>      | Immobile Objects |
>  Gen layout symbol   code  Boxed   Cons    Raw   Code  SmMix  Mixed  LgRaw LgCode  LgMix Waste%       Alloc        Trig   Dirty GCs Mem-age
>   1      0      0      0    114     29  33605      0      3     19      0      0      0   48.3   571777760   343405428   33770   1  1.3432
>   2     72     55    981   4568   1038  24212      2    109    113     20      0    174   38.6   608450512    21474836   24630   0  0.8457
>   3    956   2657  12319    444    247    254     10     67     91      0      0     21    3.5    35873216     2000000     324   0  0.0000
>   4      0      0      0      0      0      0      0      0      0      0      0      0    0.0           0     2000000       0   0  0.0000
>   5      0      0      0      0      0      0      0      0      0      0      0      0    0.0           0     2000000       0   0  0.0000
>   6    769  23343  19621    223    114      7      3     11     12      0      0      9    3.5    11990576     2000000      19   0  0.0000
> Tot   1797  26055  32921   5349   1428  58078     15    190    235     20      0    204   42.8  1228090448 [57.2% of 2147483648 max]
> ```
> 
> So maybe that means that something on GitHub messed up or ran out of space? I could be interpreting it wrong. I'll try again tomorrow.
> 
> I might also need to back off the inlining.

Ok, I simplified the code a good bit and removed some inlining. (I took advantage of the fact that you don't need the string segment information in between the directives to either typecheck the arguments or validate the control string, so I removed them earlier on in the parsing.) I also added some tests to verify some failures, like having mismatched `~{` and `~}` directives.

Should be ready for review now!

stylewarning:
I'll try to work through this weekend!

Jason94:
> I'll try to work through this weekend!

Thanks! Definitely take a look at the types. Right now I'm using the `Num` class for arguments that can be any number. But this is probably wrong, because then a caller can wrap their own types with `Num`, which will fail when `cl:format` tries to use them at runtime. Probably the right thing there is to make a similar `NativeNum` typeclass that works just like `NativeInt` does.

---

An alternative is to actually provide real typeclasses for this. Something like:
```lisp
(repr :native (cl:or cl:single-float ...))
(define-type NativeNum)

(define-class (FormatNum :a)
  (format-as-native-num
    (:a -> NativeNum)))
```
or something like that. This is the approach Haskell's PrintF package uses. My concern here was runtime costs. I suppose that if the native num types had an inlined identity type function, that should have zero overhead cost. And this is what I'm doing with `~a` for show, which means that in `~{ ~}` clauses, it actually maps over the list with `show-as-string` first.

If it wasn't for iteration, the latter approach would probably be better. printf doesn't have iteration (honestly iteration in a control string is wild but I digress). The problem is, even if it was some smart identity function like I describe above, if you do this:
`(format Out "~{~F~}" (make-list 0.1 0.2 0.3))`
it still has to map over the list and allocate a new one first.

Possibly you could get around this by modifying the codegen so it passes the mapping code it uses a toplevel function instead of an inline map call like this:
```lisp
(declare to-native-num-list (FormatNum :a => List :a -> List NativeNum))
(define (to-native-num-list lst)
   (map format-as-native-num lst))
```

Then for all of Coalton's builtin number types, you could specialize `to-native-num-list` to just return the list. I think that would work for single lists. My concern would be for double-iteration, it might not be possible to prevent it mapping over the outer list in something like `List (List F32))` or whatever.

---

All of that to say, there are three approaches to directives that can take args of multiple types:
1. Use other classes, like I'm currently doing for some of them with Num. No performance tradeoff, but unsafe and not extensible.
2. Use new, private classes, like I'm currently doing for some of them with IntAtRuntime, which isn't exported so it can only ever have the builtin Coalton primitives that are valid arguments to those format directives. Safe and no performance tradeoff, but not extensible.
3. Use  new, open classes, like Haskell's printF package. Safe and extensible, but more complex and probably still a performance tradeoff for-at the very least-nested iterations.

library/format.ct
```diff
@@ -0,0 +1,1208 @@
...
+  "Format CTRL-STR using ARGS. Syntax is the same as Common Lisp format. The number of
+arguments, their types, and the syntax of CTRL-STR are checked at compile time.
+Directives are case-insensitive.
+
+TARGET can be:
+  - Str, returns a string
+  - Out, prints to *STANDARD-OUTPUT*
+  - any String or Lisp Stream value
+
+The following directives are supported with the given argument requirements:
```
stylewarning, line 1156: Format as Markdown with code backticks and the like.

library/format.ct
```diff
@@ -0,0 +1,1208 @@
...
+(coalton-toplevel
+  (define Str
+    (lisp (-> sy:Symbol) ()
+      'Str))
+  (define Out
+    (lisp (-> sy:Symbol) ()
+      'Out)))
+               
+(defmacro format (target ctrl-str cl:&rest args)
+  "Format CTRL-STR using ARGS. Syntax is the same as Common Lisp format. The number of
```
stylewarning: Use markdown convention (not Lisp convention) for docstring arguments.

```
"Format `ctrl-str` using `args`. ...
```

library/format.ct
```diff
@@ -0,0 +1,1208 @@
...
+            (new-args (transform-format-args! ctrl-str args))
+            (arg-syms (cl:mapcar (cl:lambda (_)
+                                   (cl:declare (cl:ignore _))
+                                   (cl:gensym "format-arg"))
+                                 new-args))
+            (let-forms (cl:loop :for sym :in arg-syms
+                                :for arg-form :in new-args
+                                :collect `(,sym ,arg-form))))
+    `(let (,@let-forms)
+       (lisp (-> String) (,@arg-syms)
```
stylewarning: If `target` is `cl:t` then `cl:format` returns `cl:nil`.

library/format.ct
stylewarning: Use a Markdown table if you must.

stylewarning (changes requested):
initial comments

library/format.ct
```diff
@@ -0,0 +1,1208 @@
...
+             (c:update! (fmap (fn (c_)
+                                (lisp (-> Code) (c_)
+                                  `(req-num-list ,c_))))
+                        code))
+           (when req-runtime?
+             (c:update! (fmap (fn (c_)
+                                (lisp (-> Code) (c_)
+                                  `(req-runtime-list ,c_))))
+                        code))
+           ;; For primitive types, those should look the same with ~a anyway
```
Jason94, line 1017: It's probably worth adding that without this check against other primitive types, this would break:
`~{ ~a ~f ~}`

Jason94: (adding to the comment)

library/format.ct
```diff
@@ -0,0 +1,1208 @@
...
+  - U32
+  - U64
+  - UFix
+
+The following directives are not supported:
+
+  ~[
+  ~]
+  ~;
+  ~*
```
Jason94: `~*` can be trivially supported within iteration clauses. That's nice to have (i guess) because you can show the same element in multiple formats in some kind of table by moving back one between directives:

`~{~d~t~1:*~o~}`

note: make sure to implement `~*` with a minimum & max of 1 arg

Jason94: Oh, and because ~* only works from its param, not reading from an argument to format, it's definitely statically determinable. I don't think I'm going to do it for this PR, but it is doable.

library/classes.ct
```diff
@@ -241,6 +243,26 @@ The name `fmap` comes from Haskell."
...
     (liftA2 ((:a * :b -> :c) * (:f :a) * (:f :b) -> (:f :c))))
 
+  (inline)
+  (declare take-left (Applicative :f => :f :a * :f :b -> :f :a))
+  (define (take-left fa fb)
+    "Evaluate two applicatives, and take the value returned by the left one.
+Useful when the right applicative could produce a meaningful structural
+change that would alter the computation, such as returning a `None`.
+
+Equivalent of Haskell's <*"
```
stylewarning: backticks around Haskellism

library/classes.ct
```diff
@@ -241,6 +243,26 @@ The name `fmap` comes from Haskell."
...
+    (lifta2 ƒa_.a fa fb))
+
+  (inline)
+  (declare take-right (Applicative :f => :f :a * :f :b -> :f :b))
+  (define (take-right fa fb)
+    "Evaluate two applicatives, and take the value returned by the right one.
+Useful when the left applicative could produce a meaningful structural
+change that would alter the computation, such as returning a `None`.
+
+Equivalent of Haskell's *>"
```
stylewarning: backticks around Haskellism

stylewarning (changes requested):

stylewarning (approved):

PR #2001, opened 2026-05-13
unboxed product types for direct calls/returns

stylewarning:


andyleclair:
I'm seeing this when I try to compile the raytracer on this branch

image

This is the definition of `refract`:

image

https://tangled.org/andyleclair.dev/ray/blob/main/src/vec3.ct#L72

PR #2025, opened 2026-05-24
Change String Show instance to output string

Jason94:
This PR changes the `(Show String)` instance to simply output the string.

Practically, the current definition of `(Show String)` makes it difficult to use in cases where you need to rely on nesting calls to `show-as-string`. It tends to return noisy output. Here are some examples:

```lisp:
(coalton
 (show
  (show-as-string "hi")))
;; => "\"hi\""

(coalton
 (format Out
         "~a ~a"
         (format Str "~a" "hi")
         10))
 ;; => #A((4) BASE-CHAR . "\"hi\"") 10
```

Theoretically, there have been several conversations about the proper role of `Show` and what it should be doing. It seems like the agreed long-term design to move to is to use `Show` to produce user-readable output, and to make another typeclass to create Common Lisp `read`able output. This is the definition of `(Show String)` that fits that role.

### DIFF
--- a/library/show.ct
+++ b/library/show.ct
@@ -117,6 +117,11 @@ This is not necessarily identical to `(the String (into x))`."
     (define (show-to f x)
       (f (if x "True" "False"))))
 
+  (define-instance (Show String)
+    (inline)
+    (define (show-to f s)
+      (f s)))
+
   (define-instance (Show :a => Show (Optional :a))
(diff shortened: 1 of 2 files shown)


Jason94:
It looks like a few tests depend on the old behavior. I'll fix those this weekend!

Edit: Fixed! Ready for review.

PR #2029, opened 2026-06-02
Docs: Split Whirlwind Tour into "Language Guide" and "Advanced Features"

Jason94:
This PR splits the "Whirlwind Tour of Coalton" into two pages: the "Coalton Language Guide" and the "Advanced Coalton Features Guide".

I thought that the following sections in the whirlwind tour are pretty advanced, and moved them to the new doc:
* instance defaulting
* functional dependencies
* specialization

This is also somewhat in anticipation of the GADT PR, since I think that GADTs definitely belong in the "advanced" section. Most users probably don't need to even worry about GADTs.


stylewarning:
I need to double check this doesn't break deployment.

PR #2033, opened 2026-06-18
doc: Fix export of instances on private types.

### LINKED ISSUE #1991: stdlib docs shouldn't show instances for private symbols

judah-sotomayor:
In the stdlib documentation generator,
do not list class instances on types with a private/non-exported component.

This is achieved by recursively calling stdlib-p on the symbol of each type that makes up the overall instance type.

Closes #1991.

### DIFF
--- a/src/doc/environment.lisp
+++ b/src/doc/environment.lisp
@@ -108,9 +108,22 @@ If non-nil, restrict to names defined in PACKAGE."
 By default the global environment is queried.
 If non-nil, restrict to instances defined in PACKAGE."
   (remove-if (lambda (class-instance)
-               (and package
-                    (not (exported-symbol-p (tc:ty-predicate-class
-                                             (tc:ty-class-instance-predicate class-instance))
-                                            package t))))
+               (let ((symbol (tc:ty-predicate-class
+                              (tc:ty-class-instance-predicate class-instance)))
+                     (types (mapcan #'tc:flatten-type
+                                    (tc:ty-predicate-types
+                                     (tc:ty-class-instance-predicate class-instance)))))
+                 
+                 (not
+                  (if package
+                      (exported-symbol-p symbol package t)
+                      (labels ((all-in-stdlib-p (ty)
+                                               (typecase ty
+                                                 ((or tc:tyvar tc:function-ty) t)
+                                                 (tc:tycon (stdlib-p (tc:tycon-name ty)))
+                                                 (tc:tapp (every #'all-in-stdlib-p (tc:flatten-type ty)))
+                                                 (t (stdlib-p ty)))))
+                             (every #'all-in-stdlib-p
+                              (cons symbol types)))))))
              (%lm-values (tc:instance-environment-instances
                           (tc:environment-instance-environment environment)))))
--- a/src/doc/environment.lisp
+++ b/src/doc/environment.lisp
@@ -102,28 +102,27 @@ If non-nil, restrict to names defined in PACKAGE."
              (%values (tc:environment-name-environment environment))))
 
 (defun find-instances (&key (environment entry:*global-environment*)
-                            (package nil))
+                         (package nil))
   "Return all instances in ENVIRONMENT.
 
 By default the global environment is queried.
 If non-nil, restrict to instances defined in PACKAGE."
-  (remove-if (lambda (class-instance)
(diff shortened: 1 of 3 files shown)


src/doc/environment.lisp
```diff
@@ -108,9 +108,22 @@ If non-nil, restrict to names defined in PACKAGE."
...
+               (let ((symbol (tc:ty-predicate-class
+                              (tc:ty-class-instance-predicate class-instance)))
+                     (types (mapcan #'tc:flatten-type
+                                    (tc:ty-predicate-types
+                                     (tc:ty-class-instance-predicate class-instance)))))
+                 
+                 (not
+                  (if package
+                      (exported-symbol-p symbol package t)
+                      (labels ((all-in-stdlib-p (ty)
```
stylewarning: can we lift this LABELS up higher and/or make this code a little more direct or succinct? it's ultimately a single boolean expression but having embedded lets/labels is a little smelly.

stylewarning (changes requested):

src/doc/environment.lisp
judah-sotomayor:
```lisp
(defun find-instances (&key (environment entry:*global-environment*)
                         (package nil))
  "Return all instances in ENVIRONMENT.

By default the global environment is queried.
If non-nil, restrict to instances defined in PACKAGE."
(labels
    ((public-ty-p (ty)
       (typecase ty
         ((or tc:tyvar tc:function-ty) t)
         (tc:tycon (stdlib-p (tc:tycon-name ty)))
         (tc:tapp (every #'public-ty-p (tc:flatten-type ty)))
         (t (stdlib-p ty)))))

  (remove-if-not (lambda (instance)
                   (if package
                       (exported-symbol-p (tc:ty-predicate-class (tc:ty-class-instance-predicate instance)) package t)
                       (every #'public-ty-p
                              (cons (tc:ty-predicate-class (tc:ty-class-instance-predicate instance))
                                    (mapcan #'tc:flatten-type (tc:ty-predicate-types
                                                               (tc:ty-class-instance-predicate instance)))))))

                 (%lm-values (tc:instance-environment-instances
                              (tc:environment-instance-environment environment))))))
```

Is this more in line? Or is it better to pull out the struct accesses into &aux variables to avoid the repetition?

judah-sotomayor:
```lisp
(defun find-instances (&key (environment entry:*global-environment*)
                         (package nil))
  "Return all instances in ENVIRONMENT.

By default the global environment is queried.
If non-nil, restrict to instances defined in PACKAGE."
  (labels
      ((public-ty-p (ty)
         (typecase ty
           ((or tc:tyvar tc:function-ty) t)
           (tc:tycon (stdlib-p (tc:tycon-name ty)))
           (tc:tapp (every #'public-ty-p (tc:flatten-type ty)))
           (t (stdlib-p ty)))))

    (remove-if-not (lambda (instance &aux
                                       (predicate (tc:ty-class-instance-predicate instance))
                                       (class (tc:ty-predicate-class predicate)))
                     (if package
                         (exported-symbol-p class package t)
                         (every #'public-ty-p
                                (cons class
                                      (mapcan #'tc:flatten-type (tc:ty-predicate-types predicate))))))

                   (%lm-values (tc:instance-environment-instances
                                (tc:environment-instance-environment environment))))))
```

I hesitated to suggest `&aux` because it's not used elsewhere, but in this case it is a bit cleaner. The only other thing I can think of is to use a symbol-macrolet to hide the accessor calls, but that seems a bit overkill.

judah-sotomayor:
Something else to note is that I pass through all function types without further checking, I don't think that's reasonable but I did not deep-dive the function type to see how it should be introspected

src/doc/environment.lisp
stylewarning: No aux.

You can just use LET (not LABELS). You're not using the function in calling position anyway. Something schematically like this:

```
(let* ((public? (lambda ...))
       (aux-vars ...)
       (filter-fn (lambda ...)))
 ...)
```

judah-sotomayor: Because public? is recursive for nested types, I've extracted it to a separate function.

stylewarning:
This should be fixed now as a result of a pile of other recent refactorings. Thank you very much for looking into this and making a contribution.

Issue #2028, opened 2026-06-01
Bug: define-type shouldn't allow constructor definitions if (repr :native)


Jason94:
If a `define-type` is tagged with a `(repr :native x)` annotation, then it is represented at runtime as an instance of the `x` Common Lisp type. As a consequence, `repr :native` types shouldn't allow constructors to be defined.

Unfortunately, that is currently allowed. This compiles with no errors or warnings:
```lisp
(cl:in-package :cl-user)

(defpackage #:test
  (:use
   #:coalton
   #:coalton-prelude
   ))

(in-package :test)

(named-readtables:in-readtable coalton:coalton)

(coalton-toplevel
  (repr :native cl:string)
  (define-type Test
    (TestConstructor Integer))
  )
```

Actually trying to use `TestConstructor` *does* compile in Coalton, but SBCL throws a type warning:
```lisp
(coalton-toplevel
  (declare foo (Void -> Test))
  (define (foo)
    (TestConstructor 1))
  )
```

Then using this function causes a runtime type error (on standard safety settings. God only knows if `(safety 0)`):
```
Values
  [#.(TESTCONSTRUCTOR 1)]
from (SB-KERNEL:%MAKE-STRUCTURE-INSTANCE
      #<SB-KERNEL:DEFSTRUCT-DESCRIPTION TEST/TESTCONSTRUCTOR {1202FA3ED3}>
      '((:SLOT COMMON-LISP:T . 0))
      (COMMON-LISP:THE COMMON-LISP:INTEGER |_0|)) in
(COMMON-LISP:THE (COMMON-LISP:VALUES COMMON-LISP:STRING COMMON-LISP:&OPTIONAL)
                 (TESTCONSTRUCTOR 1))
is not of type
  (VALUES STRING &OPTIONAL).
   [Condition of type COMMON-LISP:SIMPLE-TYPE-ERROR]
```

PR #2034, opened 2026-07-07
Fix define type repr native

iwami4438:
This PR introduces a validation check to ensure that define-type with (repr :native ...) does not include constructors.(Ref: https://github.com/coalton-lang/coalton/issues/2028)

Since :native types are intended to map directly to Lisp types, allowing constructors creates a design conflict within Coalton's type system. This validation enforces consistency by raising a compile-time error when such a definition is encountered.

Impact on Tests:

    Added native-repr-invalid.txt to verify that constructors for :native types are correctly rejected.

    Added native-repr-valid.txt to confirm that valid :native definitions are still supported.

    Note: This change identified a design conflict in recursive-let.txt, where a :native type was used with constructors. I have removed the :native attribute to align with the new validation rule.

Discussion Point:
During the implementation, I found that recursive-let.txt relied on :native as a workaround to provide Lisp-level type hints (declaim) while maintaining recursive structure.

    Should we migrate such patterns to standard Coalton types?

    Or, is there a need for a formal repr (e.g., :lisp-struct) to support recursive types with native-hinted representations?

I would appreciate your feedback on the preferred design direction for such cases.

### DIFF
--- a/src/typechecker/define-type.lisp
+++ b/src/typechecker/define-type.lisp
@@ -605,7 +605,7 @@ This is conservative and intentionally aligns with mutable native wrappers."
                 :collect (tc:quantify-using-tvar-order tvars (tc:qualify nil ty)))
 
          ;; Check that repr :enum types do not have any constructors with fields
-         :when (eq repr-type :enum)
+         :when (eq repr-type ':enum)
            :do (loop
                  :for ctor :in (parser:toplevel-define-type-ctors type)
                  :unless (endp (parser:constructor-fields ctor))
@@ -614,21 +614,21 @@ This is conservative and intentionally aligns with mutable native wrappers."
                                           "constructors of repr :enum types cannot have fields")))
 
          ;; Check that repr :transparent types have a single constructor
-         :when (eq repr-type :transparent)
+         :when (eq repr-type ':transparent)
            :do (unless (= 1 (length (parser:type-definition-ctors type)))
                  (tc-error "Invalid repr :transparent attribute"
                            (tc-note type
                                     "repr :transparent types must have a single constructor")))
 
          ;; Check that the single constructor of a repr :transparent type has a single field
-         :when (eq repr-type :transparent)
+         :when (eq repr-type ':transparent)
            :do (unless (= 1 (length (parser:type-definition-ctor-field-types (first (parser:type-definition-ctors type)))))
                  (tc-error "Invalid repr :transparent attribute"
                            (tc-note (first (parser:type-definition-ctors type))
                                     "constructors of repr :transparent types must have a single field")))
 
          ;; Check that repr :native types has no constructor
-         :when (eq repr-type :native)
+         :when (eq repr-type ':native)
            :do (unless (endp (parser:type-definition-ctors type))
                  (tc-error "Invalid repr :native attribute"
                            (tc-note type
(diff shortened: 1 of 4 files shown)


stylewarning:
This generally seems like the right thing, though in the future I hope that we can allow constructors to have their own native repr. But that's definitely a future thing requiring extensive design.

src/typechecker/define-type.lisp
```diff
@@ -626,6 +626,14 @@ This is conservative and intentionally aligns with mutable native wrappers."
                  (tc-error "Invalid repr :transparent attribute"
                            (tc-note (first (parser:type-definition-ctors type))
                                     "constructors of repr :transparent types must have a single field")))
+
+         ;; Check that repr :native types has no constructor
+         :when (eq repr-type :native)
```
stylewarning: please quote data keywords: `':native`

stylewarning (changes requested):
Just a small change.

src/typechecker/define-type.lisp
iwami4438: Should I also quote other keywords like :transparent for consistency, or is this specific to this?
I'm happy to align with your preference before pushing the changes.

stylewarning: All data keywords should be this way. It looks like others may have slipped through previous reviews.

iwami4438: Understood. I'll apply the quoting to :native in this PR. Should I also submit a separate PR later to fix the inconsistent keyword quoting throughout the codebase?

stylewarning: No need to do it throughout unless you want to. In this case it's fine if it's limited to the surrounding area.

iwami4438: Sure thing. I'll update the keywords in the immediate context of the changes I'm making.

iwami4438:
Load this with `sbcl --load`
```lisp
; CI reproduction
(ql:quickload :coalton)
(cl:in-package :cl-user)

(defpackage #:test
  (:use
   #:coalton
   #:coalton-prelude))


(in-package :test)

(coalton-toplevel
  (define-type (MyNullableLoop :elt)
    (NonNull :elt)
    (MyLoop (MyNullableLoop :elt)))
  (declare make-loop (Void -> (MyNullableLoop :any)))
  (define (make-loop)
    (let ((lp (MyLoop lp)))
      lp)))
```
then return `T`.

So I debugging now...
I'll probably fix `check-for-invalid-recursive-scc` in `define.lisp`.

stylewarning:
Tests are still failing, probably for a mostly trivial reason.

iwami4438:
It passed the other tests, but failed only this one due to insufficient stack size, and I was unable to reproduce the issue locally.
`tests/runtime-tests.ct`
```lisp
(cl:in-package #:tail-call-elimination-tests)

(coalton-toplevel

  (define-class (Bar1 :t)
    (bar1 (:t -> :t)))

  (define-instance (Bar1 UFix)
    (define (bar1 x)
      (if (< x 10000000)
          (foo1 x)
          x)))

  (define-instance (Bar1 Integer)
    (define (bar1 x)
      (if (< x 10000000)
          (foo1 x)
          x)))

  (define (foo1 x)
    (bar1 (1+ x))))

```

PR #2042, opened 2026-07-21
Define IntoIterator instance for FileStream

dieggsy:
I've been working on [coalton-utf-8](https://git.sr.ht/~dieggsy/coalton-utf-8) and thought it would be useful to have an iterator instance on FileStream. That way the rich iterator library or anything else that works on iterators can be used on files.

I'm not sure if signalling an error is the "right thing" here, but in any case I felt an iterator on the direct data was more useful than one on Result.

Side question: any plans to support general non-file streams (in the CL sense, e.g. string stream, concatenated stream, broadcast stream)?

### DIFF
--- a/library/file.ct
+++ b/library/file.ct
@@ -543,7 +543,16 @@ Automatically returns the lisp condition if one is thrown."
     (define (read fs)
       (read-char fs))
     (define (write fs data)
-      (write-char fs data))))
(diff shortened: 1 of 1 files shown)


stylewarning (approved):
seems like a reasonable convenience to me

PR #2052, opened 2026-08-01
Small improvements to char STL package

Jason94:
* Inline a few functions
* Refactor ASCII digit constants
* Add digit->ufix function

### DIFF
--- a/library/char.ct
+++ b/library/char.ct
@@ -70,6 +70,11 @@
       (lisp (-> Boolean) (x y)
         (to-boolean (cl:char= x y)))))
 
+  (define-instance (Default Char)
+    (inline)
+    (define (default)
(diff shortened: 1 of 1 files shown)


Jason94:
I'm also adding a `Default` instance which I recognize is a _little_ suspicious, but I needed it for something. I'd argue it's defensible, though, because there's no guarantee that a default needs to be some kind of identity in a group over the type or something. And it's very nice to be able to count on being able to create an arbitrary value of any kind of primitive type.

Edit: Actually, I don't think it's suspicious. I think it's fine. It's not declaring it a `Monoid` or anything - that would be _very_ suspicious.

stylewarning (approved):

